PR #32, opened 2023-09-13
upload POC

pierrotsmnrd:
This is a POC demonstrating the upload of files, with a mechanism similar to AWS S3 pre-signed URLs.

The goal is for the developers to enable upload to S3, or to a local server, with the same kind of features (upload URL has a TTL, use of a hash to prevent forging URLs and spam-upload, etc)

The POC is hence made of two parts :
- the FastAPI backend, that handles the upload endpoint
- a panel app, that serves a frontend, to select files and upload them.

All the POC is under `examples/upload`, check the Readme there to start both at the same time.

Extra informations : 
- The UI is very simple, just a basic HTML input type file. the important part is the JS part.
- the files are uploaded under `/tmp`
- some values are hardcoded ( like the URL of the upload backend in the JS file, or the `/tmp` dir), there are easy improvements and good enough for a POC.


pierrotsmnrd:
Superseeded by #61

PR #61, opened 2023-10-05
Add UI

### LINKED ISSUE #22: META - Demo User Interface Creation
```[tasklist]
### Tasks
- [ ] https://github.com/Quansight/ragna/issues/23
- [ ] Create Panel App as per the design
```

pmeier:
Closes #22


environment-dev.yml
```diff
@@ -17,6 +17,11 @@ dependencies:
     - types-aiofiles
     - sqlalchemy-stubs
     # documentation
+    
+    # Temporary dependency for the new Panel Chat Interface.
+    # TODO: remove this dependency once the new Panel Chat Interface is released,
+    # and pin Panel's version
+    - git+https://github.com/holoviz/panel@chat_components
```
pmeier, line 24: I've added it to `pyproject.toml`. Release should be soon and until then we can `pip install --pre panel`.

environment-dev.yaml
```diff
@@ -0,0 +1,38 @@
...
+  - redis-server
+  - pip
+  - aiofiles
+  - emoji
+
+  # for development only
+  - pre-commit
+
+  # For use with Panel branch for Chat Interface
+  - nodejs
```
pmeier, line 15: I couldn't find any usage of `node`. What was this for? Can we remove this whole file?

examples/panel_app/README.md
```diff
@@ -0,0 +1,3 @@
+
+
+ragna api -c demo & python examples/panel_app/main.py && fg
```
pmeier, line 3: I've added a CLI for the UI, which automatically starts the API if needed. Can we remove `examples/panel_app`?

pyproject.toml
```diff
@@ -107,11 +108,20 @@ warn_unused_ignores = true
 warn_return_any = true
 warn_unused_configs = true
 
+[[tool.mypy.overrides]]
```
pmeier, line 111: No need to fight `mypy` while still in heavy development mode. We can enable this later as it is not user facing anyway.

ragna/_api/core.py
```diff
@@ -17,6 +20,13 @@ def app(config: Config) -> FastAPI:
     rag = Rag(config)
 
     app = FastAPI(title="ragna", version=ragna.__version__)
+    app.add_middleware(
+        CORSMiddleware,
+        allow_origins=get_origins(config.ui.url),
```
pmeier, line 25: @pierrotsmnrd You added a patch that allowed connections from the API as well. Could you explain why? I've reverted here to UI url only and for me everything worked fine.

ragna/ui/api_wrapper.py
```diff
@@ -0,0 +1,108 @@
...
+import emoji
+
+import httpx
+
+
+# The goal is this class is to provide ready-to-use functions to interact with the API
+class ApiWrapper:
+    def __init__(self, api_url):
+        # FIXME: this should be an async client
+        self.client = httpx.Client(base_url=api_url)
```
pmeier, line 14: Why do we have any sync functions on the UI side anyway. Shouldn't everything by async?

ragna/ui/api_wrapper.py
```diff
@@ -0,0 +1,108 @@
...
+        for chat in json_data:
+            for msg in chat["messages"]:
+                # convert timestamps to datetime
+                msg["timestamp"] = datetime.strptime(
+                    msg["timestamp"], "%Y-%m-%dT%H:%M:%S.%f"
+                )
+
+                msg["content"] = self.replace_emoji_shortcodes_with_emoji(
+                    msg["content"]
+                )
```
pmeier, line 47: This is why the emoji rendering works after page reload. We should apply this function to every message that gets added to the chat widget.

ragna/ui/api_wrapper.py
```diff
@@ -0,0 +1,108 @@
...
+    async def get_components_async(self):
+        async with httpx.AsyncClient(
+            base_url=self.client.base_url, headers=self.client.headers
+        ) as async_client:
+            return (await async_client.get("/components")).raise_for_status().json()
+
+    # Upload and related functions
+    def upload_endpoints(self):
+        return {
+            "informations_endpoint": f"{self.client.base_url}/document",
```
pmeier, line 67: I've removed the upload endpoint here, since it is not needed. More on that later.

ragna/ui/app.py
```diff
@@ -0,0 +1,72 @@
...
+            profiler="pyinstrument",
+            allow_websocket_origin=[
+                urlsplit(origin).netloc for origin in get_origins(self.url)
+            ],
+            static_dirs={"imgs": str(IMGS), "resources": str(RES)},  # "css": str(CSS),
+        )
+
+
+def app(config: Config) -> App:
+    return App(url=config.ui.url, api_url=config.api.url)
```
pmeier, line 72: We should probably pass `config` directly to the `App`.

ragna/ui/components/file_uploader.py
```diff
@@ -0,0 +1,249 @@
...
+            final_callback_js
+            + f""" upload( self.get_upload_files(),  '{self.token}', '{self.informations_endpoint}', final_callback) """
+        )
+
+    _child_config = {
+        "custom_js": "template",
+        "uploaded_documents_json": "template",
+    }
+
+    _template = """
```
pmeier, line 50: Can we please move all JS / CSS that is longer than say 3 lines to a `.js` / `.css` file? This module is really hard to read.

ragna/ui/imgs/ragna_logo.svg
pmeier, line 1: @pavithraes is adding logos in #100. Let's make sure we aren't duplicating files here.

ragna/ui/js.py
```diff
@@ -0,0 +1,107 @@
+from .styles import MODAL_MAX_HEIGHT
+
+# Reset modal size
+MODAL_HACK = f"""
```
pmeier, line 4: Can we have them in native files, i.e. not `.py`? Like this, it is impossible to apply any formatting.

ragna/ui/resources/upload.js
```diff
@@ -0,0 +1,51 @@
...
+function upload(files, token, informationsEndpoint, final_callback) {
+    var uploaded_documents = [] 
+
+    Array.from(files).map(file => {
+
+        getUploadInformations(file, token, informationsEndpoint).then(
+
+                function(uploadInformations) {
+                        uploadFile(uploadInformations.url, uploadInformations.data, file)
+                        uploaded_documents.push(uploadInformations.document)
```
pmeier, line 12: This originally was filled by the response of `uploadFile`. However, there is no protocol on what this should return. Our built-in endpoint indeed returns the document information. But S3 for example doesn't. Thus, we need to use `uploadInformations`.

ragna/ui/resources/upload.js
```diff
@@ -0,0 +1,51 @@
...
+
+        return uploadInformations
+}
+
+async function uploadFile(url, data, file) {
+    var body = new FormData()
+
+    for (const [key, value] of Object.entries(data)) {
+      body.append(key, value);
+    }
```
pmeier, line 39: Instead of just pulling out the token, we need to include everything that is returned by the upload info. For our endpoint this is indeed just the token, but for S3 it is a lot more.

ragna/ui/resources/upload.js
```diff
@@ -0,0 +1,51 @@
...
+
+function upload(files, token, informationsEndpoint, final_callback) {
+    var uploaded_documents = [] 
+
+    Array.from(files).map(file => {
+
+        getUploadInformations(file, token, informationsEndpoint).then(
+
+                function(uploadInformations) {
+                        uploadFile(uploadInformations.url, uploadInformations.data, file)
```
pmeier, line 11: Since the upload information includes the URL, we can safely remove the extra `uploadEndpoint` that we previously took as parameter to `upload`. For all non.default uploads that wouldn't have worked anyway, because we don't know the upload endpoint before retrieving the upload information.

pmeier (commented):
@pierrotsmnrd As discussed offline, I've updated this after I merged my auth PR. Some high level comments:

- [ ] The modal is opening if there are no chats. I'm fairly certain we discussed that we don't want that. cc @smeragoel to be sure.
- [ ] A button to delete each chat is missing.
- [x] The "start conversation" is clickable even without uploaded documents. This will get you a spinner that goes on forever.
- [ ] Emojis are not rendered when coming from the user or model (see screencast). They seem to be correct after the page is reloaded.
- [ ] For longer chats, the widget scrolls to the top after each message (see screencast)
- [ ] There is no advanced config in the modal yet
- [ ] Favicon is still the default panel one




[Screencast from 2023-10-23 01-13-08.webm](https://github.com/Quansight/ragna/assets/6849766/50b138bb-e0c2-4bd3-8288-913ff3cd86fb)

More inline

environment-dev.yaml
pierrotsmnrd: I guess my comment isn't clear enough, sorry about that. You need nodejs to install a panel version from a git repo, panel needs nodejs at some point in its installation process. This can be removed once the next Panel version is released

ragna/_api/core.py
pierrotsmnrd: I did it for the sake of indicating explicitly all allowed origins but yes, that's not necessary.

ragna/ui/components/file_uploader.py
pierrotsmnrd: Yes that's possible, that's why we have `upload.js` for instance. But I'll do that later, the things are still moving a lot so for the moment it's more convenient for me to have everything at one place than scattered.

It's hard to read because that's a custom Panel component, so it holds 3 parts **in multiple languages** : 
- its HTML template, which includes its CSS styling
- all the JS machinery to make it work in the frontend, and junction with Python
- the python part, which is the backend

The CSS will be moved somewhere else and simplified anyway, the JS part can move in a dedicated file.

ragna/ui/js.py
pierrotsmnrd: Yes, open an issue for that. I won't move them all to separate files though, some are small and not worth it.

pierrotsmnrd:
@pmeier heads up, I noticed that after merging the `auth` branch into `ui`, the move from `environment-dev.yaml` to `environment-dev.yml` result in having both files. Can you have a look please ?

environment-dev.yaml
pmeier: In the light of https://github.com/Quansight/ragna/pull/61#discussion_r1367989031, we can probably just drop the file.

ragna/ui/resources/upload.js
```diff
@@ -1,18 +1,25 @@
...
 
                 function(uploadInformations) {
-                        uploadFile(uploadInformations.url, uploadInformations.data, file)
-                        uploaded_documents.push(uploadInformations.document)
-                        if (uploaded_documents.length == files.length) {
-                            final_callback(uploaded_documents);
-                        }
+
+                        uploadFile(uploadInformations.url, uploadInformations.data, file).then(
+                            function(response) {
```
pmeier, line 14: @pierrotsmnrd While your patch here fixes the waiting for the upload to be completed, it reverts one that I made earlier (see https://github.com/Quansight/ragna/pull/61#discussion_r1367991066). To reiterate: we **cannot** use the `response` of the upload to fill the `uploaded_documents`. Our local upload endpoint indeed returns the document data, i.e. id and name. This response is "user configured".

We need to revert back to what I had, but need to wait for the upload to complete. Maybe it is as easy as `uploadFile(...).then(function(response) {})`?

pmeier: @pierrotsmnrd I've ventured into JS again and refactored the whole file. With the snippet below we now perform the upload in parallel since this is heavily I/O bound.

```js
function upload(files, token, informationsEndpoint, final_callback) {
  Promise.all(
    Array.from(files).map((file) => {
      return uploadFile(file, token, informationsEndpoint);
    }),
  ).then(final_callback);
}

async function uploadFile(file, token, informationEndpoint) {
  const response = await fetch(`${informationEndpoint}?name=${file.name}`, {
    headers: { Authorization: `Bearer ${token}` },
  });
  const documentInfo = await response.json();

  var body = new FormData();
  for (const [key, value] of Object.entries(documentInfo.data)) {
    body.append(key, value);
  }
  body.append("file", file);

  await fetch(documentInfo.url, {
    method: "POST",
    body: body,
  });

  return documentInfo.document;
}
```

The signature of the `upload` function is the same, so there is no need to change anything besides this file. Please have a look and LMK if we can go for this. I made sure we properly await everything this time :innocent:

pierrotsmnrd: Looks good to me, tested and approved :) I'm going to commit and credit you

ragna/ui/__init__.py
```diff
@@ -0,0 +1 @@
+from .app import app
```
pmeier, line 1: One thing that we need to do before merge is to rename `ragna.ui` to `ragna._ui`. This should be private from the user perspective. Will do that as soon as everything else is ready.

ragna/assistants/_api.py
```diff
@@ -28,7 +28,7 @@ def __init__(
 
         self._client = httpx.Client(
             headers={"User-Agent": f"{ragna.__version__}/{self}"},
-            timeout=10,
+            timeout=None,
```
pmeier, line 31: Ok, to increase, but we should not wait indefinitely.

```suggestion
            timeout=30,
```

ragna/ui/api_wrapper.py
```diff
@@ -0,0 +1,143 @@
...
+                    )
+                )
+                .raise_for_status()
+                .json()
+            )
+
+            self.client.headers["Authorization"] = f"Bearer {self.token}"
+
+            print("token : ", self.token)
+
```
pmeier, line 55:
```suggestion
```

ragna/ui/app.py
```diff
@@ -0,0 +1,111 @@
...
+
+        self.template = self.get_template()
+
+    def get_template(self):
+        template = pn.template.FastListTemplate(
+            # We need to set a title to have it appearing on the browser's tab
+            # but it means we need to hide it from the header bar
+            title="Ragna",
+            # neutral_color="#FF0000", #ui.MAIN_COLOR,
+            # header_background="#FF0000", #ui.MAIN_COLOR,
```
pmeier, line 47: DO we need to handle this?

ragna/ui/app.py
```diff
@@ -0,0 +1,111 @@
...
+            accent_base_color=ui.MAIN_COLOR,
+            theme_toggle=False,
+            collapsed_sidebar=True,
+            # main_layout=None
+            raw_css=[ui.APP_RAW],
+            favicon="imgs/ragna_logo.svg",
+            css_files=["https://rsms.me/", "https://rsms.me/inter/inter.css"],
```
pmeier, line 54: This is for the font, right?

ragna/ui/app.py
```diff
@@ -0,0 +1,111 @@
...
+        auth_page = AuthPage(api_wrapper=self.api_wrapper)
+        self.template.main.objects = [auth_page.page()]
+        return self.template
+
+    def health_page(self):
+        return pn.pane.HTML("<h1>Ok</h1>")
+
+    def serve(self):
+        # logging.init(log_level=args.log_level)
+
```
pmeier, line 84:
```suggestion
```

ragna/ui/central_view.py
```diff
@@ -0,0 +1,681 @@
...
+
+    def trigger_on_click_source_info_callback(self, event):
+        if self.on_click_source_info_callback is not None:
+            self.on_click_source_info_callback(event, self)
+
+    def update_css_classes(self):
+        role = self.msg_data["role"] if "role" in self.msg_data else None
+        self.css_classes = [
+            "chat-entry",
+            "chat-entry-user" if role == "user" else "chat-entry-ragna",
```
pmeier, line 180:
```suggestion
            f"chat-entry-{role}",
```

ragna/ui/central_view.py
```diff
@@ -0,0 +1,681 @@
...
+
+        # should never happen
+        return "?"
+
+    @classmethod
+    def chat_entry_value_renderer(cls, txt, role):
+        markdown_css_classes = []
+        if role is not None:
+            markdown_css_classes = [
+                "chat-entry-user" if role == "user" else "chat-entry-ragna"
```
pmeier, line 209: See above.

ragna/ui/central_view.py
```diff
@@ -0,0 +1,681 @@
...
+                                                                    
+                                                                }   
+                                                 
+
+                                                """.replace(
+                                "{{GRID_HEIGHT}}", str(grid_height)
+                            )
+                        ],
+                    ),
+                    # stylesheets=[""" :host {background-color:red; width:100%;} .bk-panel-models-layout-Column { width: 100%; } """ ]
```
pmeier, line 295: Remove?

```suggestion
```

ragna/ui/central_view.py
```diff
@@ -0,0 +1,681 @@
...
+                        stylesheets=[""" hr { width: 94%; height:1px;  }  """],
+                    ),
+                    # Debug bloc :
+                    # pn.pane.Markdown(
+                    #         f"""Chat ID: {self.current_chat['id']} <br />
+                    #                     Message ID: {msg.msg_data['id']} <br />
+                    #                     Sources: {msg.msg_data['sources']}""",
+                    #         dedent=True,
+                    #     stylesheets=[""" :host {  background-color: red ; }  """]
+                    # ),
```
pmeier, line 330: Remove?

```suggestion
```

ragna/ui/central_view.py
```diff
@@ -0,0 +1,681 @@
...
+        try:
+            answer = self.api_wrapper.answer(self.current_chat["id"], contents)
+
+            self.current_chat["messages"].append(answer["message"])
+
+            yield {
+                "user": "Ragna",
+                "avatar": "🤖",
+                "value": answer["message"]["content"],
+            }
```
pmeier, line 351: Not sure I understand. User cannot be hardcoded here and the avatar should also be dynamic?

ragna/ui/central_view.py
```diff
@@ -0,0 +1,681 @@
...
+        except Exception as e:
+            print(e)
+
+            yield {
+                "user": "Ragna",
+                "avatar": "❌",
```
pmeier, line 357: 1. Are we using `"Ragna"` as user for the panel chat interface? Can we just use the role of the message? Here that would be "system"
2. Can we use our logo for the avatar as we do for the other system messages like the welcome message?

ragna/ui/central_view.py
```diff
@@ -0,0 +1,681 @@
...
+        #     [
+        #     w
+        #     for w in chat_interface._composite.objects
+        #     if not isinstance(w, pn.layout.spacer.VSpacer)
+        # ])
+        # chat_interface._chat_log.stylesheets.append(
+        #     """ :host .chat-feed-log {
+        #             height: unset; max-height: 90%; }
+        #     """
+        # )
```
pmeier, line 463: We no longer want that, do we? If not, remove?

```suggestion
```

ragna/ui/central_view.py
```diff
@@ -0,0 +1,681 @@
...
+                trigger_scroll_to_latest is triggered in the chat_interface method, when chat_interface.objects changes.
+
+        Twist : the HTML script node needs to have a different ID each time it is rendered,
+                otherwise the browser doesn't re-render it / doesn't execute the JS part.
+                Hence the random ID.
+        """
+
+        random_id = "".join(
+            random.choice("".join([chr(n) for n in range(65, 91)])) for _ in range(16)
+        )
```
pmeier, line 533: That looks rather complicated to just get a random ID. What about

```suggestion
        random_id = str(uuid.uuid4())
```

ragna/ui/central_view.py
```diff
@@ -0,0 +1,681 @@
...
+            ],
+        )
+
+    def set_loading(self, is_loading):
+        self.main_column.loading = is_loading
+
+    def __panel__(self):
+        """
+        The ChatInterface.view_latest option doesn't seem to work.
+        So to scroll to the latest message, we use some JS trick.
```
pmeier, line 654: We need to report this to `panel`. @pierrotsmnrd Could you compile a list of stuff like this that we encountered so they can be fixed upstream. Not blocking for this PR.

ragna/ui/components/file_uploader.py
```diff
@@ -0,0 +1,274 @@
...
+        final_callback_js = """
+            var final_callback = function(uploaded_documents) {
+                self.get_uploaded_documents_json().innerText = JSON.stringify(uploaded_documents);
+            };
+        """
+
+        # This is a hack to force the re-execution of the javascript.
+        # If the whole javascript is the same, and doesn't change,
+        # the panel widget is not re-renderer, and the upload function is not called.
+        random_id = f" var random_id = '{str(random())}';"
```
pmeier, line 61: This gives a random float. Should we use a different scheme like I suggested above with a random UUID?

ragna/ui/left_sidebar.py
```diff
@@ -0,0 +1,203 @@
...
+
+class LeftSidebar(pn.viewable.Viewer):
+    chats = param.List(default=[])
+    current_chat_id = param.String(default=None)
+    refresh_counter = param.Integer(default=0)
+
+    def __init__(self, api_wrapper, **params):
+        super().__init__(**params)
+
+        # self.api_wrapper = api_wrapper
```
pmeier, line 15:
```suggestion
```

ragna/ui/main_page.py
```diff
@@ -0,0 +1,175 @@
...
+                if c["id"] == self.current_chat_id:
+                    self.central_view.set_current_chat(c)
+                    break
+
+    # Modal and callbacks
+    def open_modal(self):
+        self.modal = ModalConfiguration(
+            api_wrapper=self.api_wrapper,
+            new_chat_ready_callback=lambda new_chat_id: self.open_new_chat(new_chat_id),
+            cancel_button_callback=lambda event: self.on_click_cancel_button(event),
```
pmeier, line 64:
```suggestion
            new_chat_ready_callback=self.open_new_chat,
            cancel_button_callback=self.on_click_cancel_button,
```

ragna/ui/main_page.py
```diff
@@ -0,0 +1,175 @@
...
+        self.template = template
+
+        self.modal = None
+        self.central_view = CentralView(api_wrapper=self.api_wrapper)
+        self.central_view.on_click_chat_info = (
+            lambda event, title, content: self.show_right_sidebar(title, content)
+        )
+
+        self.left_sidebar = LeftSidebar(api_wrapper=self.api_wrapper)
+        self.left_sidebar.on_click_chat = lambda chat: self.on_click_chat(chat)
```
pmeier, line 27:
```suggestion
        self.left_sidebar.on_click_chat = self.on_click_chat
```

ragna/ui/main_page.py
```diff
@@ -0,0 +1,175 @@
...
+        self.right_sidebar.param.trigger("content")
+        self.right_sidebar.show()
+
+    @param.depends("current_chat_id", watch=True)
+    def update_subviews_current_chat_id(self, avoid_senders=[]):
+        try:
+            if self.left_sidebar is not None and self.left_sidebar not in avoid_senders:
+                self.left_sidebar.current_chat_id = self.current_chat_id
+        except Exception:
+            pass
```
pmeier, line 108: What could cause this exception and why can we ignore it?

ragna/ui/modal_configuration.py
```diff
@@ -0,0 +1,314 @@
...
+        if not self.document_uploader.can_proceed_to_upload():
+            self.change_upload_files_label("missing_file")
+        else:
+            self.start_chat_button.disabled = True
+            self.document_uploader.perform_upload(event, self.did_finish_upload)
+
+    def did_finish_upload(self, uploaded_documents):
+        # at this point, the UI has uploaded the files to the API.
+        # We can now start the chat
+        print("did finish upload", uploaded_documents)
```
pmeier, line 116:
```suggestion
```

ragna/ui/modal_configuration.py
```diff
@@ -0,0 +1,314 @@
...
+            self.start_chat_button.disabled = False
+
+            if self.new_chat_ready_callback is not None:
+                self.new_chat_ready_callback(new_chat_id)
+
+        except Exception:
+            self.change_upload_files_label("upload_error")
+            self.document_uploader.loading = False
+            self.start_chat_button.disabled = False
+            pass
```
pmeier, line 136:
```suggestion
```

pmeier (commented):
Two comments here and more inline.

- Button says "Chat info", but side bar says "Chat config"

  image
-  Chat info sidebar does not display advanced configuration

Thanks a lot @pierrotsmnrd! This is coming along great. Let's get these comments resolved and iterate on it further after merge.

ragna/ui/app.py
pierrotsmnrd: I'll remove that

ragna/ui/app.py
pierrotsmnrd: Yes it is

ragna/ui/central_view.py
pierrotsmnrd: Well that's not that simple. This would lead to have a CSS class `chat-entry-system`, which doesn't exist - unless I define it as identical to `chat-entry-ragna` which is supposed to be the default... I'll sort something out.

ragna/ui/central_view.py
pierrotsmnrd: We don't need it anymore, I kept it just in case, and for anyone curious to learn.

ragna/ui/components/file_uploader.py
pierrotsmnrd: it totally doesn't matter here, but I'll switch to uuid as well.

ragna/ui/main_page.py
pierrotsmnrd: Like many other things you reported, that's a leftover of the dev process

pierrotsmnrd:
I've fixed all the mentioned points, many were leftovers of the dev process. 

I've also fixed the Chat info button to look like the figma design, and added the chat params in the right sidebar
Capture d’écran 2023-10-26 à 14 15 01

ragna/ui/__init__.py
pmeier: Done in 3a0e2f0ac118882a679fdcfa488cd92502ee276e

ragna/ui/central_view.py
pmeier: Still don't understand how it works here. We are still yielding the generic robot avatar here, but we get the OpenAI logos later on? Something is fishy, but since it works for now, we can look into it later.

PR #74, opened 2023-10-16
add authentication

pmeier:
This adds a couple of changes. Here are the highlights:

- There is the `ragna.core.Authentication` base class now to define your own authentication scheme. The REST API got a `/token` endpoint that calls into the authentication class to produce a token. Each request to the other endpoints no also calls into the class to authenticate and return the user.
- The default implementation, i.e. `ragna.core.RagnaDemoAuthentication` generates JWTs based on username / password inputs. The username is arbitrary and the password can either be set through an environment variable or, if that doesn't exists, just has to match the username. The token has to be embedded into the header as bearer authorization, i.e. `headers["Authorization"] = f"Bearer {token}"`.
- I moved away from the fine grained optional dependencies. `pip install ragna` will now pull everything in that a user needs to have a fully running demo. Only the requirements of the builtin components stay optional, but are now also grouped under a single key.
- I've renamed `RagConfig` -> `CoreConfig` and in turn `Config().rag` -> `Config().core`, since it fits our naming scheme inside the package better.


ragna/source_storages/_chroma.py
```diff
@@ -8,7 +8,7 @@ class Chroma(SourceStorage):
     @classmethod
     def requirements(cls) -> list[Requirement]:
         return [
-            PackageRequirement("chromadb >=0.4"),
+            PackageRequirement("chromadb>=0.4.13"),
```
pmeier, line 11: We need `>=0.4.13`, because lower version require `pydantic<2`

scripts/add_chats.py
```diff
@@ -1,94 +1,125 @@
...
 
 def main():
-    url = "http://127.0.0.1:31476"
-    user = "Ragna"
+    client = httpx.Client(base_url="http://127.0.0.1:31476")
+    client.get("/").raise_for_status()
 
-    client = httpx.Client(base_url=url, headers={"X-User": user})
-    client.get("").raise_for_status()
+    ## authentication
```
pmeier, line 13: This is the only substantial change in this file. The remaining changes are just better error handling.

pierrotsmnrd (approved):
Tested locally, and works.

PR #89, opened 2023-10-19
use mkdocs-gallery for example / tutorial documentation

pmeier:
Per title. This PR removes the `examples/` folder and static markdown examples / tutorials and replaces them with executable gallery scripts.

We are going to lose the following two examples for now:

- Local LLMs: this was an absolute bare bones example with no narration / text. Plus, it is not up to date anyway and thus cannot be run. We should re-add such an example in the future.
- S3 documents: this functionality was moved into [`ragna-aws`](https://github.com/pmeier/ragna-aws) anyway, so we don't need to keep the example.

For that we gain one more example to showcase streaming LLM responses.


pmeier:
@pavithraes In smarie/mkdocs-gallery#90 I'm working on adding support for async galleries. With that, writing a gallery feels exactly like a notebook in the sense that we can use `await` and other async syntax without the need of wrapping everything in `asyncio.run` or the like.

I've switched this PR to the branch where I'm adding async galleries. Have a look at the script that is part of this PR and how it renders: https://ragna--89.org.readthedocs.build/en/89/generated/gallery/example_python_api/

environment-dev.yml
```diff
@@ -21,3 +21,5 @@ dependencies:
       - mkdocs-gen-files
       - material-plausible-plugin
       - mktestdocs
+      # FIXME: replace this with a stable release when available
+      - mkdocs-gallery @ git+https://github.com/pmeier/mkdocs-gallery@async-gallery
```
pmeier, line 25: My PR was merged and is available starting with

```suggestion
      - mkdocs-gallery >=0.10
```

pmeier:
Blocked by #316 and #317.

pavithraes:
Thanks for this PR, @pmeier :)

Overall, this looks good. My main (non-blocking) comment in around consistency in narrative text. It'll be nice to use a s[tandard style guide](https://developers.google.com/style/highlights) for all narrative elements. However, we can always update for this later, so we can merge for now . :)

PR #91, opened 2023-10-19
Update README.md

pavithraes:
Very quick clean-up because we made the repo public, will continue to make updates in future PRs.


pmeier:
LGTM

PR #92, opened 2023-10-19
add hook to ignore autogenerated files

### LINKED ISSUE #88: [BUG] `mkdocs serve` reloads continuously when using `hooks`
During local development using `mkdocs serve` to have a live docs render - the website reloads continuously even when the there are no changes. Commenting out the `hooks` section in `mkdocs.yml` solves this issues.

This needs more investigation, opening an issue to keep track of this behaviour and document the workaround.

pmeier:
Fixes #88. Unfortunately, there is no built-in option in `mkdocs` to solve this. Thus, we resort to a hack here, by monkeypatching the reload server at runtime. I'll open an issue upstream to officially support this.

### DIFF
--- a/docs/hooks/gitignore.py
+++ b/docs/hooks/gitignore.py
@@ -7,12 +7,7 @@ DOCS_ROOT = HERE.parent
 
 with open(DOCS_ROOT / ".gitignore") as file:
     GITIGNORE = pathspec.PathSpec.from_lines(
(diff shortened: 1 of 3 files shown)


pmeier:
`mkdocs` maintainer suggests to avoid this https://github.com/mkdocs/mkdocs/issues/3433#issuecomment-1771767142. #93 supersedes this PR by circumventing the original problem in the first place.

PR #94, opened 2023-10-20
Add mypy as static type checker

pmeier:


ragna/_api/orm.py
pavithraes, line 1: I'm seeing `error: Unused "type: ignore" comment  [unused-ignore]` on for all the type ignores in this file

ragna/_api/schemas.py
pavithraes, line 1: "Column[UUID]"; expected "UUID" I see several similar+related errors of the form: `incompatible type "Column[UUID]"; expected "UUID"  [arg-type]` (also, for `bool` & `str`), on the `Document`, `Chat`, and `ChatMetadata` classes in this file

ragna/_api/database.py
```diff
@@ -211,5 +210,5 @@ def update_chat(session: Session, user: str, chat: schemas.Chat) -> None:
 
 def delete_chat(session: Session, user: str, id: uuid.UUID) -> None:
     orm_chat = _get_orm_chat(session, user=user, id=id)
-    session.delete(orm_chat)
+    session.delete(orm_chat)  # type: ignore[no-untyped-call]
```
pavithraes, line 213: This line also raises `error: Unused "type: ignore" comment  [unused-ignore]`

pavithraes (commented):
Thanks, @pmeier! 

I ran mypy against this PR locally, and I've shared some errors that it raised. I have very brief experience with mypy, so please feel free to ignore or make updates as you see fit. :)

pavithraes (approved):

ragna/_api/schemas.py
pmeier: This and the error below as well come from `sqlalchemy`. For that one needs to `pip install sqlalchemy-stubs`. I did so in CI, but forgot to make them explicit as part of the dev env. This is why CI was green here.

I made them explicit now. Thanks for the heads up!

PR #95, opened 2023-10-20
Netlify configuration

pavithraes:


requirements.txt
```diff
@@ -0,0 +1,13 @@
+####
```
pmeier, line 1: Ok with this, but can we maybe call the file `requirements-dev.txt`? We could also rename `environment.yml` to `environment-dev.yml`. Note that the latter also need to be changed in 

https://github.com/Quansight/ragna/blob/e14e6e1d9032f0924b6359919fdb296c00e85d86/.github/actions/setup-env/action.yml#L40

and more occurrences in the same file.

pavithraes: I think Netlify requires the exact file name `requirements.txt` - but, while working through this, I see they've very slow to support new Python versions - currently they only support till python 3.8 🤷🏼‍♀️ 

They had a very smooth process for Docusaurus, but their Python story is not nice from what I can tell.

I think to future-proof, I'd rather move to ReadTheDocs.

pavithraes:
ref: https://github.com/Quansight/ragna/pull/95#discussion_r1366835116

requirements.txt
pavithraes: I do think renaming to `environment-dev.yml` is a good idea. I'll do that while setting up ReadTheDocs

PR #96, opened 2023-10-20
ReadTheDocs configuration

pavithraes:
- Add `.readthedocs.yml`
- Rename env file to `environment-dev.yml` per https://github.com/Quansight/ragna/pull/95#discussion_r1366812787


pmeier (commented):
How can we see the builds based on this PR? I have no idea if the RTD config works.

pmeier (approved):
As discussed offline: let's merge without actually seeing the result because it seems to be easier to setup this way. We can always iterate on it.

PR #97, opened 2023-10-20
Fix .readthedocs.yml to use conda

pavithraes:
Also, verify that build preview is triggered

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -22,4 +22,4 @@ python:
     - method: pip
       path: .
       extra_requirements:
-        - builtin
+        - complete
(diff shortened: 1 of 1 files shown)


.readthedocs.yml
```diff
@@ -14,6 +14,5 @@ mkdocs:
   configuration: mkdocs.yml
 
 # Optionally declare the Python requirements required to build your docs
-python:
-  install:
-  - requirements: environment-dev.yml
+conda:
```
pavithraes, line 17: Ref: https://docs.readthedocs.io/en/stable/guides/conda.html

pmeier (approved):
Build works now, but I saw that the REST API is broken. Need to look into that, but no need for this PR.

PR #100, opened 2023-10-22
Add initial narrative docs

pavithraes:
Supersedes #87 
Addresses #24
Closes #36 

- [x]  Landing page
- [x]  Installation
- [x]  Tutorial - Ragna 101 (Python)
- [x]  Set config file
- [x]  Community
- [x] Contribute
- [x] Add brand asset files (addresses: #80)
- [x]  FAQ
- [x]  And "Coming soon" message to other pages in the pipeline

PR #101, opened 2023-10-23
use prettier as markdown / yaml / toml formatter

pmeier:
If possible, I want to merge this before #100 to avoid too much noise afterwards. We can use the same hook to also format `.js` / `.css` / `.html` files that will be part of the repo after #61.


pavithraes (approved):
LGTM! Thank you!

Issue #25, opened 2023-09-07
META - Project health files


pavithraes:
```[tasklist]
### Tasks
- [x] Labels and milestones
- [x] Issue & PR Templates
- [x] README enhancement (CoC note)
- [x] LICENSE
- [x] Branch protections
- [x] security policy
- [x] precommit hooks (#38)
```

Removed roadmap and release strategy as non-critical, will open a separate issue with nice to have's

pavithraes:
Closing as complete 🚀

PR #102, opened 2023-10-23
Add security policy

pavithraes:
Needs some discussion before we can merge.

Addresses #25

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -2,10 +2,15 @@
 
 ## Supported Versions
 
-We support only the latest version, and we use [SemVer](https://calver.org/) for versioning which means minor version are backwards compatible.
+We support only the latest version, and we use [SemVer](https://semver.org/) for
+versioning which means minor version are backwards compatible.
 
-You should feel comfortable upgrading if you're using our documented public APIs and pay attention to `DeprecationWarnings`. Whenever there is a need to break compatibility, it is announced in the [Changelog](docs/references/changelog.md) and will raise a `DeprecationWarning` before it's finally really broken.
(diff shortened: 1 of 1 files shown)


pavithraes:
@pmeier - Some quick questions:

- I believe the first release is `0.1` - will we continue to use SemVer?
- Are the notes about supporting only the latest version and using `DeprecationWarnings` alright with you?
- Shall we enable vulnerability reporting on this repo (can be enabled here: https://github.com/Quansight/ragna/security) - again per the policy message?

Please feel free to suggest changes. :)

SECURITY.md
```diff
@@ -0,0 +1,16 @@
+# Security Policy
+
+## Supported Versions
+
+We support only the latest version, and we use [SemVer](https://semver.org/) for
+versioning which means minor version are backwards compatible.
```
pmeier, line 6: We should add a sentence here that as long as we are on a `0.x` release, there are no BC guarantees whatsoever. We are not planning to break BC, but it can happen without deprecation cycle even if we are still just upping the minor version. SemVer will only really kick in after the `1.0` release.

SECURITY.md
```diff
@@ -0,0 +1,16 @@
...
+
+You should feel comfortable upgrading if you're using our documented public APIs and pay
+attention to `DeprecationWarnings`. Whenever there is a need to break compatibility, it
```
pmeier, line 9:
```suggestion
attention to `DeprecationWarning`s. Whenever there is a need to break compatibility, it
```

pmeier (approved):
LGTM if my comments are addressed.

PR #103, opened 2023-10-23
Add issue+pr templates

pavithraes:
Addresses #25 

I understand issues templates are a bit polarising. I do feel if used well, we can increase the bar for issue/PR quality (which saves several back-and-forths & triage cycles in the long term) while also allowing maintainers flexibility. I've tried to use minimal templates with most fields marked optional (inspired by Nebari, but super trimmed-down), and added blank issues for maintainers.

Also, I think we can use the issue tracker to track usage questions for now. At this early stage, most questions are likely to be bugs or will need better documentation which we'll need to track on the issue tracker anyway. We can enable GitHub discussions if folks feel strongly, we can also enable it later if we see the need.


.github/ISSUE_TEMPLATE/config.yml
```diff
@@ -0,0 +1,8 @@
+blank_issues_enabled: false
+contact_links:
+  - name: Ragna documentation
+    url: https://docs.ragna.chat/
+    about: Check out the Ragna documentation
+  - name: Blank issue (maintainers only)
```
pmeier, line 6: Do L1 and L6 conflict?

.github/ISSUE_TEMPLATE/feature-request.yml
```diff
@@ -0,0 +1,36 @@
...
+        Are you having any problems? Briefly describe what your painpoints are. For example: "I'm always frustrated when ..."
+    validations:
+      required: true
+
+  - type: textarea
+    attributes:
+      label: Value and/or benefit
+      description: |
+        What is the value of adding this feature, and who will benefit from it?
+        Include any information that could help us prioritize the issue.
```
pmeier, line 26: I'm always frustrating to see these two separated from each other. If I actually write something like "I'm always frustrated when ..." in the first, the second from just becomes "The benefit is that I'm no longer frustrated".

What value do you think we'll have by having two fields here?

.github/ISSUE_TEMPLATE/support.yml
```diff
@@ -0,0 +1,37 @@
+name: "Usage question 🙋🏽‍♀️"
+description: "Create a report to help us fix something that is currently broken."
+title: "[Q/A] - <title>"
+labels: ["type: question 🤔"]
```
pmeier, line 4: Can we just redirect this to GH discussions? The issue tracker is not well suited for usage questions.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,20 @@
+Closes # .
```
pmeier, line 1: The PR template is where I usually take the greatest offense. People are regularly ignore or simply delete the template. I feel we should make these things clear in the contribution guide rather than have this here and just add noise to the process.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,20 @@
...
+
+<!-- Reference the issue corresponding to this PR. If an issues does not exist, consider opening one or writing a detailed description for what this PR changes and it's value in the following sections. -->
```
pmeier, line 3: I'm favor of a stronger language (as explained above inside the contributing guidelines not as PR template). Something along the lines of "Unless it is a trivial bug fix or documentation improvements, please open an issue first discussing the change with us. Not doing so, risks us not accepting the PR and wasting your time."

.github/ISSUE_TEMPLATE/config.yml
pavithraes: No, it's an alternative chosen to be able to add addition: "maintainers only" text there

.github/ISSUE_TEMPLATE/feature-request.yml
pavithraes: It's separated to encourage contributors to think about this, and manage expectations if their request is not prioritized. While seasoned contributors include this information by default, there are many people who don't.

For example: If the only benefit is that it's slightly less frustrating to one user, it may not get worked on soon - but if they share that the feature will enable their team/project/community to adopt Ragna, we can try to prioritize it.

If it still feels unnecessary, we can remove it.

.github/ISSUE_TEMPLATE/support.yml
pavithraes: Yep, will update and enable discussions.

I created this because questions in this stage are likely to be bugs/feature-requests/documentation-blindspots, which can be easier to track with issues. I don't feel super strongly about this though, so happy to change. :)

.github/PULL_REQUEST_TEMPLATE.md
pavithraes: Thanks for sharing. While I do disagree here[^1], I'm happy to remove it for simplicity at this early stage. :)

[^1]: It's useful to have a template to remind folks about some core PR requirements, even with clear contribution docs.

.github/ISSUE_TEMPLATE/feature-request.yml
pavithraes: (Aside, we can simplify the description for the first prompt. Will update in next commit.)

.github/ISSUE_TEMPLATE/bug-report.yml
```diff
@@ -0,0 +1,49 @@
...
+    validations:
+      required: true
+
+  - type: textarea
+    attributes:
+      label: How to reproduce the bug?
+      description: |
+        Provide a minimal code example and detailed step-by-step guidelines to reproduce the error.
+    validations:
+      required: false
```
pmeier, line 29: Let's make this required if we already have a template. Bug report without repro is hard to act on. Worst case, someone could just put "N/A" or the like in there.

```suggestion
      required: true
```

.github/ISSUE_TEMPLATE/config.yml
pmeier: @pavithraes Are you sure? The contact links do not show up:

image

pavithraes: I was wrong - we do need to omit the first line. Sorry!

PR #108, opened 2023-10-25
ruff everywhere

pmeier:
With the release of `ruff format`, we don't need anything else anymore :tada:


pavithraes (approved):
Thank you, this is quite neat! ✨

PR #111, opened 2023-10-25
cleanup usort

pmeier:
Follow-up to #108. I missed that we need to use `# isort` directives to split import blocks. Plus, there was some configuration left for `usort` that is now obsolete.


pavithraes (approved):
LGTM!

PR #113, opened 2023-10-25
move public utils into vector db base class

pmeier:
The stuff in `ragna.utils` was exclusively used by our vector DB source storages. Now that we have a new and fancy base class for them (#107), we can move the utils there. This makes the vector DB implementation ever so slightly easier, but most importantly decreases our public API surface.


pavithraes (approved):
Tested `LanceDB` and `Chroma` locally, and the changes look good to me. Thank you!

PR #114, opened 2023-10-26
Fix blank issues display in template config

pavithraes:
ref: https://github.com/Quansight/ragna/pull/103#discussion_r1372837279

cc @pmeier


pavithraes:
Small PR, so self-merging. Let me know if there are any issues. :)

PR #117, opened 2023-10-26
increase test timeout

pmeier:
I'm seeing quite a few timeouts in #61. Using a too short timeout is recipe for flaky CI. Let's up it a little and hope that this is the end of it.


pavithraes (approved):
YES, was just going to open an issue about this! Thank you!! ✨

PR #123, opened 2023-10-27
forward subprocess STDOUT / STDERR to main process

pmeier:
Offline report from @dharhas. For simplicity when starting the UI, we also start the API as well as a worker if needed. However, we are not forwarding their log messages to the main thread. This makes it confusing for the user if an error happens, because there is trace of it in the console. This PR fixes that.


pmeier:
Merging this to ease debugging. Let's revert in case someone is against it.

Issue #124, opened 2023-10-27
Remove the special demo and builtin option for the config?


pmeier:
The original intention was to make it easier for the user to select these two without the need of a config file. However, it is beginning to dawn on me that this actually makes it more confusing:

By having `builtin` as default, starting the worker / API / UI without a config is possible. However, we are only including components for which the requirements are met. So if one does `pip install ragna[builtin]`, they would only see the builtin source storages, since all assistants so far require an API key to be set as environment variable. Worse, if one does `pip install ragna`, the default `builtin` configuration is the same as `demo`, since now we likely not even meet the package requirements.

By removing the `demo` and `builtin` option and in turn require the `-c` / `--config` option, we would "force" users through the config wizard. And by that they should have a better understand what they are running.

smeragoel:
I think that makes sense, since we want them to go through the config to get a better understanding.

pmeier:
Closed in #125.

PR #125, opened 2023-10-27
refactor config wizard

pmeier:
This is my proposed solution for #124. It implements the following changes:

- Remove the `demo` and `builtin` special configuration options. Note that the previous `Config.demo()` is still available, since that just created a `Config()` with default values anyway:
  https://github.com/Quansight/ragna/blob/a04efc2fbde3f44d541f6c49f84a92b4c91f2fc4/ragna/core/_config.py#L123-L125
- Remove the default value for `-c` / `--config`. `ragna worker`, `ragna api`, and `ragna ui` now require passing a configuration.
- Split `ragna config` and `ragna config --check` into `ragna init` and `ragna check`. `ragna init` is the configuration wizard and `ragna check` can be with configuration files to check availability of the requirements.
- Improve the wizard by printing a summary of the unmet requirements as well as instructions on how to meet them.
- A few phrasing improvements in the wizard.

### DIFF
--- a/ragna/_cli/config.py
+++ b/ragna/_cli/config.py
@@ -4,6 +4,7 @@ from pathlib import Path
 from types import ModuleType
 from typing import TYPE_CHECKING, Annotated, Iterable, Type, TypeVar, cast
 
+import pydantic
 import questionary
 import rich
 import typer
@@ -15,6 +16,7 @@ from ragna.core import (
     Config,
     EnvVarRequirement,
     PackageRequirement,
+    RagnaException,
     Requirement,
     SourceStorage,
 )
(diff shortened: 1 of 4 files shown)


pmeier:
One more change after an offline chat with @dharhas:

Instead of asking for confirmation after every single component that has unmet requirements, just assume that is what the user wants, since they have selected this in the step before. Since this PR also adds a summary of all the required packages and environment variables at the end, no information is lost through this.

We also discussed to have a "I just want everything that ragna has to offer" option, but I decided against it for now. Due to the change above, the selection is now only two questions. Coupled with the fact that you can toggle all options by pressing the `a` key when the choices are visible, IMO, we would make the process more complicated by adding another question. LMK if you feel different.

smeragoel:
Agreed, simpler is better. Two options should be fine, and instead of over-optimising now, we can wait for user feedback to iterate.

pmeier:
One more final adjustment after another offline chat with @dharhas:

We now use the config file located at `./ragna.toml` as default in case it exists. Meaning, if someone generates a config with `ragna init`, all the other commands work without explicitly setting the `-c` / `--config` option.

pmeier:
Doc error is unrelated:

> ERROR   -  mkdocstrings: Couldn't load inventory {'url': 'https://fastapi.tiangolo.com/objects.inv'} through PythonHandler.load_inventory: HTTP Error 500: Internal Server Error

The FastAPI site is down.

PR #132, opened 2023-10-29
fix chroma only retrieving one source

pmeier:
Offline report from @dharhas.

### DIFF
--- a/ragna/source_storages/_chroma.py
+++ b/ragna/source_storages/_chroma.py
@@ -27,7 +27,7 @@ class Chroma(VectorDatabaseSourceStorage):
 
         self._client = chromadb.Client(
             chromadb.config.Settings(
-                is_persistent=True,
+                # is_persistent=True,
                 persist_directory=str(self.config.local_cache_root / "chroma"),
                 anonymized_telemetry=False,
             )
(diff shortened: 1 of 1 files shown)


ragna/source_storages/_chroma.py
```diff
@@ -96,19 +96,19 @@ def retrieve(
             include=["distances", "metadatas", "documents"],
         )
 
-        num_results = len(result["ids"])
+        num_results = len(result["ids"][0])
```
pmeier, line 99: This is the actual fix. For some reason, `result["ids"]` is a list, which contains nothing but another list, which ultimately holds the IDs we are after here.

ragna/source_storages/_chroma.py
```diff
@@ -96,19 +96,19 @@ def retrieve(
...
         result = {
             key: [None] * num_results if value is None else value[0]  # type: ignore[index]
             for key, value in result.items()
         }
         # dict of lists -> list of dicts
         results = [
-            {key: value[idx] for key, value in result.items()}
+            {key[:-1]: value[idx] for key, value in result.items()}
```
pmeier, line 106: Ignoring the last character here drops the "s" and thus makes the keys singular. Might sound nitpicky here, but I found myself confused when looking at the source again and having something like `id=result["ids"]` in the `Source` instantiation below.

dharhas (approved):

PR #134, opened 2023-10-30
Recommend [builtin] for installation

### LINKED ISSUE #126: Improve installation instructions
Documentation currently only states

https://github.com/Quansight/ragna/blob/a04efc2fbde3f44d541f6c49f84a92b4c91f2fc4/docs/install.md?plain=1#L15-L19

with no indication of what this means and why it is needed.

pavithraes:
Closes #126

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -18,7 +18,9 @@ If you want to install a minimal version[^1]:
 pip install ragna
 ```
 
-[^1]: The minimal version is for users interested in writing plugins to extend Ragna.
+[^1]:
+    The minimal version is for users who want fine-grained control over the dependencies
+    needed for the builtin components.
 
 And, upgrade to latest versions with:
 
--- a/docs/install.md
+++ b/docs/install.md
@@ -9,7 +9,7 @@ You need Python 3.9 or above in your working environment to use Ragna.
 You can install Ragna and all recommended dependencies with:
 
 ```bash
-pip install 'ragna[builtin]'
+pip install 'ragna[all]'
 ```
 
 If you want to install a minimal version[^1]:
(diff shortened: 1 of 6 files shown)


docs/install.md
```diff
@@ -6,18 +6,20 @@ You need Python 3.9 or above in your working environment to use Ragna.
...
 
-If you want to install all optional dependencies, append `[builtin]`:
+If you want to install a minimal version[^1]:
 
 ```bash
-pip install 'ragna[builtin]'
+pip install ragna
 ```
 
+[^1]: The minimal version is for users interested in writing plugins to extend Ragna.
```
pmeier, line 21: You can do that with the builtin installation as well. If we want a footnote here, it should be something like "the minimal version is for users that want fine-grained control over the dependencies needed for the builtin components"

pmeier:
@pavithraes As per offline opinion poll, it seems that most people prefer the term "all" over "builtin". Do we want to do this in this PR as well? If so, apart from all the docs changes, you need to change it in few more places:

- https://github.com/Quansight/ragna/blob/dd9ead1a1f822590bdb4e43e06cb2fd87c8f94d1/pyproject.toml#L55
- https://github.com/Quansight/ragna/blob/dd9ead1a1f822590bdb4e43e06cb2fd87c8f94d1/scripts/update_optional_dependencies.py#L64
- https://github.com/Quansight/ragna/blob/dd9ead1a1f822590bdb4e43e06cb2fd87c8f94d1/.github/actions/setup-env/action.yml#L66

pmeier (approved):
I've fixed my comments. Thanks for the base @pavithraes!

PR #135, opened 2023-10-30
Auth + async answer + loading indicator in placeholder

pierrotsmnrd:
Async answer for a better, non-blocking UX when asking a question.

Making it async revealed the default loading indicator of Panel's chat message, so I replaced it with one with Ragna's colors.

UPDATE : also contains Auth after a rebase, see PR #137

### DIFF
--- a/ragna/_ui/api_wrapper.py
+++ b/ragna/_ui/api_wrapper.py
@@ -19,9 +19,24 @@ class ApiWrapper(param.Parameterized):
     def __init__(self, api_url, **params):
         self.client = httpx.AsyncClient(base_url=api_url)
 
-        httpx.get(api_url).raise_for_status()
-
         super().__init__(**params)
+
+        try:
+            # If no auth token is provided, we use the API base URL and only test the API is up.
+            # else, we test the API is up *and* the token is valid.
+            endpoint = (
+                api_url + "/components" if self.auth_token is not None else api_url
+            )
+            httpx.get(
+                endpoint, headers={"Authorization": f"Bearer {self.auth_token}"}
+            ).raise_for_status()
+
+        except httpx.HTTPStatusError as e:
+            # unauthorized - the token is invalid
+            if e.response.status_code == 401:
+                raise RagnaAuthTokenExpiredException("Unauthorized")
+            else:
+                raise e
 
     async def auth(self, username, password):
         self.auth_token = (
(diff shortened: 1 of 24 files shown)


ragna/_ui/api_wrapper.py
```diff
@@ -61,21 +61,28 @@ def get_chats(self):
...
         return json_data
 
-    def answer(self, chat_id, prompt):
-        json_data = (
-            self.client.post(
-                f"/chats/{chat_id}/answer", params={"prompt": prompt}, timeout=None
+    async def answer(self, chat_id, prompt):
+        async with httpx.AsyncClient(
+            base_url=self.client.base_url, headers=self.client.headers
+        ) as async_client:
```
pmeier, line 67: How much work is it to make all functions async and only use a `AsyncClient` for this whole class?

ragna/_ui/central_view.py
```diff
@@ -143,12 +143,12 @@ def __init__(self, msg_data, user, on_click_source_info_callback=None):
...
         params["avatar"] = RagnaChatMessage.get_avatar(self.role, user)
 
-        super().__init__(**params)
+        super().__init__(**(params | kwargs))
 
         self.update_css_classes()
         self.chat_copy_icon.visible = False
 
-        if self.role != "user":
+        if self.role not in ("user", "system"):
```
pmeier, line 151:
```suggestion
        if self.role == "assistant":
```

ragna/_ui/api_wrapper.py
pierrotsmnrd: There's one pain point : the check on the API when building the object, that leads to add async/await to the `__init__`, which isn't proper working code.
I suggest to move this check out of the init. 
I also suggest we merge #137 first to reduce conflicts on `api_wrapper.py`

pmeier: > There's one pain point : the check on the API when building the object, that leads to add async/await to the `__init__`, which isn't proper working code.

We could just use a sync client for that (or rather just use `httpx.post` since we won't reuse the client afterwards), since at this point there is no UI yet that we could render unresponsive. Plus, blocking until we actually have a proper connection to the API is probably a good thing anyway.

pierrotsmnrd: We will need the sync client at other places, because there are Panel functions that donc support being async. I'm on it.

pmeier:
@pierrotsmnrd I've pushed 094105c8293b9987e93c4b43e0ddaa072c112a80 to get rid of all the sync API calls. Unfortunately, this PR seems to be mixed with #137 and I'm facing issues there. Meaning, I can't properly review this one.

pierrotsmnrd:
Thanks for that, that's the piece I have been looking for. I wish `asyncio`'s docs were clearer.

ragna/_ui/api_wrapper.py
```diff
@@ -1,70 +1,70 @@
...
+    async def get_chats(self):
+        try:
+            json_data = (await self.client.get("/chats")).raise_for_status().json()
+            for chat in json_data:
+                chat["messages"] = [
+                    self.improve_message(msg) for msg in chat["messages"]
+                ]
+            return json_data
+
+        except httpx.HTTPStatusError as e:
```
pmeier, line 53: Didn't we want to do that in `__init__` and as part of getting the components?

ragna/_ui/main_page.py
```diff
@@ -104,10 +111,17 @@ def update_subviews_current_chat_id(self, avoid_senders=[]):
         if self.left_sidebar is not None and self.left_sidebar not in avoid_senders:
             self.left_sidebar.current_chat_id = self.current_chat_id
 
-    def page(self):
+    def __panel__(self):
+        try:
+            asyncio.ensure_future(self.refresh_data())
+        except RagnaAuthTokenExpiredException:
+            print("token expired, redirect logout")
```
pmeier, line 118:
```suggestion
```

ragna/_ui/api_wrapper.py
pierrotsmnrd: Done in my last commit

ragna/_ui/main_page.py
pierrotsmnrd: Removed in my last commit

pmeier (approved):
One minor comment for simplification, but we can do that in a follow-up PR. Thanks Pierre!

Issue #136, opened 2023-10-31
[BUG] - Don't use fastapi Depends for database sessions


costrouc:
### Bug description

I just saw https://github.com/Quansight/ragna/blob/main/ragna/_api/core.py#L104. This is the exact issue that we had in conda-store that Natika found. Depends does not reliably terminate the session when the request ends. FastAPI doesn't guarantee when the depends statement will be terminated. In conda-store the fix was to create the session explicitly with a context manager or middleware. Deep dive in this PR https://github.com/conda-incubator/conda-store/pull/622

### How to reproduce the bug?

Open a bunch of api requests concurently and database connection pool will become exhausted (I don't have a test for this) just can report this issue happended in conda-store.

pmeier:
Thanks for the report. This comes straight from the [FastAPI tutorial](https://fastapi.tiangolo.com/tutorial/sql-databases/#create-a-dependency) on how to use SQL databases https://github.com/tiangolo/fastapi/blob/1c25e2d8dc99331290e99e17814ec4d83adce725/docs_src/sql_databases/sql_app_py39/main.py#L12-L18

pmeier:
I did a deep dive. The dependencies handling happens [here](https://github.com/tiangolo/fastapi/blob/1c25e2d8dc99331290e99e17814ec4d83adce725/fastapi/dependencies/utils.py#L516) and the actual call [here](https://github.com/tiangolo/fastapi/blob/1c25e2d8dc99331290e99e17814ec4d83adce725/fastapi/dependencies/utils.py#L506-L513).

> ```python
> async def solve_generator(
>     *, call: Callable[..., Any], stack: AsyncExitStack, sub_values: Dict[str, Any]
> ) -> Any:
>     if is_gen_callable(call):
>         cm = contextmanager_in_threadpool(contextmanager(call)(**sub_values))
>     elif is_async_gen_callable(call):
>         cm = asynccontextmanager(call)(**sub_values)
>     return await stack.enter_async_context(cm)
> ```

So FastAPI does wrap the dependencies into a context manager and thus should terminate the session correctly.

nkaretnikov:
@pmeier I haven't looked at what ragna does. But here's context from conda-store:
- it's a known issue with FastAPI and dependencies, the docs might be out of date
[...]

PR #138, opened 2023-10-31
dont create DB session as dependency

pmeier:
This is a "better safe than sorry" PR that addresses #136. I still need to do a proper investigation if this is actually needed. However, the solution in this PR works flawlessly as reported by other users that faced the same issue. Thus, in interest of time, let's go with this for now, even it if its slightly less elegant.


nkaretnikov (approved):
Seems fine. I've only done a surface-level review. If there are problems later, I'd look into how make_session() is defined.

PR #140, opened 2023-10-31
only display unmet packages and env vars of the selected components

pmeier:


ragna/_cli/config.py
```diff
@@ -180,54 +180,56 @@ def _select_components(
 
 def _handle_unmet_requirements(components: Iterable[Type[Component]]) -> None:
     unmet_requirements = set(
-        itertools.chain.from_iterable(
-            component.requirements() for component in components
-        )
+        requirement
+        for component in components
+        for requirement in component.requirements()
+        if not requirement.is_available()
```
pmeier, line 186: This is the actual fix. The diff below is just GH refusing to render it properly. There is no change below other than unindenting one level.

PR #141, opened 2023-11-01
allow API token secret to be set through env var

pmeier:
Without this, restarting the API will invalidate all active tokens up to this point, since they API has a new secret to verify them. The old default behavior is preserved. Meaning, if you don't set the `RAGNA_DEMO_AUTHENTICATION_SECRET` env var, we'll use a random on that changes with every start of the API.

This also removes the option to set the document upload secret for local documents through the configuration file. We shouldn't have done that in the first place. It can still be set through the same env var `RAGNA_API_DOCUMENT_UPLOAD_SECRET` as before.

### DIFF
--- a/ragna/core/_authentication.py
+++ b/ragna/core/_authentication.py
@@ -49,10 +49,11 @@ class RagnaDemoAuthentication(Authentication):
     def __init__(self) -> None:
         self._password = os.environ.get("RAGNA_DEMO_AUTHENTICATION_PASSWORD")
 
-    _JWT_SECRET = secrets.token_urlsafe(32)
+    _JWT_SECRET = os.environ.get(
+        "RAGNA_DEMO_AUTHENTICATION_SECRET", secrets.token_urlsafe(32)[:32]
+    )
     _JWT_ALGORITHM = "HS256"
-
-    _ONE_WEEK = 60 * 60 * 24 * 7
+    _JWT_TTL = int(os.environ.get("RAGNA_DEMO_AUTHENTICATION_TTL", 60 * 60 * 24 * 7))
 
     async def create_token(self, request: Request) -> str:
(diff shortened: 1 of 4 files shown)


pierrotsmnrd (approved):
Looks good to me. tested locally

ragna/core/_authentication.py
```diff
@@ -49,10 +49,11 @@ class RagnaDemoAuthentication(Authentication):
     def __init__(self) -> None:
         self._password = os.environ.get("RAGNA_DEMO_AUTHENTICATION_PASSWORD")
 
-    _JWT_SECRET = secrets.token_urlsafe(32)
+    _JWT_SECRET = os.environ.get(
+        "RAGNA_DEMO_AUTHENTICATION_SECRET", secrets.token_urlsafe(32)[:32]
```
pierrotsmnrd, line 53: I might have approved a bit too fast :) 
Why is there a truncation over the last 32 chars only on ` secrets.token_urlsafe(32)[:32] ` ? 
Was it maybe supposed to be something like this ?
```python 
os.environ.get(
        "RAGNA_DEMO_AUTHENTICATION_SECRET", secrets.token_urlsafe(32)
)[:32]
```

pmeier: The number passed to `token_urlsafe` is the number of bytes to generate. This is greater than the number of characters:

```pycon
>>> import secrets
>>> len(secrets.token_urlsafe(32))
43
```

Meaning, if we want 32 characters for the token, we have to slice the returned secret.

pierrotsmnrd: Ok got it, its a matter of number of bytes that can result from 1 to 3 chars : 
https://docs.python.org/fr/3.8/library/secrets.html

PR #143, opened 2023-11-01
add plausible analytics

pmeier:
@pavithraes I'm merging this on my own to have the analytics now. Let's revert if you disagree.


pmeier:
Can confirm that the analytics is working. I see that there is also an option to add a feedback widget: https://material-plausible-plugin.ale.sh/#feedback-widget. Not sure if we want that, but open to discuss.

PR #149, opened 2023-11-02
Make API / UI origins configurable

### LINKED ISSUE #130: Add UI origin URL to config
Currently we have

https://github.com/Quansight/ragna/blob/47e9105b422b5b4887af289280a20be13c61757e/ragna/core/_config.py#L75-L79

which is used for the API

https://github.com/Quansight/ragna/blob/47e9105b422b5b4887af289280a20be13c61757e/ragna/_api/core.py#L23-L29

as well as the UI

https://github.com/Quansight/ragna/blob/47e9105b422b5b4887af289280a20be13c61757e/ragna/_ui/app.py#L100-L102

However, this is not feasible in an actual deployment as this are actually two separate concepts:

1. An URL that the UI binds to, usually something like `http://0.0.0.0:31477`. This is usually internal and not exposed to the users.
2. One or multiple URLs that connect to URL above. These are usually the ones users connect to, e.g. `ragna.my_company.com`.

I would keep URL for 1. and introduce a `config.ui.origins: list[str]` for 2.

pmeier:
Closes #130.

### DIFF
--- a/ragna/_api/core.py
+++ b/ragna/_api/core.py
@@ -9,7 +9,7 @@ from fastapi.responses import JSONResponse
 
 import ragna
 import ragna.core
-from ragna._utils import get_origins
+from ragna._utils import handle_localhost_origins
 from ragna.core import Config, Rag, RagnaException
 from ragna.core._components import Component
 from ragna.core._rag import SpecialChatParams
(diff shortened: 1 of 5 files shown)


pmeier:
Merging without review to move forward.

PR #151, opened 2023-11-02
Add testing for docs

### LINKED ISSUE #148: Add tests for code in the documentation
Issues like #144 highlight why we shouldn't have code in parts of the static, unchecked documentation. The code evolves and documentation gets outdated quickly. Sometimes the errors even happen when the code is put in initially, because typos happen and human reviewers don't always catch them.

We need a way to test the code that we put into our documentation. I think there are two main ways forward:

1. Add some kind of test for docs that is run automatically as part of the CI. I don't have experience with this.
2. Use something like [`mkdocs-gallery`](https://smarie.github.io/mkdocs-gallery/) to create the documentation at runtime. I have an (outdated) example PR #89 to showcase this.

pavithraes:
Closes #148 

- [x] Use https://github.com/koaning/mktestdocs - I think this does what we need, and the author is an active PyData community member
- [ ] Fix docs to pass the tests - tracking separately in https://github.com/Quansight/ragna/issues/153
- [x] Add to CI 

Note - this will currently fail on `tutorials/rest-api.md` -- working on it


docs/tutorials/python-api.md
```diff
@@ -35,7 +35,7 @@ The first step is to setup the configuration for components like the source stor
 For this minimal tutorial on basics, start with the default configuration:
 
 ```python
-from ragna import Config
+from ragna.core import Config
```
pmeier, line 38: This should not be needed. We expose `Config` form the main namespace.

.gitignore
```diff
@@ -167,3 +164,8 @@ cython_debug/
 
 # VSCode
 .vscode/
+
+# Ragna
+ragna.toml
+ragna/_version.py
```
pmeier, line 170: Can we keep this at the top please? These are Ragna specific things and should be easily discoverable.

.gitignore
```diff
@@ -167,3 +164,8 @@ cython_debug/
...
+ragna.txt
```
pmeier, line 171: Is this also generated by the tests? If so, can we maybe configure the test to a use a tmp directory? As is, this seems kinda arbitrary.

docs/how-tos/set-configuration.md
```diff
@@ -45,6 +45,11 @@ url = "http://127.0.0.1:31477"
 
 You can use `ragna.toml` for setting configurations in your applications:
 
+<!--
+Using `py``` instesd of `python`` allows for syntax highlighting without doctesting.
+This is a work around until https://githsub.com/koaning/mktestdocs/issues/7 is implemented.
```
pmeier, line 50:
```suggestion
This is a work around until https://github.com/koaning/mktestdocs/issues/7 is implemented.
```

docs/tutorials/python-api.md
```diff
@@ -132,16 +131,18 @@ the `async` and `await` keywords with the function definition and call respectiv
...
-async with rag.chat(
-    documents=[path],
-    source_storage=RagnaDemoSourceStorage,
-    assistant=RagnaDemoAssistant,
-) as chat:
-    prompt = "What is Ragna?"
-    answer = await chat.answer(prompt)
+```python
+
+async def create_chat():
```
pmeier, line 136: "Create chat" is not sufficient, because we also prompt for an answer.

```suggestion
async def main():
```

pmeier (approved):
Few comments, but otherwise LGTM for an initial setup. Thanks @pavithraes!

pmeier (approved):
Fixed my own comments

PR #155, opened 2023-11-02
config quick fix : disable sliders if "Ragna/Demo"

pierrotsmnrd:
This PR is a quick fix to disable sliders in advanced config when the assistant or the source storage has "Ragna/Demo" in it.

Later we'll generate the advanced config dynamically based on the API result for the "components" endpoint (I actually have quite advanced work on that but I have a bug in Panel that prevents me from delivering a proper PR)


pmeier (approved):
Thanks Pierre!

PR #157, opened 2023-11-02
add print for demo auth

pmeier:
It might still be non-obvious how to login, but this is a start.


pmeier:
Self-merging since this is minor.

PR #164, opened 2023-11-03
Add announcement banner with link to blog

pavithraes:
Ref:

Screenshot 2023-11-03 at 5 12 01 PM

It will be hidden on scroll, and can be dismissed


pavithraes:
actually, one more update - just a sec!

pmeier (approved):
LGTM, thanks Pavithra!

PR #168, opened 2023-11-07
document default login credentials

pmeier:
Per default, since this has popped up multiple times now:

- #167
- https://github.com/Quansight/ragna/discussions/156#discussioncomment-7460848

This also activates [`git lfs`]() for `*.png` images to avoid tracking them as binary blobs inside the repository. In the diff it looks like the old images are now broken, but you can see that everything works as expected when looking at the [branch](https://github.com/Quansight/ragna/tree/ui-login-doc) directly.


pavithraes (approved):
Looks good to me, thank you! 🌻

Issue #75, opened 2023-10-16
Reinitialising as single-threaded warning on import


pmeier:
```shell
$ python -c "import ragna"
platform/c++/implementation/internal.cpp:205:reinit_singlethreaded(): Reinitialising as single-threaded.
```

This is coming from `pymupdf`

```shell
$ python -c "import fitz_new"
platform/c++/implementation/internal.cpp:205:reinit_singlethreaded(): Reinitialising as single-threaded.
```

Since this comes from C++, there is no way of disabling it. One of the devs [stated](https://github.com/pymupdf/PyMuPDF/issues/2742#issuecomment-1764190052) that this will be fixed in a future release.

PR #169, opened 2023-11-07
pin pymupdf to latest release

pmeier:
The latest release removes the warning on import that we saw in #75.


pmeier:
Self-merging to get CI back to green.

Issue #171, opened 2023-11-08
[ENH] - Add support for Amazon Bedrock


094459:
### Feature description

I am looking to see how I can add support for Amazon Bedrock, but trying to figure out the process for adding new assistants. I am happy to give this a go myself, but at the moment, the code I have put together is not working, but i think that is because I am not sure how Ragna works with assistants. 

### Value and/or benefit

Open this up to users who are accessing some of these foundational models via Amazon Bedrock rather than the providers (e.g. Anthropic)

pmeier:
Hey @094459 and thanks for the request. Our general policy (we haven't written it down :smiling_face_with_tear: ) is that we won't accept any components that require additional infrastructure to run. Packages installable from PyPI or secrets in the form of environment variables are fine.

That being said, we of course recognize the need for services that require additional infrastructure. This is why we build Ragna to be fully extensible. I acknowledge that it is abysmally documented right now. Just subclass [`SourceStorage`](https://ragna.chat/en/stable/references/python-api/#ragna.core.SourceStorage) and [`Assistant`](https://ragna.chat/en/stable/references/python-api/#ragna.core.Assistant) and implement whatever you like. As long as it adheres to the signatures of the abstract methods, it will work with Ragna. 

For the Python API, you can pass your classes directly to the chat. Have a look at this notebook: https://github.com/Quansight/ragna/blob/main/examples/local_llm/local_llm.ipynb

For the REST API and UI, you need to add them to the `source_storages = ` and `assistants = ` list in the configuration. For example

```toml
[core]
# ...
source_storages = [
    "ragna.source_storages.RagnaDemoSourceStorage",
    "my_ragna_plugin.MySourceStorage",
]
assistants = [
    "ragna.assistants.RagnaDemoAssistant",
    "my_ragna_plugin.MyAssistant",
]
```

Make sure that that the `my_ragna_plugin.py` Python module is on the `PYTHONPATH` when starting the API / UI. Just as heads up: to meet the deadline, we hardcoded some assumptions about the source storages in the UI. This will be cleaned up for the next release, but might conflict with your custom classes for now.

094459:
It is only the additional assistant I was looking to add, specifically Amazon Bedrock's Anthropic Claude v2. I will take a look at the info and see what I can figure out. Thanks for the response.

094459:
https://github.com/Quansight/ragna/pull/173

pmeier:
Closing this following the discussion in #173.

PR #173, opened 2023-11-08
Add Amazon Bedrock Anthropic Claude v1 and v2 assistants

094459:
Additional assistant if you want to use Claudev1 or v2 hosted via Amazon Bedrock.

You need to set AWS environment variables for your AWS credentials (i.e export AWS_ACCESS_KEY_ID, AWS_SECRET_ACCESS_KEY)

Code is currently region specific to where Amazon Bedrock is hosted.

### DIFF
--- a/ragna/assistants/_bedrock.py
+++ b/ragna/assistants/_bedrock.py
@@ -1,0 +1,79 @@
+from typing import cast
+import boto3
+import json
+from ragna.core import RagnaException, Source
+
+from ._api import ApiAssistant
+
+
+class AmazonBedrockAssistant(ApiAssistant):
+    _API_KEY_ENV_VAR = "AWS_ACCESS_KEY_ID"
+    _MODEL: str
+    _CONTEXT_SIZE: int
+
+    @classmethod
+    def display_name(cls) -> str:
+        return f"AmazonBedRock/{cls._MODEL}"
+
+    @property
+    def max_input_size(self) -> int:
+        return self._CONTEXT_SIZE
+
+    def _instructize_prompt(self, prompt: str, sources: list[Source]) -> str:
+        instruction = (
+            "\n\nHuman: "
+            "Use the following pieces of context to answer the question at the end. "
+            "If you don't know the answer, just say so. Don't try to make up an answer.\n"
+        )
+
+        instruction += "\n\n".join(source.content for source in sources)
+        return f"{instruction}\n\nQuestion: {prompt}\n\nAssistant:"
+
+    def _call_api(
+        self, prompt: str, sources: list[Source], *, max_new_tokens: int
+    ) -> str:
+        bedrock = boto3.client(service_name='bedrock-runtime',
+                               region_name='eu-central-1')
+        # See https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_InvokeModel.html
+        prompt_config = {
+            "prompt": self._instructize_prompt(prompt, sources),
+            "max_tokens_to_sample": max_new_tokens,
+            "temperature": 0.0
+            }
+        
+        try:
+            response = bedrock.invoke_model(
+            body=json.dumps(prompt_config),
+            modelId=f"anthropic.{self._MODEL}"
+            )
+        
+            response_body = json.loads(response.get("body").read())
+        
+            return cast(str, response_body.get("completion"))
+        except Exception as e:
+            raise ValueError(f"Error raised by inference endpoint: {e}")
+    
+
+
+class AmazonBedRockClaude(AmazonBedrockAssistant):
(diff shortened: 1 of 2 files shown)


094459:
This PR is for https://github.com/Quansight/ragna/issues/171

ragna/assistants/_bedrock.py
```diff
@@ -0,0 +1,79 @@
+from typing import cast
+import boto3
```
pmeier, line 2: Ignore this if you go for a third party package that can hard-depend on `boto3`. Otherwise, this should be a requirement on the class below and imported locally to avoid introducing a hard runtime dependency.

ragna/assistants/_bedrock.py
```diff
@@ -0,0 +1,79 @@
...
+import json
+from ragna.core import RagnaException, Source
+
+from ._api import ApiAssistant
+
+
+class AmazonBedrockAssistant(ApiAssistant):
+    _API_KEY_ENV_VAR = "AWS_ACCESS_KEY_ID"
```
pmeier, line 10: Although we set it here, it is never used anywhere. That hints to the fact that `ApiAssistant` is probably not the right abstraction and you are better off subclassing from `ragna.core.Assistant` directly.

ragna/assistants/_bedrock.py
```diff
@@ -0,0 +1,79 @@
...
+
+    def _instructize_prompt(self, prompt: str, sources: list[Source]) -> str:
+        instruction = (
+            "\n\nHuman: "
+            "Use the following pieces of context to answer the question at the end. "
+            "If you don't know the answer, just say so. Don't try to make up an answer.\n"
+        )
+
+        instruction += "\n\n".join(source.content for source in sources)
+        return f"{instruction}\n\nQuestion: {prompt}\n\nAssistant:"
```
pmeier, line 30: This prompt is special for Anthropic, although it should also be working for other assistants.

ragna/assistants/_bedrock.py
```diff
@@ -0,0 +1,79 @@
...
+            response_body = json.loads(response.get("body").read())
+        
+            return cast(str, response_body.get("completion"))
+        except Exception as e:
+            raise ValueError(f"Error raised by inference endpoint: {e}")
+    
+
+
+class AmazonBedRockClaude(AmazonBedrockAssistant):
+    """[Amazon Bedrock Claud v2](https://eu-central-1.console.aws.amazon.com/bedrock/home?region=eu-central-1#/providers?model=anthropic.claude-v2)
```
pmeier, line 59: That is an internal link that requires an AWS account.

pmeier (commented):
@094459 it seems there was a misunderstanding. To quote myself from https://github.com/Quansight/ragna/issues/171#issuecomment-1801713734

> we won't accept any components that require additional infrastructure to run.

AWS Bedrock certainly fits that bill. We had some initial ideas to provide a `ragna.contrib` namespace, but scrapped that idea again in favor of community maintained third-party packages. 

Seeing that commit is authored by @ricsue-aws a potential solution would be to create a `ragna-aws` package that holds configuration for Bedrock assistants. Plus, it could also include something like the S3 document that we have in https://github.com/Quansight/ragna/blob/main/examples/s3_documents/ragna_s3_document.py.

That being said, regardless of where the implementation is ultimately put, it needs to somehow find its way into a configuration file. @094459 could you explain your reasoning why this should be part of the Ragna core library rather than just a local extension that you wrote?

094459:
Hi there. I have no strong opinions where this goes, and only wanted to contribute if you found it useful. I guess I was confused by the term 'additional infrastructure' given that Bedrock acts kind of like Anthropic, OpenAI, so apologies for the misunderstanding. I put this together for folks who might want to try this out but be using Claude via Bedrock rather than directly via Anthropic. Hope this makes sense.

094459:
I'm not sure what a local extension means but if you mean a local fork then if that is the recommendation of the project then I have no issues with that.

pmeier:
> apologies for the misunderstanding.

No worries here.

> I put this together for folks who might want to try this out but be using Claude via Bedrock rather than directly via Anthropic. Hope this makes sense.

Yeah, and that is fine. Like I stated above, maybe we can have a `ragna-aws` package in the future that contains just this functionality. However, since we won't accept the PR as is in the near future, I'm going to close this. If our stance on this changes, we can always re-open.

> I'm not sure what a local extension means but if you mean a local fork then if that is the recommendation of the project then I have no issues with that.

What I mean is that you can use the implementation you have with Ragna without it being part of the core package. Just throw your implementation in a `ragna_bedrock.py` module (make sure that is on the PYTHONPATH; you can test with `python -c import ragna_bedrock.py`) and add this to your configuration file

```toml
[core]
# ...
assistants = [
    # ...
    "ragna_bedrock.AmazonBedRockClaude",
    "ragna_bedrock.AmazonBedRockClaudev1",
]
```

If you start the UI now, you should be able to select and use them. If not, please open an issue.

Issue #174, opened 2023-11-08
Increase API assistant timeout


pmeier:
As reported in https://github.com/Quansight/ragna/issues/166#issuecomment-1801813313, 30 seconds timeout might still not be enough. I don't feel we should go to 90 seconds right away, but maybe meet in the middle at 60 seconds first. I have never had a request take more than 30 seconds, so tripling it might be a overkill. Remember that higher timeout also increase the time to signal if there is an actual timeout happening.

Timeout can be adjusted here:

https://github.com/Quansight/ragna/blob/62098457588cef258f4ea6a8b35989257559a27f/ragna/assistants/_api.py#L29

cc @nenb

petrpan26:
Can I take this issue. I saw that Open AI have streaming api, do we want in the future to use this feature to verify that we still waiting for output ?

pmeier:
@petrpan26 I've opened #185 to explain why streaming the response is a lot harder to implement than you might initially think.

PR #184, opened 2023-11-08
Update assistant timeout to accomodate longer response time.

petrpan26:
Relating to:
https://github.com/Quansight/ragna/issues/174

An improvement idea is that we can have streaming input check for those api that support this (e.g GPT). This way we know that it's still generating output rather than a network problem.

### DIFF
--- a/ragna/assistants/_api.py
+++ b/ragna/assistants/_api.py
@@ -26,7 +26,7 @@ class ApiAssistant(Assistant):
 
         self._client = httpx.Client(
             headers={"User-Agent": f"{ragna.__version__}/{self}"},
-            timeout=30,
+            timeout=60,
         )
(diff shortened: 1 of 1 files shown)


pmeier (approved):
Thanks for the fix @petrpan26!

> An improvement idea is that we can have streaming input check for those api that support this (e.g GPT).

Unfortunately, this is not really possible or will be a lot harder than you might think. I would very much like to have this feature and have played around with it in the past.

This is so hard, because we receive the response on the worker, but consume it on the API side. In between them is our result storage. Meaning, we would need to somehow implement a streaming approach for the results and finally also implement streaming on our own API.

If you want to have a go at this, be my guest. But be warned that this likely could turn really ugly.

PR #187, opened 2023-11-08
fix note in Ragna 101

pmeier:
before:

image

after:

image

Thanks @jaimergp for the report!


pmeier:
https://ragna--187.org.readthedocs.build/en/187/tutorials/python-api/#step-4-select-an-assistant-llm

Issue #189, opened 2023-11-09
[ENH] - Support Entry Points for Plugins


nenb:
### Feature description

Add support for [Entry Points](https://setuptools.pypa.io/en/latest/userguide/entry_point.html#entry-points-for-plugins) for Plugins to allow easy code-sharing eg for a new assistant in #173. 

### Value and/or benefit

Extensions can be easily shared - it's just a simple `pip` install. The current behaviour requires users to 'discover' an extension for themselves, and then to make sure that they install it correctly so that it's recognised by `ragna`.

### Anything else?

This has been used very successfully (IMO) by [`llm`](https://llm.datasette.io/en/stable/plugins/index.html).

Most of the hard work has already been done (ie the plug-in interfaces like `ApiAssistant`), there is just the final-step of adding some packagaing details and some brief logic related to plug-in discovery.

I can mock-up a quick example if this is of interest.

pmeier:
In general, I'm in favor of this. Like you said, it makes sharing really easy. One question arises though: where do we hook in external components? My current idea would be to do it as part of the component selection in the config wizard. We could just chain them to the builtin ones

https://github.com/Quansight/ragna/blob/dc497ef4768b8f8bd4d5c3736eb1d67b27050387/ragna/_cli/config.py#L161-L165

Of course we need to change the phrasing in the wizard to account for that. Plus, we probably need to put the class path as part of the selection (as well as `ragna check`) to make it clear where a component is coming from.

nenb:
I just spent a little time on this and made an attempt with the `pydantic2` [`model_post_init`](https://docs.pydantic.dev/latest/api/base_model/#pydantic.main.BaseModel.model_post_init) hook. I was reading through some of the comments in the code, and I realised that there are lots of ways for users to interact with the config settings (env vars, kwargs, TOML files etc). I settled on `model_post_init` as it felt like the most robust way to load the extensions. But perhaps it causes issues in other ways (and I think it does rely on `pydantic` version 2) - what is your opinion?

I also added a mock package and a notebook to show it works. All available in my branch [here](https://github.com/nenb/ragna/tree/entrypoints-plugin).

nenb:
I think that I may only have a partial understanding of the intended workflows. I'm going to be more explicit here about my current understanding. Please correct the bits that I have wrong (and then I will follow-up with a response to your previous questions):

**Workflow 1 -  UI entrypoint (`ragna ui`)**
[...]

PR #192, opened 2023-11-10
Entrypoints plugin

nenb:
This is an initial attempt at adding support for entry points for plugins following #189.

Current status: Question marks over the implementation and whether it is truly needed. See #189 for latest details.


pmeier:
@nenb I just released `ragna-aws`: https://github.com/pmeier/ragna-aws. I'm going to share a few pain points in form of issues soon. Forcing users to manually edit their `ragna.toml` to get the functionality they want is one of them. We can use the package as test bed for the entrypoints.

Issue #185, opened 2023-11-08
Streaming assistant responses


pmeier:
Some APIs like OpenAI and Anthropic support streaming the generated response. This is a really nice UX, because the user gets instant feedback and doesn't have to wait for a long time to receive a large blob of text.

However, due to Ragnas backend structure, this is either "impossible" (read infeasible for the upside) or really hard. The problem is that we receive the response on the worker, but consume it on the API side. In between them is our result storage. Meaning, we would need to somehow implement a streaming approach for the results and finally also implement streaming on our own API.

I really would like to have this feature and toyed with it in the past. Unfortunately, I gave up at some point for the reasons stated above. If someone wants to have a go at this, be my guest. But be warned that this likely will turn really ugly.

petrpan26:
It makes sense to me ! Streaming is definitely something that is hard. My initial thought is to used it as a additional check with timeout maybe at request level. My thought is for the client dont shut down the connection while they still fetching it. But I will revisit this when I have more familiarity with the code base. Is there any issue that you think I can help with ?

pmeier:
If we just want to circumvent the timeout on the worker side, we could still implement streaming there. It wouldn't make a difference to the user as we would only return if we have the full response, but since we should get an initial response super fast, we could maybe lower the timeout again.

nenb:
Does communicating with the LLM (`answer`) need to be done via the worker? I understand that the embeddings (`prepare`) absolutely need this. But communicating with the LLM is more of an I/O bound problem, and I've had no problem running it on the main thread (asynchronously of course) in previous applications.

This would then make it a lot easier to handle the streaming use-case (which I agree is absolutely essential for RAG). Their might need to be a few more changes to deal with the fact that the streamed response will return an `Iterable`, but I don't see this as being a major issue right now.

petrpan26:
I assume that's there can also be self hosted LLM on the computer ? But I agreed. Since the final output of Rag is a synthesis step it should be able to support streaming if the underlying LLM support it

pmeier:
@petrpan26 is right about the reason. We use the same abstraction for local LLMs as for hitting an external API. Thus, everything goes through the task queue.

@nenb

> is absolutely essential for RAG

I disagree here. Streaming is nice to have and not essential. Everything works perfectly fine without streaming.

nenb:
>@petrpan26 is right about the reason. We use the same abstraction for local LLMs as for hitting an external API. Thus, everything goes through the task queue.

[...]

PR #205, opened 2023-11-17
remove task queue

### LINKED ISSUE #183: Change the default queue implementation something other than memory
_originally reported by @nenb in https://github.com/Quansight/ragna/issues/166#issuecomment-1801813313_

- **I changed the default queue implementation from 'memory' to 'file'** - (I'm curious why 'file' was offered over `sqlite` here?). The main reason I did this is that the 'memory' implementation is currently blocking (see [here](https://github.com/Quansight/ragna/blob/62098457588cef258f4ea6a8b35989257559a27f/ragna/core/_queue.py#L148)) and this was a bit of a frustrating experience for me when OpenAI was taking almost 60 seconds sometimes to respond. I'm not familiar enough with `huey` to know if the in-memory implementation can also be adapted to work in async mode (probably not I guess), so I just changed the default setting to 'file' and things went much more smoothly.

### LINKED ISSUE #204: queueless Ragna?
While responding to all issues since the launch (thank you everyone that took the time to get in touch!) it became somewhat evident that the queue backend is confusing / limiting. 

In general, the task queue is used to scale deployments of Ragna. When auto scaling worker nodes on the queue length, that is all there is to do to load balance Ragnas backend, while keeping the REST API nodes constant (of cource they also need to be autoscaled at some point if you have a lot of concurrent users). Other options like a service mesh were a lot more complicated when we want to have one consistent REST API the user can hit extend with plugins.

Still, this doesn't answer why the queue is part of the Python API. Let me give you some historical context on how it came into being. In the very beginning, Ragna had no public Python API (#17). The REST API was the only real entrypoint for users. After I rewrote the project in #28, I failed to cleanly separate the two. Back then, the Python API even still had the database baked in (#65). I removed that in #67.

Meaning, right now, I can't come up with a good reason for why we have a task queue as part of the Python API. Not having it could potentially simplify the implementation and enable us to implement a few nice to have features, such as #185.

The only downside I can currently come up with is that parallelizing tasks the Ragna API will be harder in the future. Right now, you can simply select a non-memory task queue and start a worker and be done with it. Without a task queue the user has to implement their own parallelization scheme. That being said, since the Python API is mostly meant for experimentation, I'm not sure if this is a strong argument.

I'm going to start a branch soon to move the task queue to the Python API to see if there are more downsides that I'm currently not seeing. Will post updates here.

pmeier:
Closes #183, closes #204. TL;DR this PR gets rid of `huey` / the task queue. It has been the source of confusion, limited some nice to have features (looking at you #185), and overall made the implementation more complex.

Here are the highlights:

- Remove the task queue and `huey` as dependency
- ~Rename `Chat.prepare` and `Chat.answer` to `Chat.aprepare` and `Chat.aanswer` to better reflect their `async` nature~ See https://github.com/Quansight/ragna/pull/205#issuecomment-1817848193
- ~Re-add `Chat.prepare` and `Chat.answer` as synchronous versions of their `async` counterparts.~ See https://github.com/Quansight/ragna/pull/205#issuecomment-1817848193
- Add a new `ragna.local_root` function that functions as global configuration
- Remove the config as parameter for components. If they need local storage, the new `ragna.local_root` function should be used.
- Enable the ability to implement regular and `async` functions on the components. A regular function will now be run on a separate thread and thus keeping Ragna "async first"
- Create a `ragna.deploy` namespace and move the config, authentication, REST API, and UI implementation to it.
- Rename `config.core` to `config.components`
- Move authentication and document settings to the "global" config scope, i.e. `config.authentication` and `config.document`

I didn't bother to update the documentation properly as we need to have a complete overhaul before the next release anyway.


ragna/core/_rag.py
```diff
@@ -237,42 +278,64 @@ def _parse_documents(self, documents: Iterable[Any]) -> list[Document]:
...
-                component, action, args, self._unpacked_params[(component, action)]
+    def _run_async(self, fn: Awaitable[T]) -> T:
+        # This is the best way I could come up with to be able to run an async function
+        # synchronously when there is already an event-loop running. This happens for
+        # example inside a jupyter notebook. Without this, it would be impossible to use
+        # synchronous methods in such a scenario.
+        # FIXME: That being said, it feels kinda hacky and is also causing issues in the
+        #  test suite. If anyone knows a better way to achieve this, please send a PR :)
+        nest_asyncio.apply()
+        return asyncio.run(fn)  # type: ignore[arg-type]
```
pmeier, line 311: This is really a bummer :disappointed: We have this in the first place, since the actual functionality is implemented in our `async` endpoints, e.g. `aanswer`. To avoid re-implementing the same body for the sync counterpart, I just run the `async` endpoint here. However, we need to resort to monkey-patching (it is done with care in `nest_asyncio`, but ultimately that is what it is) to do this when another event loop is already running. This happens for example in a jupyter notebook and thus is critical feature.

pmeier:
@nenb Would you be able to test this PR and see if it works as expected. Just note that I didn't implement any streaming for #185 just yet. This will come in a follow-up PR.

ragna/core/_rag.py
```diff
@@ -237,42 +278,64 @@ def _parse_documents(self, documents: Iterable[Any]) -> list[Document]:
...
+            for fn, model in component_models.items()
         }
 
-    async def _enqueue(
-        self, component: Type[Component], action: str, *args: Any
-    ) -> Any:
-        try:
-            return await self._rag._queue.enqueue(
-                component, action, args, self._unpacked_params[(component, action)]
+    def _run_async(self, fn: Awaitable[T]) -> T:
```
pmeier, line 303: Ok, this whole function is a problem. We can't merge like this as it is causing issues on Windows and macOS.

pmeier:
Sleeping on this, I think we should not block the removal of the task queue over the issues with the sync endpoints. They were only suggested by me in https://github.com/Quansight/ragna/issues/183#issuecomment-1804834416 to make it more clear that there is blocking behavior. But with this PR, there is no blocking anymore. Even without workers, everything is async. Thus, we can leave the sync endpoints for later. I'll remove them again from this PR and open an issue to fix this later.

pmeier:
@nenb I'm merging this without waiting for your reply to move on to the other features that build on top of this. Since we are not adding any sync endpoints for now, I'm confident that I didn't break anything. Feel free to comment here or a new issue if you find something that I missed.

nenb:
@pmeier Nice work, sorry I haven't been terribly responsive the last few days, I should be back online from tomorrow. 

From looking through the PR now, the only bit that caught my eye was `anyio.to_thread.run_sync`. Some of the chunking/embedding can be quite CPU intensive, and I was wondering if it might be better to run in a separate process. I downloaded 5 books from Project Gutenberg in .txt format and tried to do a quick profile of threads vs processes for these 5 books. Unfortunately I got blocked on a bunch of serialisation issues (related to the `chromadb` ONNX format for their default model, and then related to `tiktoken`) and so I wasn't able to perform this comparison.

I'm mentioning this here for future reference. I know you have a bunch of work planned, and this is likely premature optimisation. But perhaps worth keeping in mind ways to mitigate the impact of CPU-heavy workloads in future designs. And also some of the defaults, like the values that are used when starting the API with `uvicorn`. Again, likely premature optimisation on my part!

pmeier:
> Unfortunately I got blocked on a bunch of serialisation issues (related to the `chromadb` ONNX format for their default model, and then related to `tiktoken`) and so I wasn't able to perform this comparison.

Do you think this is an issue with Ragna? If so, could you write up an issue so I can have a look?

> I'm mentioning this here for future reference. I know you have a bunch of work planned, and this is likely premature optimisation. But perhaps worth keeping in mind ways to mitigate the impact of CPU-heavy workloads in future designs. 

Yeah, I would put it under premature optimization unless we have valid complains about this. And in that case I would maybe go a step further and not just add the subprocess option, but a task queue would also fit nicely in there.

nenb:
>Do you think this is an issue with Ragna? If so, could you write up an issue so I can have a look?

I don't think it's an issue with `ragna`, but I will write up some details (along with a bunch of other things I haven't responded to) throughout today, so you can have a look.,

nenb:
@pmeier I have opened an issue [here](https://github.com/Quansight/ragna/issues/216) with details.

PR #215, opened 2023-11-20
implement streaming for assistants

pmeier:
Addresses the Python API part of #185.

### DIFF
--- a/environment-dev.yml
+++ b/environment-dev.yml
@@ -10,6 +10,7 @@ dependencies:
       - boto3
       - pytest >=6
       - pytest-mock
+      - pytest-asyncio
       - mypy ==1.6.1
       - pre-commit
       - types-aiofiles
--- a/environment-dev.yml
+++ b/environment-dev.yml
@@ -10,7 +10,6 @@ dependencies:
       - boto3
       - pytest >=6
       - pytest-mock
-      - pytest-asyncio
       - mypy ==1.6.1
       - pre-commit
       - types-aiofiles
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -37,6 +37,7 @@ dependencies = [
     "redis",
     "questionary",
     "rich",
+    "sse-starlette",
     "sqlalchemy>=2",
     "tomlkit",
     "typer",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -54,6 +54,7 @@ Repository = "https://github.com/Quansight/ragna"
 # to update the array below, run scripts/update_optional_dependencies.py
 all = [
     "chromadb>=0.4.13",
+    "httpx_sse",
     "lancedb>=0.2",
     "pyarrow",
     "pymupdf>=1.23.6",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -22,7 +22,6 @@ classifiers = [
 requires-python = ">=3.9"
 dependencies = [
     "aiofiles",
-    "anyio",
     "emoji",
     "fastapi",
     "httpx",
@@ -38,8 +37,9 @@ dependencies = [
     "redis",
     "questionary",
     "rich",
-    "sse-starlette",
     "sqlalchemy>=2",
+    "sse-starlette",
+    "starlette",
     "tomlkit",
     "typer",
     "uvicorn",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -26,6 +26,7 @@ dependencies = [
     "emoji",
     "fastapi",
     "httpx",
+    "httpx_sse",
     "importlib_metadata>=4.6; python_version<'3.10'",
     "packaging",
     "panel>=1.3,<1.4",
@@ -55,7 +56,6 @@ Repository = "https://github.com/Quansight/ragna"
 # to update the array below, run scripts/update_optional_dependencies.py
 all = [
     "chromadb>=0.4.13",
-    "httpx_sse",
     "lancedb>=0.2",
     "pyarrow",
     "pymupdf>=1.23.6",
--- a/ragna/_compat.py
+++ b/ragna/_compat.py
@@ -1,6 +1,14 @@
 import builtins
 import sys
-from typing import AsyncIterator, Callable, Iterable, Iterator, Mapping, TypeVar
+from typing import (
+    AsyncIterator,
+    Awaitable,
+    Callable,
+    Iterable,
+    Iterator,
+    Mapping,
+    TypeVar,
+)
 
 __all__ = ["itertools_pairwise", "importlib_metadata_package_distributions", "anext"]
 
@@ -41,7 +49,7 @@ def _importlib_metadata_package_distributions() -> (
 importlib_metadata_package_distributions = _importlib_metadata_package_distributions()
 
 
-def _anext() -> Callable[[AsyncIterator[T]], T]:
+def _anext() -> Callable[[AsyncIterator[T]], Awaitable[T]]:
     if sys.version_info[:2] >= (3, 10):
         anext = builtins.anext
     else:
--- a/ragna/_compat.py
+++ b/ragna/_compat.py
@@ -1,7 +1,8 @@
+import builtins
 import sys
-from typing import Callable, Iterable, Iterator, Mapping, TypeVar
+from typing import AsyncIterator, Callable, Iterable, Iterator, Mapping, TypeVar
 
-__all__ = ["itertools_pairwise", "importlib_metadata_package_distributions"]
+__all__ = ["itertools_pairwise", "importlib_metadata_package_distributions", "anext"]
 
 T = TypeVar("T")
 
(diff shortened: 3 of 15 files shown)


pmeier:
The current state implements the idea in https://github.com/Quansight/ragna/issues/185#issuecomment-1840399070. The UX is now as follows:

### `main`

```python
import asyncio
from ragna import Rag, assistants, source_storages

async def main():
    async with Rag().chat(
        documents=["ragna.txt"],
        source_storage=source_storages.Chroma,
        assistant=assistants.ClaudeInstant,
    ) as chat:
        message = await chat.answer("What is Ragna in ten words or less?")
        print(message)

asyncio.run(main())
```

```
 Ragna is an open source framework for building RAG-based AI applications.
```

### PR

```python
import asyncio
from ragna import Rag, assistants, source_storages

async def main():
    async with Rag().chat(
        documents=["ragna.txt"],
        source_storage=source_storages.Chroma,
        assistant=assistants.ClaudeInstant,
    ) as chat:
        message1 = await chat.answer("What is Ragna in ten words or less?")
        print(message1)
        print()
        print(await message1.content())
        print()
        print(message1)

        print("#" * 20)

        message2 = await chat.answer("What is Ragna in ten words or less?")
        print(message2)
        print()
        async for chunk in message2:
            print(chunk, end="")
        print()
        print()
        print(message2)

asyncio.run(main())
```

```
...

 Ragna is an open source framework for building RAG-based AI applications.

 Ragna is an open source framework for building RAG-based AI applications.
####################
...

 Ragna is an open source framework for building RAG-based AI applications.

 Ragna is an open source framework for building RAG-based AI applications.
```

So the trade-off for the ability to stream the answer is that user needs to await twice. Once to get the message object from the chat and a second time to get its content (either `await message.content()` or `async for chunk in message`). Printing the message when it hasn't been awaited yet currently just shows `...`, but that also could be a source on confusion.

I'm not a 100% happy with the result. One option might be to put a `stream: bool = False` parameter on `Chat.answer()` and unless that is `True`, i.e. the user explicitly needs to set it, we exhaust the stream inside the function and thus remove the requirement of awaiting again.

Thoughts @nenb?

pmeier:
One thing that also bugs me is that we no longer can access the `message.content` attribute from the outside. After the message is awaited, we can only access it, by "awaiting `Message.content()` again" (we'll just return the stored value) or by `Message.__str__`. I'll have another look at the response classes of HTTP clients on how they do it.

pmeier:
Ok, now I'm actually happy with it:

```python
import asyncio

from ragna import Rag, assistants, source_storages

import dotenv

assert dotenv.load_dotenv()


async def main():
    async with Rag().chat(
        documents=["ragna.txt"],
        source_storage=source_storages.Chroma,
        assistant=assistants.ClaudeInstant,
    ) as chat:
        print("Without stream=True")
        message = await chat.answer("What is Ragna in ten words or less?")
        print(message)

        print("#" * 20)

        print("With stream=True and iterating")
        message = await chat.answer("What is Ragna in ten words or less?", stream=True)

        try:
            print(message)
        except RuntimeError as error:
            print(f"{type(error).__name__}: {error}")

        async for chunk in message:
            print(chunk, end="")
        print()
        print(message)
        print(message.content)

        print("#" * 20)

        print("With stream=True but no iterating")

        message = await chat.answer("What is Ragna in ten words or less?", stream=True)
        print(await message.read())
        print(message)
        print(message.content)


asyncio.run(main())

**Comment by nenb** [comment]
@pmeier Nice work. The code snippet looks quite sensible and intuitive to me.

Let me know if you need me to look over the PR in more depth (but a quick skim seemed fine to me!)

**Comment by pmeier** [comment]
@peachkeel I could use some input from you on this. You rightfully advocated in #242 to trim the output of the answer REST API endpoint to avoid unnecessary data being sent. I'm currently in the process of deciding what the "message chunk" object looks like if one streams from answer endpoint. As of the latest commit we are using

https://github.com/Quansight/ragna/blob/c9fd83dd3116649cb007cc384cb70cf178623d9a/ragna/deploy/_api/schemas.py#L50-L57

in which the `content` attribute is just the message chunk. Meaning, for every chunk we transmit the full source object. While already wasteful, this will be greatly amplified by #248, which we agreed to implement. So we need to come up with a better scheme.

One option could be to transmit the chunk on its own for most of the stream and only on the first or last chunk transmit the remaining message data as well. If we go for this, I would prefer sending the message data together with the first chunk for a two reasons:

1. All the other information is available when we get the first chunk. Thus, there is no reason to not send it to the client right away.
1. This would match the Python API. In there we have direct access to `.sources` etc. attributes when the object is created.

Happy for other opinions on this or completely different suggestions on how to handle streaming for the REST API.

**Comment by peachkeel** [comment]
Your preference for how to implement this feature is eminently reasonable. The only reason not to send `sources` with the first chunk might be the added latency. If that's a factor for people, maybe having an option like `/chats/{id}/answer?sources=none` or `/chats/{id}/answer?sources=last` would be the best way to go so all bases are covered.

**Comment by pmeier** [comment]
> If that's a factor for people, maybe having an option like `/chats/{id}/answer?sources=none` or `/chats/{id}/answer?sources=last` would be the best way to go so all bases are covered.

Let's wait for demand on this. Thanks for the input.

**Comment by pmeier** [comment]
With #261 finally merged, adding streaming support for the UI was a breeze. Or so I thought. It kinda works, but we have an horrible "blinking" due to the rapid updates:

[Screencast from 2024-01-05 00-13-32.webm](https://github.com/Quansight/ragna/assets/6849766/60bc0894-6854-49a8-bbcd-829d1fff86dd)

@ahuang any idea where this is coming from or how to fix it?

**Comment by pierrotsmnrd** [comment]
I have had a look and my understanding is that the flickering is due to the repeated calls to [`ChatMessage._update_object_pane`](https://github.com/holoviz/panel/blob/2e805657a735b7938481ac77317ab7ec23441d8f/panel/chat/message.py#L567) when updating `message.object` [here](https://github.com/Quansight/ragna/blob/60d37a663bca5712c5f14f9942f11be2240103f0/ragna/deploy/_ui/central_view.py#L388).

I've tried various things to fix this but haven't found a proper solution yet.

**Comment by ahuang11** [comment]
My first thought is that it looks like CSS being reset then reloaded every time so it jumps around.

Also, this is how we stream internally: https://github.com/holoviz/panel/blob/main/panel/chat/feed.py#L379-L387

**Comment by pmeier** [comment]
@ahuang11 

> Can you check the browser console to see if that's true?

How would I do that?

**Comment by ahuang11** [comment]
If on Chrome, if the stylesheets are updated, it should show flashing / highlighting I think?

I am expecting the styles to be cached, but if not, it could be a bug in Panel.

**Comment by ahuang11** [comment]
Also, I find it strange that your Copy / Source Info buttons disappear and re-appear. Are they implemented in ReactiveHTML?

**Comment by pmeier** [comment]
@ahuang11 

> If on Chrome, if the stylesheets are updated, it should show flashing / highlighting I think?

Previous screengrab was from Firefox. The one below is from Chrome. I see some flashing style attributes. More important, on Chrome the message is not rendered at all until the very end:

[Screencast from 2024-01-10 21-21-47.webm](https://github.com/Quansight/ragna/assets/6849766/82f9cee2-d506-4aed-85e8-b78fa6d4d509)

> Also, I find it strange that your Copy / Source Info buttons disappear and re-appear. Are they implemented in ReactiveHTML?

They are:

- https://github.com/Quansight/ragna/blob/20ecade8992ecf458575a668a944b229e45e6f11/ragna/deploy/_ui/central_view.py#L47-L87
- https://github.com/Quansight/ragna/blob/20ecade8992ecf458575a668a944b229e45e6f11/ragna/deploy/_ui/central_view.py#L151-L170

We didn't find a better way to move them below our message than by replacing the original object pane with a column that includes the message and the buttons:

https://github.com/Quansight/ragna/blob/20ecade8992ecf458575a668a944b229e45e6f11/ragna/deploy/_ui/central_view.py#L137-L149

**Comment by ahuang11** [comment]
Thanks! Can you expand that flashing style attribute? 

If I understand Ragna's codebase correctly, I think rather than calling _update_object_pane repeatedly, through `message.object += chunk["content"]`, I think you can store a `_contents_pane` internally

```
_contents_pane = pn.pane.Markdown()
message.object = pn.Column(_contents_pane, ...)  # include the other two things you have
```

Then, call
`message._contents_pane.object += ...`

So you're only affecting the contents pane, and not updating an entire column each time.

pmeier:
I've refactored the message class a little after an offline chat with @ahuang11. Previously, we created with a `str` as `object` and had a renderer set that turned it into a `pn.pane.Markdown` down the line. Now, we do this inside the constructor and pass that up to the base class constructor. This has two benefits:

1. We don't need to overwrite private functionality from the base class to get our buttons in. In case we have a message from an assistant, we just wrap the markdown into a `pn.Column` right away.
2. We can create a `self.content_pane` object that can be updated while streaming the message without rendering anything else anew.

This is the result:

[Screencast from 2024-01-16 23-56-44.webm](https://github.com/Quansight/ragna/assets/6849766/836a90c9-b562-45f9-9a1c-8a0e461a973d)

pierrotsmnrd (approved):
Looks good to me. tested locally and it works.

PR #244, opened 2023-12-11
fix large upload #176

### LINKED ISSUE #176: [BUG] - net::ERR_INSUFFICIENT_RESOURCES Upon Uploading Large Corpus
### Bug description

Ragna UI hangs while attempting to upload 2261 documents via the file chooser.

### How to reproduce the bug?

1. Prepare a corpus of more than 1349 files.
2. Start Ragna UI.
3. Use browser to navigate to UI.
5. Start a conversation by uploading the corpus via the file chooser.
6. Wait for browser to hang.
7. Inspect JavaScript on the browser side.
8. Look for "net::ERR_INSUFFICIENT_RESOURCES" message(s).

- Ragna v0.1.1
- Chromium v117.0.5938.132 (Official Build) Linux

### Anything else?

It seems like this might be a known bug in Chrome/Chromium that has to do with making too many requests at once.

See also:
https://scratch.mit.edu/discuss/m/topic/88418/

peachkeel:
Fixes #176.

Batch concurrent file uploads from upload dialog to handle thousands of documents without browser lock ups.

### DIFF
--- a/ragna/deploy/_ui/resources/upload.js
+++ b/ragna/deploy/_ui/resources/upload.js
@@ -1,26 +1,23 @@
 function upload(files, token, informationEndpoint, final_callback) {
+  uploadBatches(files, token, informationEndpoint).then(final_callback);
+}
 
-  async function uploadBatches() {
+async function uploadBatches(files, token, informationEndpoint) {
+  const batchSize = 500;
+  const queue = Array.from(files);
 
-    let index = 0;
-    let uploaded = [];
+  let uploaded = [];
 
-    const batchSize = 500;
-    const queue = Array.from(files);
-
-    while(index < files.length) {
-      const batch = queue.splice(0, batchSize);
-      await Promise.all(
-        batch.map(file => uploadFile(file, token, informationEndpoint))
-      ).then(results => {
-        uploaded.push(...results);
-      });
-      index += batch.length;
-    }
-    return(uploaded);
+  while (queue.length) {
+    const batch = queue.splice(0, batchSize);
+    await Promise.all(
+      batch.map((file) => uploadFile(file, token, informationEndpoint)),
+    ).then((results) => {
+      uploaded.push(...results);
+    });
   }
 
-  uploadBatches().then(final_callback);
+  return uploaded;
 }
 
 async function uploadFile(file, token, informationEndpoint) {
--- a/ragna/deploy/_ui/resources/upload.js
+++ b/ragna/deploy/_ui/resources/upload.js
@@ -1,35 +1,25 @@
-function upload(files, token, informationsEndpoint, final_callback) {
-  const batchSize = 500; // Maximum number of concurrent uploads
-  let currentIndex = 0; // Tracks the index of the current file to be uploaded
-  let successfullyUploaded = []; // Array to hold the results
+function upload(files, token, informationEndpoint, final_callback) {
 
-  // Function to upload a single batch of files
-  function uploadBatch() {
-    // Get the next batch of files based on the current index and batch size
-    const batch = Array.from(files).slice(currentIndex, currentIndex + batchSize);
-    currentIndex += batchSize;
-    // Map over the batch to create an array of upload promises
-    return Promise.all(
-      batch.map(file => uploadFile(file, token, informationsEndpoint))
-    ).then(results => {
-      // Concatenate the results of the current batch to the main array
-      successfullyUploaded = successfullyUploaded.concat(results);
-      // If there are more files to upload, recursively call uploadBatch again
-      if (currentIndex < files.length) {
-        return uploadBatch();
-      } else {
-        // If all files are uploaded, invoke the final callback with the results
-        final_callback(successfullyUploaded);
-      }
-    });
+  async function uploadBatches() {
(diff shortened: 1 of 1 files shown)


pmeier (changes requested):
Thanks for the PR @peachkeel! A few high level comments:

1. Can we drop the comments on almost every single line? The only describe what is written and thus provide very little extra value. I feel the code is clear enough to go completely without comments TBH.
2. Instead of letting `uploadBatch` recursively call itself, can't we have a simple while loop with `currentIndex < files.length` as condition? Or better yet use [`.splice`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/splice) instead of manually tracking the index. For example the second code block in https://stackoverflow.com/a/7273794.

peachkeel:
These changes are doable, but I might not get to them until the weekend.

peachkeel:
Upon further reflection, I don't think we can just have a simple while loop here because we can only precede with a batch once we have confirmation an earlier batch has terminated. I'm file with removing all the comments, but I think the flow control must remain due to the asynchronous nature of the problem.

peachkeel:
I suppose the other option would be to make everything synchronous, which actually might not be a bad idea.

pmeier:
> Upon further reflection, I don't think we can just have a simple while loop here because we can only precede with a batch once we have confirmation an earlier batch has terminated.

That is true, but I don't see why that would be a problem. If the loop body processes one batch we are guaranteed that no more uploads are running at the same time. 

That is exactly the same as for the recursion solution: we only ever recurse further if the previous batch is completely uploaded.

peachkeel:
It's probably because I'm a terrible programmer and I don't like asynchronous programming nor JavaScript. I'll make one more attempt at refactoring.

pmeier:
> I'll make one more attempt at refactoring.

I can take over if you like. No need to stress over this.

peachkeel:
Lol, I'm OK. I'm into it now. I might as well finish!

peachkeel:
I'm afraid I just don't see what you're seeing.

Here is my revised code:

```javascript
function upload(files, token, informationEndpoint, final_callback) {

  const filesArray = Array.from(files)

  const batchSize = 500;
  let index = 0;
  let uploaded = [];

  while(index < filesArray.length) {

    const batch = filesArray.splice(index, batchSize);

    await Promise.all(
      batch.map(file => uploadFile(file, token, informationEndpoint))
    ).then(results => {
      uploaded.push(...results);
    });
    index += batchSize;
  }
  console.log(`Uploaded: ${uploaded.length}`);
  final_callback(uploaded);
}
```

**NOTE**: You have to turn `files` into an array to use `splice()`.

Beyond that, the above code won't work with `await` because the definition of `upload()` isn't `async` and I don't want change that definition. If I remove the `await` the code will barrel ahead and prematurely call the `final_callback`.

If you can give me a pointer, I'll make the appropriate changes, test them, and re-request a review.

peachkeel:
I think I got it!

ragna/deploy/_ui/resources/upload.js
```diff
@@ -1,9 +1,26 @@
-function upload(files, token, informationsEndpoint, final_callback) {
-  Promise.all(
-    Array.from(files).map((file) => {
-      return uploadFile(file, token, informationsEndpoint);
-    }),
-  ).then(final_callback);
+function upload(files, token, informationEndpoint, final_callback) {
+
+  async function uploadBatches() {
```
pmeier, line 3: Can we make this a standalone function instead of defining it inside `upload`?

ragna/deploy/_ui/resources/upload.js
```diff
@@ -1,9 +1,26 @@
...
+    const queue = Array.from(files);
+
+    while(index < files.length) {
+      const batch = queue.splice(0, batchSize);
+      await Promise.all(
+        batch.map(file => uploadFile(file, token, informationEndpoint))
+      ).then(results => {
+        uploaded.push(...results);
+      });
+      index += batch.length;
```
pmeier, line 18: It seems we are only using the `index` to track the loop iterations. Instead, could we just directly track the length of the `queue`?

```suggestion
    while(queue.length) {
      const batch = queue.splice(0, batchSize);
      await Promise.all(
        batch.map(file => uploadFile(file, token, informationEndpoint))
      ).then(results => {
        uploaded.push(...results);
      });
```

Since `.splice` actually removes elements the loop should again terminate when we pulled the last batch.

pmeier (commented):
Looking a lot better, thanks @peachkeel! Two minor comments left to potentially streamline this even more. Afterwards this is good to go!

pmeier (approved):
You gotta love that JS lets you define a variable as `const`, but then allows a destructive operation like `splice` on it :smiling_face_with_tear: Anyway, thanks a ton @peachkeel, LGTM!

peachkeel:
Ugh, now that you mention it, `uploaded` could probably also be a `const` given the way it is used in the code. I'll let you make that change if you want, I'm done with looking at JavaScript for another decade!

pmeier:
Nah, this is too insignificant to make a difference. Not going to oppose a PR, but I'll not send one either.

Issue #229, opened 2023-12-04
Chat info button displays wrong assistant


pmeier:
### Bug description

Per title, However, the selected assistant is correctly shown in the side bar when clicking the button.

### How to reproduce the bug?

Select an assistant that is not the default one and start the chat.

### Versions and dependencies used

0.1.2rc0

### Anything else?

![image](https://github.com/Quansight/ragna/assets/6849766/caab680c-a681-4805-b5f1-b9b05abd30c7)

pmeier:
This is defined at

https://github.com/Quansight/ragna/blob/b2730f1dd2b7285a13301cb3e8a7936eca7bd0bb/ragna/deploy/_ui/central_view.py#L613-L615

pmeier:
Ok, repro above is wrong. The value is stuck on whatever you select first. Meaning, if the first chat is `AssistantA | SourceStorageB`, all other chats will also have this label regardless of what components you select.

pmeier:
I'm not sure what is happening here, but as pointer for the future. The `button_name` variable holds the correct value even for the second and other later chats. For whatever reason, it is not accepted though.

nenb:
@pmeier I've just looked into this and I think #252 should fix this (it works locally for me at least).

pmeier:
Closed in #252.

PR #252, opened 2023-12-17
Fix rendering of chat info button

nenb:
Addresses #229. 

Currently a new chat info button is created whenever the UI is updated. This creates a problem as when the UI is re-rendered it is still referencing the original button (which has not actually been updated). To fix this issue a button should only be created once and then updated in place prior to each re-render.

### DIFF
--- a/ragna/deploy/_ui/central_view.py
+++ b/ragna/deploy/_ui/central_view.py
@@ -243,17 +243,14 @@ class CentralView(pn.viewable.Viewer):
 
         self.api_wrapper = api_wrapper
         self.chat_info_button = pn.widgets.Button(
-            name="", button_style="outline", icon="info-circle"
+            # The name will be filled at runtime in self.header
+            name="",
+            on_click=self.on_click_chat_info_wrapper,
+            button_style="outline",
+            icon="info-circle",
+            stylesheets=[":host { margin-top:10px; }"],
         )
         self.on_click_chat_info = None
-        self.chat_info_button.stylesheets.append(
-            """ 
-            :host {  
-                margin-top:10px;
-                }                    
-            """
-        )
-        self.chat_info_button.on_click(self.on_click_chat_info_wrapper)
 
     def on_click_chat_info_wrapper(self, event):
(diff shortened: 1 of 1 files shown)


ragna/deploy/_ui/central_view.py
```diff
@@ -613,26 +613,27 @@ def header(self):
 
                 chat_documents_pills.append(pill)
 
-        chat_info_button = None
-        if self.current_chat is not None:
-            button_name = f"{self.current_chat['metadata']['assistant']} | {self.current_chat['metadata']['source_storage']}"
-            chat_info_button = pn.widgets.Button(
+        button_name = f"{self.current_chat['metadata']['assistant']} | {self.current_chat['metadata']['source_storage']}"
+        if hasattr(self, "chat_info_button"):
```
pmeier, line 617: Would it make sense to create this button in the constructor with an empty name at first and inside this method unconditionally overwrite the name?

ragna/deploy/_ui/central_view.py
```diff
@@ -613,26 +613,27 @@ def header(self):
...
-            chat_info_button.on_click(self.on_click_chat_info_wrapper)
+        self.chat_info_button.stylesheets.append(
+            """
+                                    :host {  
+                                                margin-top:10px;
+                                    }
+                                    
+                                """
+        )
+        self.chat_info_button.on_click(self.on_click_chat_info_wrapper)
```
pmeier, line 631: 1. Unless I'm missing something here, these could be static too. Meaning, we can put that in the constructor of the class, since neither the CSS nor the callback is dependent on anything in this method.
2. Since we are already touching this: could you please format this absolutely hideous CSS formatting?

pmeier (commented):
Thanks @nenb! Can confirm that this fixes the issue. LGTM after my comments have been addressed!

nenb:
@pmeier I've added your suggestions and I think this should be ready for merge now.

pmeier (approved):
Thanks @nenb! I took the liberty to merge style sheet and callback into the constructor.

Issue #240, opened 2023-12-08
Cannot create over 25 chats


Tengal-Teemo:
### Bug description

Ragna UI fails to create over 25 chats, + New Chat button becomes non-responsive with error message:
```
ERROR:bokeh.server.protocol_handler:error handling message
 message: Message 'PATCH-DOC' content: {'events': [{'kind': 'MessageSent', 'msg_type': 'bokeh_event', 'msg_data': {'type': 'event', 'name': 'button_click', 'values': {'type': 'map', 'entries': [['model', {'id': 'p16815'}]]}}}]} 
 error: UnknownReferenceError("can't resolve reference 'p16815'")
Traceback (most recent call last):
  File "/home/tazukiswift/miniconda3/envs/ragna-dev/lib/python3.9/site-packages/bokeh/server/protocol_handler.py", line 97, in handle
    work = await handler(message, connection)
  File "/home/tazukiswift/miniconda3/envs/ragna-dev/lib/python3.9/site-packages/bokeh/server/session.py", line 94, in _needs_document_lock_wrapper
    result = func(self, *args, **kwargs)
  File "/home/tazukiswift/miniconda3/envs/ragna-dev/lib/python3.9/site-packages/bokeh/server/session.py", line 286, in _handle_patch
    message.apply_to_document(self.document, self)
  File "/home/tazukiswift/miniconda3/envs/ragna-dev/lib/python3.9/site-packages/bokeh/protocol/messages/patch_doc.py", line 104, in apply_to_document
    invoke_with_curdoc(doc, lambda: doc.apply_json_patch(self.payload, setter=setter))
  File "/home/tazukiswift/miniconda3/envs/ragna-dev/lib/python3.9/site-packages/bokeh/document/callbacks.py", line 443, in invoke_with_curdoc
    return f()
  File "/home/tazukiswift/miniconda3/envs/ragna-dev/lib/python3.9/site-packages/bokeh/protocol/messages/patch_doc.py", line 104, in <lambda>
    invoke_with_curdoc(doc, lambda: doc.apply_json_patch(self.payload, setter=setter))
  File "/home/tazukiswift/miniconda3/envs/ragna-dev/lib/python3.9/site-packages/bokeh/document/document.py", line 374, in apply_json_patch
    patch: PatchJson = deserializer.deserialize(patch_json)
  File "/home/tazukiswift/miniconda3/envs/ragna-dev/lib/python3.9/site-packages/bokeh/core/serialization.py", line 510, in deserialize
[...]

PR #259, opened 2023-12-20
Bump panel minimum version

nenb:
Addresses #240.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -28,7 +28,7 @@ dependencies = [
     "httpx",
     "importlib_metadata>=4.6; python_version<'3.10'",
     "packaging",
-    "panel>=1.3,<1.4",
+    "panel>=1.3.6,<1.4",
     "pydantic>=2",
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -28,7 +28,7 @@ dependencies = [
     "httpx",
     "importlib_metadata>=4.6; python_version<'3.10'",
     "packaging",
-    "panel>=1.3,<1.4",
+    "panel>=1.3.6",  # GitHub issue #240
```
pmeier, line 31: 1. Let's keep the upper pin for now. The chat components are under active development and had major changes between 1.2 and 1.3. I don't want Ragna to break for users if something like that happens for 1.4 and this gets automatically installed.
2. No need for the comment. `git blame` will show the PR that touched this and in turn also link to the issue.

```suggestion
    "panel>=1.3.6,<1.4",
```

pmeier (approved):
Thanks Nick!

PR #261, opened 2023-12-21
Cleanup UI code

pmeier:
While working on #215, I found it really difficult to add the change to the UI. For the most part I was confused how the chat interface was actually used. So I looked more into it with the goal of understanding how it works to be able to implement #215. And doing that lead to this PR, which majorly cleans up our code. 

There are still a few things I dislike, e.g. are calling private APIs of panel, but I think this is the best we can do for now. I've marked these places as `#TODO: @panel` to eventually discuss with them.

### DIFF
--- a/ragna/deploy/_ui/central_view.py
+++ b/ragna/deploy/_ui/central_view.py
@@ -1,6 +1,6 @@
 from __future__ import annotations
 
-from typing import Callable, Literal, Optional
+from typing import Callable, Literal, Optional, cast
 
 import panel as pn
 import param
@@ -169,8 +169,8 @@ class RagnaChatMessage(pn.chat.ChatMessage):
                 return "https://upload.wikimedia.org/wikipedia/commons/a/a4/GPT-4.png"
         elif organization == "Anthropic":
             return "https://upload.wikimedia.org/wikipedia/commons/1/14/Anthropic.png"
-        else:
-            return model[0].upper()
+
+        return model[0].upper()
 
     def _render(self, content: str) -> pn.pane.Markdown:
         return pn.pane.Markdown(
@@ -352,9 +352,9 @@ class CentralView(pn.viewable.Viewer):
         if role == "system":
             return "Ragna"
         elif role == "user":
-            return self.user
+            return cast(str, self.user)
         elif role == "assistant":
-            return self.current_chat["metadata"]["assistant"]
+            return cast(str, self.current_chat["metadata"]["assistant"])
         else:
             raise RuntimeError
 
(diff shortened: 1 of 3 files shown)


ragna/deploy/_ui/central_view.py
```diff
@@ -405,148 +390,107 @@ def chat_interface(self):
...
             callback=self.chat_callback,
-            callback_user="Ragna",
+            user=self.user,
             show_rerun=False,
             show_undo=False,
             show_clear=False,
             show_button_name=False,
             view_latest=True,
             sizing_mode="stretch_width",
+            # TODO: @panel hitting enter to send a message is fine, but clicking
```
ahuang11, line 400: I agree; we are eventually planning to have a custom Typescript chat text area input model that supports keyboard hotkeys and not send automatically when clicking away.

ragna/deploy/_ui/central_view.py
```diff
@@ -405,148 +390,107 @@ def chat_interface(self):
...
-                    msg = chat_interface.objects[i]
-                    needs_refresh = True
-
-            if needs_refresh:
-                chat_interface._chat_log.param.trigger("objects")
-
-        chat_interface.param.watch(
-            messages_changed,
-            ["objects"],
+        # TODO: @panel ChatFeed has a card_params parameter, but this isn't used
```
ahuang11, line 423: I think this is an oversight; I'll try to fix this soon.

ragna/deploy/_ui/central_view.py
```diff
@@ -117,130 +86,156 @@ class RagnaChatCopyIcon(ReactiveHTML):
...
+                        "background": "none",
+                        "border": "lightgray",
+                        "border-style": "solid",
+                        "border-width": "thin",
+                    },
+                ),
+                (
+                    ":host .message-content-assistant",
+                    {"background": "lightgray"},
+                ),
```
pmeier, line 202: This is the only part remaining that I'm not able to solve. For some reason this is not properly applied. @pierrotsmnrd could you have a look?

ragna/deploy/_ui/central_view.py
ahuang11: Fix in this PR:
https://github.com/holoviz/panel/pull/6154

ragna/deploy/_ui/central_view.py
ahuang11: An alternative is to use TextAreaInput, but the downside is you can't click enter to send the message.

pierrotsmnrd (approved):

PR #263, opened 2024-01-10
Decouple prepare from chat

nenb:
WIP


ragna/core/_rag.py
```diff
@@ -28,6 +27,17 @@
 C = TypeVar("C", bound=Component)
 
 
+async def _run(
```
pmeier, line 30: This should probably go into `core._utils` together with `run_gen`.

ragna/core/_rag.py
```diff
@@ -59,35 +69,110 @@ def _load_component(self, component: Union[Type[C], C]) -> C:
...
+            documents: Documents to use.
+            source_storage: Source storage to use.
+            corpus_name: Name of the corpus.
+            **params: Additional parameters passed to the source storage.
+        """
+        return Corpus(
+            self,
+            documents=documents,
+            source_storage=source_storage,
+            corpus_name=corpus_name,
```
pmeier, line 92: This can be just `name` since `rag.corpus(corpus_name=...)` is redundant.

ragna/core/_rag.py
```diff
@@ -59,35 +69,110 @@ def _load_component(self, component: Union[Type[C], C]) -> C:
...
+                    http_status_code=404,
+                )
+
+            documents_.append(document)
+        return documents_
+
+
 class SpecialChatParams(pydantic.BaseModel):
     user: str = pydantic.Field(default_factory=default_user)
-    chat_id: uuid.UUID = pydantic.Field(default_factory=uuid.uuid4)
```
pmeier, line 90: This should stay. Downstream components should still have access to it.

ragna/core/_rag.py
```diff
@@ -130,15 +196,13 @@ def __init__(
         self,
         rag: Rag,
         *,
-        documents: Iterable[Any],
```
pmeier, line 133: We need to be able to also pass in regular paths here. While `Corpus` is a useful abstraction, it is unnecessary for the base case. I would move the `_parse_documents` function that you moved to `Corpus` back here and let it return a `Corpus` in case we encounter a list of paths.

ragna/core/_rag.py
```diff
@@ -147,66 +211,33 @@ def __init__(
         self.params = params
         self._unpacked_params = self._unpack_chat_params(params)
 
-        self._prepared = False
         self._messages: list[Message] = []
 
-    async def prepare(self) -> Message:
```
pmeier, line 153: IMO the chat should keep the `prepare` method so users that just have local documents do not ever interact with the `Corpus` abstraction. Basically this would just be

```python
async def prepare(self):
    return await self.corpus.prepare()
```

Plus, by keeping this, we could also keep the functionality of adding the welcome message here.

ragna/core/_rag.py
```diff
@@ -147,66 +211,33 @@ def __init__(
...
 
         prompt = Message(content=prompt, role=MessageRole.USER)
         self._messages.append(prompt)
 
-        sources = await self._run(
-            self.source_storage.retrieve, self.documents, prompt.content
+        kwargs = self._unpacked_params[self.corpus.source_storage.retrieve]
+        sources = await _run(
+            self.corpus.source_storage.retrieve,
+            self.corpus.documents,
```
pmeier, line 231: `SourceStorage.store` and `SourceStorage.retrieve` should take a `Corpus` rather than a list of documents.

ragna/core/_rag.py
```diff
@@ -147,66 +211,33 @@ def __init__(
...
+            prompt.content,
+            corpus_name=self.corpus.corpus_name,
```
pmeier, line 233: This needs to be part of the `unpacked_params`.

ragna/core/_rag.py
```diff
@@ -221,33 +252,17 @@ async def answer(self, prompt: str) -> Message:
...
         component_models = {
             getattr(component, name): model
-            for component in [self.source_storage, self.assistant]
+            for component in [self.corpus.source_storage, self.assistant]
             for (_, name), model in component._protocol_models().items()
         }
 
         ChatModel = merge_models(
-            str(self.params["chat_id"]),
+            str(self.params["chat_name"]),
```
pmeier, line 265: We can roll this back if the we keep the `"chat_id"`, correct?

ragna/core/_rag.py
```diff
@@ -261,22 +276,12 @@ def _unpack_chat_params(
...
-        else:
-            fn = cast(Callable[..., T], fn)
-            return await anyio.to_thread.run_sync(
-                functools.partial(fn, *args, **kwargs)
-            )
-
-    async def __aenter__(self) -> Chat:
-        await self.prepare()
-        return self
+    # TODO: Do we need this?
```
pmeier, line 279: See explanation above why I want to keep the `prepare` and in turn this as well.

ragna/core/_rag.py
```diff
@@ -59,35 +69,110 @@ def _load_component(self, component: Union[Type[C], C]) -> C:
...
             self,
-            documents=documents,
-            source_storage=source_storage,
+            corpus=corpus,
             assistant=assistant,
             **params,
         )
 
 
+class Corpus:
```
pmeier, line 118: Since `Corpus` now effectively handles `SourceStorage.store` it should probably has its own `unpacked_params`.

ragna/core/_rag.py
```diff
@@ -59,35 +69,110 @@ def _load_component(self, component: Union[Type[C], C]) -> C:
...
+        **params: Additional parameters passed to the source storage.
+    """
+
+    def __init__(
+        self,
+        rag: Rag,
+        *,
+        documents: Iterable[Any],
+        source_storage: Union[Type[SourceStorage], SourceStorage],
+        corpus_name: str,
```
pmeier, line 138: At least in the "user just passes document paths", this needs to have a default.

pmeier (changes requested):
Thanks for the initial PR @nenb! I have a bunch of inline comments. One overarching concern that I have is how we deal with the `**params` handling. With only `Chat` in the picture this was simple. Now that `Chat` *and* `Corpus` can request params, this is harder. Both need to perform their separate validation and we need to make sure that every parameter is used either by `Corpus` or `Chat`.

pmeier:
I'm ok with this an an initial implementation when my comments are addressed and CI is green.

What I dislike here is that `Corpus` is a second class citizen. It is more like a dataclass and one can only interact with it through the `Chat`. However, and admin in the "managed Ragna" use case which we are targeting here, only wants to interact with the `Corpus` and not the `Chat`. Meaning, they need to have an assistant present and meet its requirements before they can even prepare a `Corpus`. In the presence of the demo assistant that is not terribly bad, but it is not sound design. I'm wondering if the source storage should be a component of the `Corpus` and the assistant a component of the `Chat`. That way, the `Corpus` can stand on its own, e.g. can be prepared, without the need to ever create a `Chat`.

ragna/source_storages/_chroma.py
```diff
@@ -35,20 +35,20 @@ def __init__(self) -> None:
...
-        documents: list[Document],
+        corpus: Corpus,
         *,
         chat_id: uuid.UUID,
         chunk_size: int = 500,
         chunk_overlap: int = 250,
     ) -> None:
         collection = self._client.create_collection(
-            str(chat_id), embedding_function=self._embedding_function
+            corpus.name, embedding_function=self._embedding_function
```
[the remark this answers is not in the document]
nenb: I'm not sure about this. I think it's a better UX to have names be human-readable eg the GitHub namespace of `user/repo`. If we were to generate some form of ID inside of `ragna`, then this would mean that actual embeddings stored in the `source_storage` would not be user friendly for other libraries ie user's would have to remember the ID they need.

I would prefer to just provide a useful error message in the case of conflicts, and to recommend to users to adopt a naming convention like `user/table_name` or something similar.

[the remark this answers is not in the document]
pmeier: That is a good point. In that case should we maybe drop the `name` attribute in favor of `id: str`. The user is still free to use whatever ID they want, e.g. human readable, but the name "id" implies that it should be unique.

nenb:
This PR has become quite messy. I'm closing it (to preserve the history, if we need it for whatever reason in the future), and opening a clean version in #269. (Trying to add #269 on top of this made the history completely unreadable.)

PR #264, opened 2024-01-11
return source content from the REST API

### LINKED ISSUE #248: Also send source content as part of the message object of the REST API / Web UI
### Feature description

While our `ragna.core.Source` object stores the source content 

https://github.com/Quansight/ragna/blob/7b176528c7524edc8bcf8754695caa5a363a8e23/ragna/core/_components.py#L79-L88

the source object of the REST API does not:

https://github.com/Quansight/ragna/blob/7b176528c7524edc8bcf8754695caa5a363a8e23/ragna/deploy/_api/schemas.py#L35-L39

This is an intentional choice. When creating messages "live" we have access to the sources

https://github.com/Quansight/ragna/blob/7b176528c7524edc8bcf8754695caa5a363a8e23/ragna/deploy/_api/core.py#L238

However when chats and in turn messages and sources are recreated from our state database, we would only have access to them if we stored the source content as part of the state database. We don't want to do that for two reasons:

1. We are duplicating the storage. As the name implies, the `ragna.core.SourceStorage` already stores the source content.
2. There are potential security issues since the state database might have different access control as the source storage, e.g. vector database.

Instead what we could do is to add a new protocol method `ragna.core.SourceStorage.get` (name TBD):

```python
def get(self, chat_id: uuid.UUID, source_id: str) -> ragna.core.Source:
    ...
```

In contrast to `ragna.core.SourceStorage.retrieve` will `.get()` not engage in the source selection based on the embedding of a prompt, but rather get one source by its ID. With that we could fill the source content when construction the chat object from the state database.

### Value and/or benefit

When viewing the sources in the web UI, we currently only displaying the location of the source in the document, e.g. the page number. However, for documents like `.txt` there is no page number available. Meaning, the only information a user gets is that the corresponding document was used, but not what section of it. If we had the source content available, we could display that instead.

For the web UI we need to think carefully how this would look like, since we don't want to overwhelm the user with information.

cc @peachkeel for other use cases / scenarios that I missed.

### Anything else?

Originally reported by @peachkeel in https://github.com/Quansight/ragna/issues/242#issuecomment-1848669705

pmeier:
Closes #248. Since we need to touch the database tables, this will not work with existing ones without handling. I'm going to open a separate issue for that. Edit: See #265.

I need help styling the accordion in the UI. By default it looks like

Screenshot from 2024-01-11 09-46-04

The color is off and the title doesn't seem to support Markdown. We don't actually need Markdown, but some way to make the part in `**...**` bold. When opening an item we also have another issue:

Screenshot from 2024-01-11 09-46-32

If long enough, the content of the first item happily expands behind the other items. We should probably make it scrollable.

@smeragoel is working on a design for this.

### DIFF
--- a/ragna/deploy/_api/database.py
+++ b/ragna/deploy/_api/database.py
@@ -170,6 +170,8 @@ def _schema_to_orm_source(session: Session, source: schemas.Source) -> orm.Sourc
             id=source.id,
             document_id=source.document.id,
             location=source.location,
+            content=source.content,
+            num_tokens=source.num_tokens,
         )
         session.add(orm_source)
         session.commit()
--- a/ragna/deploy/_api/database.py
+++ b/ragna/deploy/_api/database.py
@@ -113,6 +113,8 @@ def _orm_to_schema_chat(chat: orm.Chat) -> schemas.Chat:
                     id=source.id,
                     document=_orm_to_schema_document(source.document),
                     location=source.location,
+                    content=source.content,
+                    num_tokens=source.num_tokens,
                 )
                 for source in message.sources
             ],
--- a/ragna/deploy/_api/orm.py
+++ b/ragna/deploy/_api/orm.py
@@ -84,6 +84,8 @@ class Source(Base):
     document = relationship("Document", back_populates="sources")
 
     location = Column(types.String)
+    content = Column(types.String)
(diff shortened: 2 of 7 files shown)


pierrotsmnrd (approved):
Looks good to me. Tested locally, it works.

Issue #275, opened 2024-01-16
Login endpoint issue when launching Ragna on JupyterHub(NEbari)


dharhas:
### Bug description

When I launch "ragna ui" from the command line on JupyterHub/Nebari with say:

`BOKEH_ALLOW_WS_ORIGIN=nebari.quansight.dev ragna ui`

Ragna is deployed with JupyterHub Server Proxy to the url:

`https://nebari.quansight.dev/user/dharhas@quansight.com/proxy/31477/`

This works but the the login page gets redirected to 

https://nebari.quansight.dev/auth and shows a 404.

gets redirected to 

`https://nebari.quansight.dev/auth`

manually changing the auth url to  `https://nebari.quansight.dev/user/dharhas@quansight.com/proxy/31477/auth`

lets me login and then I can use ragna. 

### How to reproduce the bug?

launch a terminal on Jupyterhub/ragna and launch ragna ui

### Versions and dependencies used

ragna - 0.1.3
python  - 3.11.7
os - linux

pierrotsmnrd:
It was not intentional. I don't know if this has an impact or not. worth trying.

pmeier:
@aktech confirmed that #276 indeed fixed the issue. However, we have for more instances of hardcoded paths for the login mechanism:

- https://github.com/Quansight/ragna/blob/ea9ffeb526098c0d31c001bee2c69bbb84eb999d/ragna/deploy/_ui/app.py#L66
- https://github.com/Quansight/ragna/blob/ea9ffeb526098c0d31c001bee2c69bbb84eb999d/ragna/deploy/_ui/app.py#L76
- https://github.com/Quansight/ragna/blob/ea9ffeb526098c0d31c001bee2c69bbb84eb999d/ragna/deploy/_ui/app.py#L92
- https://github.com/Quansight/ragna/blob/ea9ffeb526098c0d31c001bee2c69bbb84eb999d/ragna/deploy/_ui/logout_page.py#L22

They also need be relative. This most likely has to happen in JS.

pmeier:
@aktech This is solved with the PRs you have send, correct?

aktech:
I have not tested it yet. I am testing is right now.

aktech:
The login endpoint issue is fixed (which is what this issue is about), I am able to login via the proxy url, which takes me to the `/auth`.

[...]

PR #280, opened 2024-01-17
Fix relative path handling for ragna panel app

aktech:
- [x] Fixes relative paths for panel apps

This doesn't handles the FastAPI paths yet, have just added an optional environment variable: `RAGNA_API_ROOT_PATH` for specifying root path for FastAPI server.


ragna/deploy/_api/core.py
```diff
@@ -37,7 +38,11 @@ def get_component(display_name: str) -> Component:
 
         return component
 
-    app = FastAPI(title="ragna", version=ragna.__version__)
+    app = FastAPI(
+        title="ragna",
+        version=ragna.__version__,
+        root_path=os.environ.get("RAGNA_API_ROOT_PATH", ""),
```
pmeier, line 44: IMO, this should be a config option. You just need to put

```python
root_path: str = ""
```

into

https://github.com/Quansight/ragna/blob/c08a223e5eb6450681cabd7cfd5843c30dbe8d60/ragna/deploy/_config.py#L55-L62

That way you can set it through the config file but also through the `RAGNA_API_ROOT_PATH` environment variable.

ragna/deploy/_ui/js_utils.py
```diff
@@ -0,0 +1,58 @@
...
+            if (currentPath.includes('/{{remove}}')) {
+              console.log("Removing {{remove}} from " + currentPath)
+              if ("{{remove}}") {
+                currentPath = currentPath.replace(/\/{{remove}}(\/)?$/, '')
+              }
+              var redirectTo = currentPath + '{{append}}';
+              console.log("Redirecting to " + redirectTo)
+              window.location.href = redirectTo;
+              if ({{remove_auth_cookie}}) {
+                console.log("Removing auth_token cookie")
```
pmeier, line 48: Let's not.

```suggestion
```

This applies to all other `console.log`s as well.

ragna/deploy/_ui/js_utils.py
```diff
@@ -0,0 +1,58 @@
...
+
+    # This will remove nothing from the end of the url and will
+    # add auth to it, so /foo/bar/car/ becomes /foo/bar/car/logout
+    >>> redirect_script(remove="", append="logout")
+
+    # This will remove "auth" from the end of the url and will add / to it
+    # so /foo/bar/car/auth becomes /foo/bar/car/
+    >>> redirect_script(remove="auth", append="/")
+    """
+    js_script = preformat(
```
pmeier, line 35: IIUC, the only reason we need the `preformat` with six `str.replace` is because we want to use `str.format` later and our original text might already have curly braces inside of it, correct? If so, wouldn't it be easier to rely on `re.sub(r"\{\{\w+\}\}")` or whatever other delimiters for the substitution you want to use?

pmeier (commented):
Thanks Amit for the PR!

> This doesn't handles the FastAPI paths yet

I assume you mean "this doesn't handle FastAPI paths **automatically** yet", correct? I think that is ok. I don't see how the API would be able to detect this in the first place.

---

I have one question for the JS redirects: it seems we are doing mostly relative redirects, which have the disadvantage that the developer needs to know exactly at which path they currently are. Is there a technical blocker from always doing absolute redirects?

pmeier:
Does this PR resolve #275 or is there even more to the story?

ragna/deploy/_ui/js_utils.py
aktech: So you mean without using `str.format`?

aktech:
> I assume you mean "this doesn't handle FastAPI paths automatically yet", correct? I think that is ok. I don't see how the API would be able to detect this in the first place.

Yes, FastAPI would need a root api path, which I have added. If it needs more changes, then it's beyond the scope of this PR.

> Does this PR resolve https://github.com/Quansight/ragna/issues/275 or is there even more to the story?

Mostly, but I am not sure as I said I have not looked at the potential issues with FastAPI redirects. I want to keep the scope of this PR limited to panel redirects only. I'll create a separate PR for potential FastAPI changes if any. I added the root api path config option because it's anyway good to have and was an easy fix to potential redirect issues.

nenb (approved):
@pmeier I'm approving this as I don't see any major issues and any potential minor issues can likely be addressed in a subsequent PR.

PR #281, opened 2024-01-18
Support using .docx files

### LINKED ISSUE #225: [ENH] - Add support for .doc / .docx
### Feature description

Word documents are everywhere in the corporate world. We should support them out of the box. There are multiple Python packages out there that provide this functionality. We should do a light comparison of them to get a clearer picture. Optimally, we find a package that can give us page information. Not sure if that is possible in these data formats though.

### Value and/or benefit

Users with corporate documents will have a lower barrier of entry. Requested in https://github.com/Quansight/ragna/issues/224#issuecomment-1838864057.

### Anything else?

_No response_

paskett:
@pmeier If you could check out the test I added, let me know if you think that's sufficient.

Resolves #225


paskett:
Looks like type checks failed because of the import of `python-docx` (imported as just `docx`) dependency. I could add a `# type: ignore` comment.

ragna/core/_document.py
```diff
@@ -269,3 +270,32 @@ def extract_pages(self, document: Document) -> Iterator[Page]:
...
+        - [`python-docx`](https://github.com/python-openxml/python-docx)
+    """
+
+    @classmethod
+    def requirements(cls) -> list[Requirement]:
+        return [PackageRequirement("python-docx>=1.1.0")]
+
+    @classmethod
+    def supported_suffixes(cls) -> list[str]:
+        return [".docx"]
```
pmeier, line 292: Does that only support .docx or the ancient .doc as well?

ragna/core/_document.py
```diff
@@ -269,3 +270,32 @@ def extract_pages(self, document: Document) -> Iterator[Page]:
         ) as document:
             for number, page in enumerate(document, 1):
                 yield Page(text=page.get_text(sort=True), number=number)
+
+
+@DOCUMENT_HANDLERS.load_if_available
+class DocxDocumentHandler(DocumentHandler):
+    """Document handler for `.docx` documents.
+
+    Note: this does *not* extract text from headers or footers.
```
pmeier, line 279: With this, we get a proper note that stands out visually to the reader.

```suggestion
    !!! note
    
        This does *not* extract text from headers or footers.
```

pyproject.toml
```diff
@@ -59,6 +59,7 @@ all = [
     "lancedb>=0.2",
     "pyarrow",
     "pymupdf>=1.23.6",
+    "python-docx>=1.1.0",
```
pmeier, line 62: Any reason for this pin? If not, I would prefer to leave it unpinned to give the solver more flexibility.

tests/core/test_document.py
```diff
@@ -0,0 +1,26 @@
+import pytest
+
+from ragna.core import DocxDocumentHandler, LocalDocument
+
+
+@pytest.fixture()
```
pmeier, line 6: If we only have a single test right now, I don't think we need these fixtures. We can just inline them.

tests/core/test_document.py
```diff
@@ -0,0 +1,26 @@
...
+def docx_text():
+    return "Ragna is neat!"
+
+
+@pytest.fixture()
+def tmp_docx_document(tmp_path, docx_text):
```
pmeier, line 12: Same as above.

tests/core/test_document.py
```diff
@@ -0,0 +1,26 @@
...
+    document = docx.Document()
+    document.add_heading(docx_text)
+    document.add_paragraph(docx_text)
+    path = tmp_path / "test_document.docx"
+    document.save(path)
+    return LocalDocument.from_path(path)
+
+
+def test_docx(tmp_docx_document, docx_text):
+    handler = DocxDocumentHandler()
```
pmeier, line 24: Instead of creating the handler manually here, you could additionally assert that the auto-selected one is correct, i.e.

```suggestion
    assert isinstance(tmp_docx_document.handler, DocxDocumentHandler)
```

tests/core/test_document.py
```diff
@@ -0,0 +1,26 @@
...
+    for page in handler.extract_pages(tmp_docx_document):
+        assert page.text == docx_text
```
pmeier, line 26: 1. Instead of using the handler directly, you can also call `Document().extract_pages()`:

   https://github.com/Quansight/ragna/blob/c08a223e5eb6450681cabd7cfd5843c30dbe8d60/ragna/core/_document.py#L79-L80
2. We can harden this check by also checking that we get exactly one page back.

```suggestion
    pages = list(tmp_docx_document.extract_pages())
    assert len(pages) == 1
    
    page = pages[0]
    assert page.text == docx_text
```

pmeier:
> Looks like type checks failed because of the import of `python-docx` (imported as just `docx`) dependency. I could add a `# type: ignore` comment.

This is a quite common thing and the best solution here is to ignore this package globally. YOu can add it here

https://github.com/Quansight/ragna/blob/c08a223e5eb6450681cabd7cfd5843c30dbe8d60/pyproject.toml#L138-L146

ragna/core/_document.py
paskett: It doesn't support the ancient .doc AFAIK

paskett:
I'll get some of those suggestions committed when I have a chance later tonight (~20:00 MST)

paskett:
@pmeier Got all the suggestions put in and pipeline passing, let's ship it! 😇

pmeier (commented):
Thanks @paskett. The PR is almost ready with one question left below.

tests/core/test_document.py
```diff
@@ -0,0 +1,23 @@
+from ragna.core import DocxDocumentHandler, LocalDocument
+
+
+def get_docx_document(tmp_path, docx_text):
+    import docx
```
pmeier, line 5: Nit: you can put this at the top level of the module. Here we are in a development environment and this is always available.

paskett: I moved the import to the top level

tests/core/test_document.py
```diff
@@ -0,0 +1,23 @@
...
+    document.save(path)
+    return LocalDocument.from_path(path)
+
+
+def test_docx(tmp_path):
+    docx_text = "ragna is neat!"
+    tmp_docx_document = get_docx_document(tmp_path, docx_text)
+    assert isinstance(tmp_docx_document.handler, DocxDocumentHandler)
+    pages = list(tmp_docx_document.extract_pages())
+    assert len(pages) == 2
```
pmeier, line 20: Why has this document two pages if we only have three words in it?

pmeier (commented):
Sorry @paskett, somehow my actual comment got lost.

tests/core/test_document.py
paskett: That's a good question. So the document we wrote here has two elements, a heading (different from a header -- it's also considered a paragraph in python_docx, but with special styling on it) and a paragraph, both with the same text in them. The way I've done the Docx handler, it retrieves a different `Page` for each paragraph in the document. We could combine them all beforehand and return a single page, more similar to the .txt handler. 

Current way:
```python
        for paragraph in document_docx.paragraphs:
            text = paragraph.text
            if len(text) > 0:
                yield Page(text=text)
```

Potential way with only one Page:
```python
        text = "\n".join(paragraph.text for paragraph in document_docx.paragraphs)
        yield Page(text=text)
```

Personally, I don't have a huge preference either way.

paskett: With your insight into the codebase, do you think fewer, bigger pages going to be more performant, or many smaller pages?

paskett: I just added an assertion for both paragraphs.

pmeier (approved):
Thanks @paskett!

PR #283, opened 2024-01-19
dockerize

pmeier:


ragna-docker.toml
```diff
@@ -0,0 +1,22 @@
...
+
+[components]
+source_storages = [
+    "ragna.source_storages.Chroma",
+    "ragna.source_storages.RagnaDemoSourceStorage",
+    "ragna.source_storages.LanceDB"
+]
+assistants = [
+    "ragna.assistants.RagnaDemoAssistant"
+]
```
pmeier, line 13: These are all the builtin components that we can activate by default. For everything else we need some input from the user, i.e. environment variables. It sucks, that we can't include everything here and "deactivate" the components for which the requirements are not met. Will open an issue about this.

.github/workflows/docker.yml
```diff
@@ -0,0 +1,91 @@
...
+on:
+  pull_request:
+    paths:
+      - ".github/workflows/docker.yml"
+      - ".dockerignore"
+      - "Dockerfile"
+      - "environment-dev.yml"
+      - "pyproject.toml"
+      - "ragna-docker.toml"
+      - "requirements-docker.txt"
```
pmeier, line 12:
```suggestion
      - "requirements-docker.lock"
```

.github/workflows/docker.yml
```diff
@@ -0,0 +1,91 @@
...
+          TAG=$(echo "${VERSION}" | sed 's/+/-/g')
+          echo "tag=${TAG}" | tee --append $GITHUB_OUTPUT
+
+      - name: Login to Quay.io
+        if: ${{ github.event_name == 'workflow_dispatch' }}
+        uses: docker/login-action@v3
+        with:
+          registry: ${{ env.DOCKER_REGISTRY }}
+          username: ${{ secrets.DOCKER_REGISTRY_USERNAME }}
+          password: ${{ secrets.DOCKER_REGISTRY_PASSWORD }}
```
pmeier, line 79: @costrouc This was the easiest right now. LMK if you want this done differently, i.e. pulling this from Vault or the like.

nenb (approved):
LGTM.

@nenb will open an issue to track future optimisations to the size of the Dockerfile.

Issue #178, opened 2023-11-08
Refactor authentication in the UI


pmeier:
Currently, the basic auth, i.e. username / password, is hardcoded in the UI. We did that to meet our presentation deadline. In the aftermath however, we need to lift this and actually implement the UI part as part of the authentication class:

https://github.com/Quansight/ragna/blob/62098457588cef258f4ea6a8b35989257559a27f/ragna/core/_authentication.py#L13

This would enable us and users to switch the authentication method smoothly.

pmeier:
We need to tackle this as this is a major blocker for anyone who wants to use Ragna in production. There is currently no way to use the UI with any login method other than username and password. However, to be able to log in with OAuth or the like, we cannot keep our current login endpoint on the REST API. Thus, we need a larger refactor here.

The scheme that I have in mind is that the API gets a `GET /login` endpoint that serves user defined HTML. From there the user can either hit `POST /login`, i.e. for form data when logging in with basic auth, or `GET /oauth-callback` if such a scheme is used. 

Internally, both endpoints funnel back into a user defined function that validates the login and returns the username that we ultimately store in our DB.

We then redirect to a yet to be designed HTML page that gives the user the option to forge a JWT that includes the user ID.

This JWT is then used as authorization header for the API as well as cookie for the UI.

pmeier:
Blocked by #417. Acceptance criteria:

- Login method must be user configurable through an `Auth` object that needs to allow basic auth (username / password) as well as OAuth
- Login must be possible regardless of the deployment mode, i.e. just REST API, just web UI, or both
- API should be accessible through an API token that can be generated after being logged in

PR #284, opened 2024-01-19
bump minimum panel version to 1.3.8

pmeier:


ragna/deploy/_ui/central_view.py
```diff
@@ -447,34 +447,29 @@ def chat_interface(self):
                     ),
                 )
             ],
+            card_params=dict(
+                stylesheets=ui.stylesheets(
```
pmeier, line 451: This is not going to work. See holoviz/panel#6240. This was "promised" in https://github.com/Quansight/ragna/pull/261#discussion_r1434489553 although the comment might also refer to making `card_params` usable in the first place. cc @ahuang11

pmeier:
As of `1.3.7rc2` there two issues that prevent adoption:

- [ ] Our auth workflow is broken. This happens because for some reason the cookie we set does not show up in `pn.state.cookies`. However, I can confirm that the cookie is indeed there through the browser console. Plus, through randomly refreshing the page and clicking login again I sometimes get past the authentication, i.e. `panel` picks up on the cookie. So far I haven't found pattern.
- [ ] When trying to send a message, we are instantly met with an error before the message is even rendered. Here is the traceback:

    ```
	  File "/home/philip/.conda/envs/ragna-dev/lib/python3.9/site-packages/panel/chat/feed.py", line 486, in _prepare_response
	    self._callback_state = CallbackState.RUNNING
	  File "/home/philip/.conda/envs/ragna-dev/lib/python3.9/contextlib.py", line 126, in __exit__
	    next(self.gen)
	  File "/home/philip/.conda/envs/ragna-dev/lib/python3.9/site-packages/param/parameterized.py", line 307, in batch_call_watchers
	    parameterized.param._batch_call_watchers()
	  File "/home/philip/.conda/envs/ragna-dev/lib/python3.9/site-packages/param/parameterized.py", line 2506, in _batch_call_watchers
	    self_._execute_watcher(watcher, events)
	  File "/home/philip/.conda/envs/ragna-dev/lib/python3.9/site-packages/param/parameterized.py", line 2468, in _execute_watcher
	    watcher.fn(*args, **kwargs)
	  File "/home/philip/.conda/envs/ragna-dev/lib/python3.9/site-packages/panel/reactive.py", line 374, in _param_change
	    self._apply_update(named_events, properties, model, ref)
	  File "/home/philip/.conda/envs/ragna-dev/lib/python3.9/site-packages/panel/reactive.py", line 302, in _apply_update
	    self._update_model(events, msg, root, model, doc, comm)
	  File "/home/philip/.conda/envs/ragna-dev/lib/python3.9/site-packages/panel/layout/base.py", line 119, in _update_model
	    super()._update_model(events, msg, root, model, doc, comm)
	  File "/home/philip/.conda/envs/ragna-dev/lib/python3.9/site-packages/panel/reactive.py", line 623, in _update_model
	    super()._update_model(events, msg, root, model, doc, comm)
	  File "/home/philip/.conda/envs/ragna-dev/lib/python3.9/site-packages/panel/reactive.py", line 318, in _update_model
	    model_val = getattr(model, attr)
	  File "/home/philip/.conda/envs/ragna-dev/lib/python3.9/site-packages/bokeh/core/has_props.py", line 357, in __getattr__
	    return super().__getattribute__(name)
	AttributeError: 'Card' object has no attribute '_callback_state'
    ```

    There is no mention of Ragna at all in the traceback so I guess this is coming from `panel`. I'll try to compile a minimal repro.

ragna/deploy/_ui/central_view.py
philippjfr: Fixed in panel 1.3.7rc3

philippjfr:
I was not able to track down any changes that could possibly have broken the auth flow. The `panel/auth` module was barely touched and the same is true of any server related code.

pmeier:
> The `panel/auth` module was barely touched and the same is true of any server related code.

Unfortunately, we are rolling our custom auth flow. There is a refactor planned (#178), but for now we are stuck with it.

Here is the idea:

1. When hitting `/` we check if we find a cookie. If that is not the case, we redirect to `/auth`
   
   https://github.com/Quansight/ragna/blob/61e8d5f6618d0533a106e5c27c350e2a2e71621e/ragna/deploy/_ui/app.py#L66-L67
2. On `/auth` we let the user sign in and if that works, we set the cookie and redirect to `/`

   https://github.com/Quansight/ragna/blob/61e8d5f6618d0533a106e5c27c350e2a2e71621e/ragna/deploy/_ui/auth_page.py#L50-L64

So far this works fine with `1.3.7`. However when we hit `/` the second time, `pn.state.cookies` still does not have the cookie we have set. But I can verify through the browser console (`document.cookie`) that the cookie is there.

So it seems however `panel` gets the cookies from the browser is somehow broken. Maybe there was a cache introduced or the like?

pmeier:
Here is the issue that we need fixed for our auth flow: holoviz/panel#6252

pmeier:
Fix for the cookie issue is available in holoviz/panel#6254. However, `1.3.7` was already released and thus we now need to wait for `1.3.8` before we can update.

nenb (approved):
LGTM!

PR #286, opened 2024-01-21
Azure OpenAI support

raphiki:
Hi

I added Azure OpenAI support. It requires three envars: AZURE_OPENAI_API_KEY, AZURE_OPENAI_API_BASE and AZURE_OPENAI_API_VERSION. 
I used them to reconstruct the Azure OpenAI endpoint with this pattern:
AZURE_OPENAI_API_BASE/openai/deployments/{self._MODEL}/completions?api_version=AZURE_OPENAI_API_VERSION

It works with my company's Azure OpenAI tenant.

Best regards,

Raphaël Semeteys

### DIFF
--- a/ragna/__init__.py
+++ b/ragna/__init__.py
@@ -1,13 +1,24 @@
-try:
-    from ._version import __version__
-except ModuleNotFoundError:
-    import warnings
+__all__ = [
+    "Claude",
+    "ClaudeInstant",
+    "Gpt35Turbo16k",
+    "Gpt4",
+    "Mpt7bInstruct",
+    "Mpt30bInstruct",
+    "RagnaDemoAssistant",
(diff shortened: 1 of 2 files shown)


pmeier (commented):
Hey Raphaël and welcome to Ragna :wave: Unfortunately, we can't accept this PR as we have a policy that only components that have no additional infrastructure requirement can make it into the core library. And since one needs to have Azure set up and in there OpenAI to be able to use this, the PR does not meet the criteria.

We (Ragna development team) currently don't have access to Azure OpenAI and thus cannot maintain it. Functionality like this is better placed in third-party extensions. Ragna makes it easy to integrate them. Here is an example for AWS: https://github.com/pmeier/ragna-aws

PR #287, opened 2024-01-22
add workflow to update the docker requirements

pmeier:
Follow-up to #283 to automatically update the requirements. Basically, we only need to dispatch the workflow and a bot will open a PR for us.


pmeier:
Failing CI is expected here, because we added a new dependency in #281 that is not yet reflected in the requirements. I'll dispatch a workflow to update the requirements when this PR is merged.

nenb (approved):
LGTM! 🚀

PR #288, opened 2024-01-22
add write permissions for GHA bot

pmeier:


pmeier:
Merging this without approval, since nothing has changed about the intentions compared to #287. I just needed to fix some details here, since they were not tested in the original PR.

PR #285, opened 2024-01-19
Add file selector widget

nenb:
WIP

PR #295, opened 2024-01-22
Add env var for customising location of static assets

nenb:


nenb:
Closing as I think it's better to keep this in #285 for now.

Issue #277, opened 2024-01-17
Increase API timeout and add custom message


pmeier:
We have a timeout behavior in place when the API is started alongside the UI:

https://github.com/Quansight/ragna/blob/38f75f6d88e74f72c030ef50045315b81eb203aa/ragna/deploy/_cli/core.py#L132-L139

This is a good thing so the user doesn't waits forever for the UI to come up when it is actually the API that is failing to start. However 30 seconds, i.e. the default,

https://github.com/Quansight/ragna/blob/38f75f6d88e74f72c030ef50045315b81eb203aa/ragna/_utils.py#L76-L78

might be a little to low. We should bump to 60 seconds to avoid killing the UI when a slower box actually needs more time to start the API. Plus, we should have a custom message like "Failed to start the API" to make it more obvious what happened.

smokestacklightnin:
@pmeier Does PR #299 accomplish what you were looking for? Thanks

PR #299, opened 2024-01-29
Increase API timeout and add custom message

smokestacklightnin:
Issue #277 asked for two small things:

1. Increase time to timeout from 30 seconds to 60 seconds
2. Add clarification to error message when timeout happens

This small pull request does those two things. 

Please let me know if I should add/change anything. All constructive feedback is welcome.

### DIFF
--- a/ragna/deploy/_cli/core.py
+++ b/ragna/deploy/_cli/core.py
@@ -131,15 +131,19 @@ def ui(
     try:
         if process is not None:
 
-            @timeout_after(
-                60,
-                'Failed to start the API in 60 seconds. Please start it manually with the command "ragna api".',
-            )
+            @timeout_after(60)
             def wait_for_api() -> None:
                 while not check_api_available():
                     time.sleep(0.5)
 
-            wait_for_api()
+            try:
+                wait_for_api()
+            except TimeoutError:
+                rich.print(
+                    "Failed to start the API in 60 seconds. "
+                    "Please start it manually with [bold]ragna api[/bold]."
+                )
+                raise typer.Exit(1)
 
         ui_app(config).serve()  # type: ignore[no-untyped-call]
     finally:
--- a/ragna/deploy/_cli/core.py
+++ b/ragna/deploy/_cli/core.py
@@ -133,7 +133,7 @@ def ui(
 
             @timeout_after(
                 60,
-                "Failed to start the API in 60 seconds. Please start it manually with ragna API.",
+                'Failed to start the API in 60 seconds. Please start it manually with the command "ragna api".',
             )
             def wait_for_api() -> None:
                 while not check_api_available():
--- a/ragna/deploy/_cli/core.py
+++ b/ragna/deploy/_cli/core.py
@@ -136,7 +136,10 @@ def ui(
                 while not check_api_available():
                     time.sleep(0.5)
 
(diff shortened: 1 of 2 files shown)


pmeier (changes requested):
Hey @smokestacklightnin and thanks for the PR! You have added these changes to the function that is used in multiple place. Instead, I was looking for parametrizing the `timeout_after` call, i.e. calling it with a custom timeout and message.

smokestacklightnin:
@pmeier very sorry for the rookie mistake! Does it look better now?

ragna/deploy/_cli/core.py
```diff
@@ -136,7 +136,10 @@ def wait_for_api() -> None:
                 while not check_api_available():
                     time.sleep(0.5)
 
-            wait_for_api()
+            wait_for_api(
```
pmeier, line 139: `wait_for_api` doesn't take any parameters. You have to pass them to the `@timeout_after` decorator.

smokestacklightnin: duh, sorry

ragna/deploy/_cli/core.py
```diff
@@ -136,7 +136,10 @@ def wait_for_api() -> None:
...
+                60,
+                "Could not launch the UI. This is likely due to a failure to start the API.",
```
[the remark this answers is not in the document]
smokestacklightnin: Made this change

ragna/deploy/_cli/core.py
```diff
@@ -131,7 +131,10 @@ def check_api_available() -> bool:
     try:
         if process is not None:
 
-            @timeout_after()
+            @timeout_after(
+                60,
+                "Failed to start the API in 60 seconds. Please start it manually with ragna API.",
```
pmeier, line 136: The lower casing here was intentional as it refers to the command the user should run.

```suggestion
                "Failed to start the API in 60 seconds. Please start it manually with ragna api.",
```

That being said, I dislike that it looks exactly like the other text and thus is hard to interpret. One idea could be to make it bold. What do you think? I can guide you through the process if you are up to it.

smokestacklightnin: I will change the capitalization.

Bold is one good idea. As another example, git puts commands in quotes like 

> `(use "git push" to publish your local commits)`

to indicate commands. I think bolded with quotes would be a good idea because the bold draws attention, and quotes explicitly tell the user what command to run. 

Your guidance for the bold text would be appreciated.

pmeier: Both quoting the command or making it bold add emphasis and I prefer not to have two ways of doing the same. To make it bold, we need print it with `rich`. To be able to do that, we first need to except the `TimeoutError`

```python
try:
    wait_for_api()
except TimeoutError:
    rich.print(
        "Failed to start the API in 60 seconds. "
        "Please start it manually with [bold]ragna api[/bold]."
    )
    raise typer.Exit(1)
```

With this we can also remove the custom message from the decorator since we ignore it anyway.

smokestacklightnin: Let me know what you think of the new changes

pmeier (approved):
Thanks for the patience @smokestacklightnin! LGTM.

PR #300, opened 2024-01-30
Add git-lfs for tracking images

### LINKED ISSUE #247: How to handle images as part of the documentation?
Currently we are storing images as part of the `git` repository. For example

- https://github.com/Quansight/ragna/blob/main/docs/assets/brand/logo-lockup-horizontal/logo-lockup-horizontal.png
- https://github.com/Quansight/ragna/blob/main/docs/how-tos/images/ragna-config-wizard.png

The former is an example where this fine, since the logo (or the brand in general), will change very rarely if ever.

The latter however is not a good use. A screenshot of components is quickly outdated (see #182) and needs replacing. Since an image is binary file, there is no diff that `git` could track. Meaning, whenever we update an image, we are completely replacing the old one. Ultimately this quickly leads to really large git repositories that I want to avoid.

I haven't done a proper survey on this. These are the options that I'm currently aware of:

1. Just ignore the issue. For active maintainers this is not really an issue, because when you have an initial checkout of the repository locally, all other changes are incremental. "Only" new contributors and our CI will be hit by this.
2. Use [`git-lfs`](https://git-lfs.com/) to store the images. I've started that in #168, but faced some (internal) opposition. The main argument against it was that [`napari`](https://napari.org/stable/) [migrated away from it](https://github.com/napari/napari/pull/5216) after facing [some issues](https://github.com/napari/napari/issues/4115). 
3. Go the `napari` way and create a [separate documentation repository](https://github.com/napari/docs) that holds the whole documentation except for the API documentation, which stays in the main repository as part of the docstrings.

My current order of preference 3. -> 2. -> 1. Opinions welcome.

pavithraes:
Closes #247

In the current setup, all images, gifs, etc. will be tracked by `git-lfs` moving forward.

If we want to override the git history and track previously committed files, we need to [`migrate` repo data](https://github.com/git-lfs/git-lfs/wiki/Tutorial#migrating-existing-repository-data-to-lfs). @pmeier would you want to migrate?

TODOs before merging:

- [ ] Add `git-lfs` to `environment-dev.yml`
- [ ] Make a note to `git lfs pull` in https://ragna.chat/en/stable/community/contribute/#contribute-documentation
- [ ] Update RTD build config


environment-dev.yml
```diff
@@ -4,6 +4,7 @@ channels:
 dependencies:
   - python =3.9
   - pip
+  - git-lfs
```
pavithraes, line 7: The [conda-forge distribution](https://anaconda.org/conda-forge/git-lfs) has the latest versions, [PyPI releases](https://pypi.org/project/git-lfs/) are outdated/unmaintained for this project.

pmeier (commented):
1. Could we keep all images in `docs/assets/brand` as regular images? They are tiny and not bound to change regularly. Plus, I don't know how well this works with other tools like building a wheel etc.
2. We also need to have a configuration for RTD, no? In #168 I followed https://docs.readthedocs.io/en/stable/build-customization.html#support-git-lfs-large-file-storage

docs/assets/brand/logo-lockup-horizontal/logo-lockup-horizontal.png
pavithraes, line 1: I don't see these tracked locally, but still see the BIN files here. :/

I'll close this PR and open a fresh one. It'll be much quicker.

PR #301, opened 2024-01-31
add Google assistants

pmeier:
This adds the Gemini Pro and Ultra models from Google to the builtin assistants. To achieve that I needed to do some minor refactoring. I'll explain it below.

### DIFF
--- a/docs/references/faq.md
+++ b/docs/references/faq.md
@@ -39,3 +39,11 @@
    ```bash
    export MOSAICML_API_KEY="XXXXX"
    ```
+
+### [Google](https://ai.google.dev/)
+
+1. ADDME
+2. Set the `GOOGLE_API_KEY` environment variable with your Google API key:
+   ```bash
+   export GOOGLE_API_KEY="XXXXX"
+   ```
--- a/docs/references/faq.md
+++ b/docs/references/faq.md
@@ -42,8 +42,10 @@
 
 ### [Google](https://ai.google.dev/)
 
-1. Get a Google AI Studio API key following these instructions:
-First, create a Google AI Studio account or sign in. Next, navigate to the API key page and "Create API key in new/exisiting project", optionally naming the key. Do not share it with anyone.
+1. > To use the API, you need an API key. You can create a key with one click in Google
+   > AI Studio.
+   >
+   > ~ [Google AI API](https://ai.google.dev/tutorials/setup)
 2. Set the `GOOGLE_API_KEY` environment variable with your Google API key:
    ```bash
    export GOOGLE_API_KEY="XXXXX"
--- a/docs/references/faq.md
+++ b/docs/references/faq.md
@@ -42,7 +42,8 @@
 
 ### [Google](https://ai.google.dev/)
 
-1. ADDME
+1. Get a Google AI Studio API key following these instructions:
+First, create a Google AI Studio account or sign in. Next, navigate to the API key page and "Create API key in new/exisiting project", optionally naming the key. Do not share it with anyone.
 2. Set the `GOOGLE_API_KEY` environment variable with your Google API key:
    ```bash
    export GOOGLE_API_KEY="XXXXX"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -139,12 +139,14 @@ disallow_incomplete_defs = false
 
 [[tool.mypy.overrides]]
 module = [
+    "docx",
     "fitz",
+    "json_stream",
+    "json_stream.httpx",
     "lancedb",
     "param",
-    "pyarrow",
-    "docx",
     "pptx",
+    "pyarrow",
     "sentence_transformers",
 ]
 ignore_missing_imports = true
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -25,7 +25,7 @@ dependencies = [
     "emoji",
     "fastapi",
     "httpx",
-    "httpx_sse",
+    "httpx-sse",
     "importlib_metadata>=4.6; python_version<'3.10'",
     "packaging",
     "panel>=1.3.6,<1.4",
@@ -56,6 +56,7 @@ Repository = "https://github.com/Quansight/ragna"
 # to update the array below, run scripts/update_optional_dependencies.py
 all = [
     "chromadb>=0.4.13",
+    "json-stream",
     "lancedb>=0.2",
     "pyarrow",
     "pymupdf>=1.23.6",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -142,8 +142,7 @@ disallow_incomplete_defs = false
 module = [
     "docx",
     "fitz",
-    "json_stream",
-    "json_stream.httpx",
+    "ijson",
     "lancedb",
     "param",
     "pptx",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -56,7 +56,7 @@ Repository = "https://github.com/Quansight/ragna"
 # to update the array below, run scripts/update_optional_dependencies.py
 all = [
     "chromadb>=0.4.13",
-    "json-stream",
+    "ijson",
     "lancedb>=0.2",
     "pyarrow",
     "pymupdf>=1.23.6",
--- a/ragna/_compat.py
+++ b/ragna/_compat.py
@@ -53,9 +53,22 @@ def _anext() -> Callable[[AsyncIterator[T]], Awaitable[T]]:
     if sys.version_info[:2] >= (3, 10):
         anext = builtins.anext
     else:
+        sentinel = object()
 
-        async def anext(ait: AsyncIterator[T]) -> T:
-            return await ait.__anext__()
+        def anext(
+            ait: AsyncIterator[T],
+            default: T = sentinel,  # type: ignore[assignment]
+        ) -> Awaitable[T]:
+            if default is sentinel:
+                return ait.__anext__()
+
+            async def anext_with_default() -> T:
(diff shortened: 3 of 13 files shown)


pyproject.toml
```diff
@@ -139,12 +139,14 @@ disallow_incomplete_defs = false
 
 [[tool.mypy.overrides]]
 module = [
+    "docx",
     "fitz",
+    "json_stream",
+    "json_stream.httpx",
```
pmeier, line 145: These are the new entries. Everything else is just sorting.

ragna/core/_rag.py
```diff
@@ -256,34 +256,19 @@ def _unpack_chat_params(
...
         kwargs = self._unpacked_params[fn]
-        if inspect.iscoroutinefunction(fn):
-            fn = cast(Callable[..., Awaitable[T]], fn)
-            coro = fn(*args, **kwargs)
-        else:
-            fn = cast(Callable[..., T], fn)
-            coro = run_in_threadpool(fn, *args, **kwargs)
-
-        return await coro
+        return as_awaitable(fn, *args, **kwargs)
```
pmeier, line 263: Previously, only `Chat` needed the ability to turn arbitrary functions into awaitables. Now, we also need this for the builtin assistants. Thus, this was factored out into `ragna._utils`. The same is true for (async) iterators below.

ragna/assistants/_google.py
```diff
@@ -0,0 +1,95 @@
...
+        return "\n".join(
+            [
+                "Answer the prompt using only the pieces of context below.",
+                "If you don't know the answer, just say so. Don't try to make up additional context.",
+                f"Prompt: {prompt}",
+                *[f"\n{source.content}" for source in sources],
+            ]
+        )
+
+    def _call_api(
```
pmeier, line 39: This cannot be an async function since the `json_stream` library that we need to parse the response cannot handle async streams.

ragna/assistants/_api.py
```diff
@@ -16,22 +17,28 @@ def requirements(cls) -> list[Requirement]:
...
-        self._client = httpx.AsyncClient(
+        self._api_key = os.environ[self._API_KEY_ENV_VAR]
+
+        kwargs: dict[str, Any] = dict(
             headers={"User-Agent": f"{ragna.__version__}/{self}"},
             timeout=60,
         )
-        self._api_key = os.environ[self._API_KEY_ENV_VAR]
+        self._sync_client = httpx.Client(**kwargs)
+        self._async_client = httpx.AsyncClient(**kwargs)
```
pmeier, line 27: Since the Google assistants need a sync client, I've renamed the `._client` attribute to `._async_client` to better reflect its nature.

ragna/assistants/_api.py
```diff
@@ -16,22 +17,28 @@ def requirements(cls) -> list[Requirement]:
...
 
     async def answer(
         self, prompt: str, sources: list[Source], *, max_new_tokens: int = 256
     ) -> AsyncIterator[str]:
-        async for chunk in self._call_api(  # type: ignore[attr-defined, misc]
-            prompt, sources, max_new_tokens=max_new_tokens
+        async for chunk in as_async_iterator(
```
pmeier, line 32: Since `self._call_api` can now return either a sync or and async iterator, we force it into an async iterator here.

ragna/assistants/_api.py
```diff
@@ -16,22 +17,28 @@ def requirements(cls) -> list[Requirement]:
...
+            self._call_api,
+            prompt,
+            sources,
+            max_new_tokens=max_new_tokens,
         ):
             yield chunk
 
     @abc.abstractmethod
-    async def _call_api(
+    def _call_api(
```
pmeier, line 41: This makes no functional difference. We just use the sync prototype in other places when the function can be overwritten as sync or async.

ragna/deploy/_ui/api_wrapper.py
```diff
@@ -18,7 +18,7 @@ class ApiWrapper(param.Parameterized):
     auth_token = param.String(default=None)
 
     def __init__(self, api_url, **params):
-        self.client = httpx.AsyncClient(base_url=api_url)
+        self.client = httpx.AsyncClient(base_url=api_url, timeout=60)
```
pmeier, line 21: Even when streaming, the Google assistants return really large chunks and thus easily go over the default timeout. The new timeout is in line with what we use for our builtin assistants as well:

https://github.com/Quansight/ragna/blob/1b53e62cb873ba16968b7c25a81baff0947d705c/ragna/assistants/_api.py#L19-L22

docs/references/faq.md
```diff
@@ -39,3 +39,11 @@
    ```bash
    export MOSAICML_API_KEY="XXXXX"
    ```
+
+### [Google](https://ai.google.dev/)
+
+1. ADDME
```
pmeier, line 45: @nenb Could you fill out this similar to what we did for the others?

nenb:
Why did you choose `json-stream` rather than `ijson` for the streaming package? `ijson` has async support currently, which may make it more attractive for `ragna`.

pmeier:
Two reasons:

1. I didn't find it when searching for a JSON stream parser. But that might be on me. `ijson` is download [7M/month](https://pypistats.org/packages/ijson) while `json-stream` sits at "only" [80k/month](https://pypistats.org/packages/json-stream). Meaning, `ijson` is certainly the more popular choice.
2. After looking into it, one issue that we need to tackle is that `ijson` currently doesn't work with generators (sync and async) as source (ICRAR/ijson#44). While inconvenient, this is not terribly bad and we can emulate a file-like object. I've pushed 7d0d2403af4f4ba8d7e1c75edf7aebd558007890 so we can have a look together if it is worthwhile to use. What I didn't do in the commit is rolling back all the other changes. That would come on top, but is a benefit for `ijson` IMO.

ragna/_compat.py
```diff
@@ -53,9 +53,22 @@ def _anext() -> Callable[[AsyncIterator[T]], Awaitable[T]]:
     if sys.version_info[:2] >= (3, 10):
         anext = builtins.anext
     else:
-
-        async def anext(ait: AsyncIterator[T]) -> T:
-            return await ait.__anext__()
+        sentinel = object()
```
pmeier, line 56: This is a slight refactor as I needed the default return value in case of exhaustion.

pmeier:
I've opted to go with `ijson` now. Proving a small wrapper object to turn the generator into a file-like object is quite a bit simpler than the larger refactor I had in mind. The generic move from `Chat._run` / `Chat._run_gen` to `as_awaitable` / `as_async_iterator` will happen eventually, since I'm sure we need this functionality elsewhere. But now is not the time.

ragna/_compat.py
```diff
@@ -53,9 +53,22 @@ def _anext() -> Callable[[AsyncIterator[T]], Awaitable[T]]:
...
+
+        def anext(
+            ait: AsyncIterator[T],
+            default: T = sentinel,  # type: ignore[assignment]
+        ) -> Awaitable[T]:
+            if default is sentinel:
+                return ait.__anext__()
```
nenb, line 63: I don't fully grok this. Do we not need to `await` this? And in what situation will this arise?

docs/references/faq.md
nenb, line 45:
```suggestion
1. Get a Google AI Studio API key following these instructions:
First, create a Google AI Studio account or sign in. Next, navigate to the API key page and "Create API key in new/exisiting project", optionally naming the key. Do not share it with anyone.
```

nenb (approved):
LGTM! Nice work on this, I learned a bunch. I left some comments that you might want to address, but don't feel like they are preventing a merge.

ragna/assistants/_google.py
```diff
@@ -0,0 +1,117 @@
...
+                    for category in [
+                        "HARASSMENT",
+                        "HATE_SPEECH",
+                        "SEXUALLY_EXPLICIT",
+                        "DANGEROUS_CONTENT",
+                    ]
+                ],
+                # https://ai.google.dev/tutorials/rest_quickstart#configuration
+                "generationConfig": {
+                    "temperature": 0.0,
```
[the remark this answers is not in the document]
pmeier: It is the other way around: we want to hardcode `0.0` here, because that means determinism. If we learned one thing from trying to bring RAG to businesses is that they want to get exactly the same answer if they ask the same question twice. Of course we can't guarantee it since we don't control the model, but we can do our best to at least avoid sampling during generation.

This is the same for all other assistants that we currently have

https://github.com/Quansight/ragna/blob/1b53e62cb873ba16968b7c25a81baff0947d705c/ragna/assistants/_anthropic.py#L52
https://github.com/Quansight/ragna/blob/1b53e62cb873ba16968b7c25a81baff0947d705c/ragna/assistants/_mosaicml.py#L43
https://github.com/Quansight/ragna/blob/1b53e62cb873ba16968b7c25a81baff0947d705c/ragna/assistants/_openai.py#L58

ragna/_compat.py
pmeier: > I don't fully grok this. Do we not need to `await` this?

We don't and we actually can't here. Note that `anext` is *not* an `async def` function. `anext` returns an awaitable, e.g.

```python
async_iterator = ...
awaitable = anext(async_iterator)
result = await awaitable
```

> And in what situation will this arise?

This is the default case, i.e. no `default` value is set. Let me refactor this function to make it more clear to what is going on.

PR #305, opened 2024-02-02
Relative paths in the API wrapper

aktech:
- Fixes relative paths in the API wrapper
- Adds a `proxy_url`: for deployed environments, where the app is created on a particular port and deployed on a different url via a proxy.


pmeier (approved):
IIUC, the only change is that we append the root path to the API URL for the UI, correct?

I'm ok with merging as is, but I would prefer to keep the diff smaller and just fix the `api_url` that we pass to `App`.

aktech:
closing in favor of #346

Issue #227, opened 2023-12-04
[ENH] - add support for Cohere assistants


pmeier:
### Feature description

Per title. Sign up on https://www.cohere.com/, get an API key and build a new `CohereAssistant(ApiAssistant)` similar to what we have for OpenAI:

https://github.com/Quansight/ragna/blob/0e1c35a20259fc55615ef303e7c90a9700eb377b/ragna/assistants/_openai.py#L8

### Value and/or benefit

One can get API keys from them without any waitlist. Thus, this would be one more assistant apart from OpenAI that users can try.

pmeier:
Closed in #307.

PR #307, opened 2024-02-02
[ENH] - add support for Cohere assistants

smokestacklightnin:
This addresses Issue #227 

As always, comments and change requests are welcome and encouraged

### DIFF
--- a/docs/references/faq.md
+++ b/docs/references/faq.md
@@ -50,3 +50,14 @@
    ```bash
    export GOOGLE_API_KEY="XXXXX"
    ```
+
+### [Cohere](https://cohere.com/)
+
+1. > To use the API, you need an API key. You can create a key in the Cohere dashboard
+   > using your Cohere account
+   >
+   > ~ [Cohere Dashboard](https://dashboard.cohere.com/api-keys)
+2. Set the `COHERE_API_KEY` environment variable with your Cohere API key:
+   ```bash
+   export COHERE_API_KEY="XXXXX"
+   ```
--- a/docs/references/faq.md
+++ b/docs/references/faq.md
@@ -54,7 +54,7 @@
 ### [Cohere](https://cohere.com/)
 
 1. > To use the API, you need an API key. You can create a key in the Cohere dashboard
-   > using your Cohere account
+   > using your Cohere account.
    >
    > ~ [Cohere Dashboard](https://dashboard.cohere.com/api-keys)
 2. Set the `COHERE_API_KEY` environment variable with your Cohere API key:
--- a/ragna/assistants/_cohere.py
+++ b/ragna/assistants/_cohere.py
@@ -9,3 +9,7 @@ class CohereApiAssistant(ApiAssistant):
     @classmethod
     def display_name(cls) -> str:
         return f"Cohere/{cls._MODEL}"
+
+    @property
+    def max_input_size(self) -> int:
+        return self._CONTEXT_SIZE
--- a/ragna/assistants/_cohere.py
+++ b/ragna/assistants/_cohere.py
@@ -1,0 +1,7 @@
+from ._api import ApiAssistant
+
+
+class CohereApiAssistant(ApiAssistant):
+    _API_KEY_ENV_VAR = "COHERE_API_KEY"
+    _MODEL: str
+    _CONTEXT_SIZE: int
--- a/ragna/assistants/_cohere.py
+++ b/ragna/assistants/_cohere.py
@@ -5,3 +5,7 @@ class CohereApiAssistant(ApiAssistant):
     _API_KEY_ENV_VAR = "COHERE_API_KEY"
     _MODEL: str
     _CONTEXT_SIZE: int
+
+    @classmethod
+    def display_name(cls) -> str:
+        return f"Cohere/{cls._MODEL}"
--- a/ragna/assistants/_cohere.py
+++ b/ragna/assistants/_cohere.py
@@ -13,3 +13,11 @@ class CohereApiAssistant(ApiAssistant):
     @property
     def max_input_size(self) -> int:
         return self._CONTEXT_SIZE
+
+    def _make_system_content(self) -> str:
+        instruction = (
+            "You are a helpful assistant that answers user questions given the included context. "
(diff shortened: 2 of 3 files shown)


ragna/assistants/_cohere.py
```diff
@@ -0,0 +1,81 @@
...
+        document_sources = [
+            {"title": source.id, "snippet": source.content} for source in sources
+        ]
+        return document_sources
+
+    async def _call_api(
+        self, prompt: str, sources: list[Source], *, max_new_tokens: int
+    ) -> AsyncIterator[str]:
+        response = await self._client.post(
+            "https://api.cohere.ai/v1/chat",
```
pmeier, line 39: Cohere API is able to stream the response. This results in a better UX and we should always do that if the API supports it. From the [wording](https://docs.cohere.com/reference/chat), I guess they are using SSE to stream

> the response will be a JSON stream of events

If so, you can have a look at the OpenAI or Anthropic assistant on how to do that. If they for some reason mean that they actually respond with a JSON stream, have a look at the Google assistant on how to handle this.

smokestacklightnin: It's a JSON stream. I'm looking now how to do it

smokestacklightnin:
Quick status update

The code seems to receive the JSON stream http responses without error, but is not parsing them properly. No error is reported in the terminal console when running the UI, but the UI response says that something went wrong. I believe this has to do with the type of [streamedchatresponse](https://docs.cohere.com/reference/chat). I'll do more work to try to fix this.

Also, mypy complains when I run it after adapting [`AsyncIteratorReader`](https://github.com/Quansight/ragna/blob/8aa6bd09740be0bed6391fab00323caf4889f369/ragna/assistants/_google.py#L14-L25) from the google assistant code. I tried running mypy on that file too, and got similar complaints. Does anyone have any input?

pmeier:
@smokestacklightnin I took the liberty and had a look to unblock you. I've pushed bfc868be8ecb590c90fe467daf2a62a7d00c466e that implements the streaming. While Cohere also uses a streaming HTTP request, in contrast to Google they don't stream JSON, but rather [JSONL](https://jsonlines.org/). Meaning, each line is valid JSON. Thus, we don't need an extra library to resolve the stream, but rather just ingest the stream line by line and convert to JSON ourselves.

smokestacklightnin:
> @smokestacklightnin I took the liberty and had a look to unblock you. I've pushed [bfc868b](https://github.com/Quansight/ragna/commit/bfc868be8ecb590c90fe467daf2a62a7d00c466e) that implements the streaming. While Cohere also uses a streaming HTTP request, in contrast to Google they don't stream JSON, but rather [JSONL](https://jsonlines.org/). Meaning, each line is valid JSON. Thus, we don't need an extra library to resolve the stream, but rather just ingest the stream line by line and convert to JSON ourselves.

Thanks for the help! I'll continue working on this

smokestacklightnin:
I added a `preamble_override`, which seems to be the last substantial change that is needed. I will then add links in the code and the FAQ documentation

docs/references/faq.md
```diff
@@ -50,3 +50,14 @@
    ```bash
    export GOOGLE_API_KEY="XXXXX"
    ```
+
+### [Cohere](https://cohere.com/)
+
+1. > To use the API, you need an API key. You can create a key in the Cohere dashboard
+   > using your Cohere account
```
pmeier, line 57:
```suggestion
   > using your Cohere account.
```

ragna/assistants/_cohere.py
```diff
@@ -0,0 +1,92 @@
...
+
+    @classmethod
+    def display_name(cls) -> str:
+        return f"Cohere/{cls._MODEL}"
+
+    @property
+    def max_input_size(self) -> int:
+        return self._CONTEXT_SIZE
+
+    def _make_system_content(self) -> str:
```
pmeier, line 23: Let's call this 

```suggestion
    def _make_preamble(self) -> str:
```

to align our terminology here with Cohere. We can't really align across assistants built into Ragna given that every provider uses something different.

ragna/assistants/_cohere.py
```diff
@@ -0,0 +1,92 @@
...
+        instruction = (
+            "You are a helpful assistant that answers user questions given the included context. "
+            "If you don't know the answer, just say so. Don't try to make up an answer. "
+            "Only use the included documents below to generate the answer."
+        )
+        return instruction
```
pmeier, line 29: No need to store the string. Just return it.

```suggestion
        return (
            "You are a helpful assistant that answers user questions given the included context. "
            "If you don't know the answer, just say so. Don't try to make up an answer. "
            "Only use the included documents below to generate the answer."
        )
```

ragna/assistants/_cohere.py
```diff
@@ -0,0 +1,92 @@
...
+
+    def _make_source_documents(self, sources: list[Source]) -> list[dict[str, str]]:
+        document_sources = [
+            {"title": source.id, "snippet": source.content} for source in sources
+        ]
+        return document_sources
```
pmeier, line 35: Same as above

```suggestion
        return [
            {"title": source.id, "snippet": source.content} for source in sources
        ]
```

pmeier (commented):
Only a few minor comments left. Otherwise this is good to go. Thanks @smokestacklightnin!

docs/references/faq.md
smokestacklightnin: Done

ragna/assistants/_cohere.py
smokestacklightnin: Done

ragna/assistants/_cohere.py
smokestacklightnin: Done

ragna/assistants/_cohere.py
smokestacklightnin: Done

smokestacklightnin:
> Only a few minor comments left. Otherwise this is good to go. Thanks @smokestacklightnin!

Everything is done. Glad to contribute!

pmeier (approved):

PR #308, opened 2024-02-02
remove invalid font CSS declaration

### LINKED ISSUE #230: Stylesheet is not loaded because of wrong MIME type
### Bug description

I'm seeing the following error in the developer console:

```
The stylesheet https://rsms.me/ was not loaded because its MIME type, “text/html”, is not “text/css”.
```

### How to reproduce the bug?

Open the developer tools after accessing the Ragna web UI.

Browser: Firefox 116.0.2 (64-bit)

### Anything else?

Can't reproduce on Chromium

pmeier:
Fixes #230. https://rsms.me/ does not contain any CSS. In the [documentation](https://rsms.me/inter/#usage) it is listed as "preconnect", i.e. just telling the browser to [connect to this host as soon as possible rather than only connecting on demand](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/rel/preconnect).


pmeier:
The failures here are unrelated. `lancedb==5.0.2` was released an hour ago and that now depends on polars. However, our envs are cached on a daily basis so `polars` was not pulled in yet.

pierrotsmnrd (approved):
looks good to me

PR #312, opened 2024-02-03
CSS refactor

pierrotsmnrd:
~Work in progress - do not merge~ Ready for review

This PR uses a better approach for CSS styling in Panel.

The method we used so far showed its limitations : too much CSS code inside the python code, preventing CSS code completion, linting breaking indentation, hard to maintain.

The new method relies on design *modifiers*.
- A modifier allows to register a CSS file or raw CSS code, for a given panel widget. eg : 
```python 
# modifies the `stylesheets` property of *all* buttons, to add one css file
pn.theme.fast.Fast.modifiers[pn.widgets.Button] = {
"stylesheets":["css/button.css"]
}
```
- One key thing then is to use css class (using `css_classes=['my_class_name', ]`). Imagine we have 2 kinds of buttons, A and B, we want to style in a different way. We can add them css class `my_button_A` and `my_button_B`, do the following : 
```python 
pn.theme.fast.Fast.modifiers[pn.widgets.Button] = {
"stylesheets":["css/button_A.css", "css/button_B.css"]
}
```
and each css file should contain rules starting with : 
``` 
:host(.my_button_A) { 
  /* Rules for button type A*/
}
``` 
and similarly for button B 

This PR applies to method to Ragna UI. 
Currently WIP, the auth page and part of the main view are handled.

Caveat : not all CSS can be removed from python code.
- some CSS rules need to be built from python variables  ( see central_view.py:L307)
- some widgets seem immune to modifiers, depending on how they're built. For instance, all the widgets from the ChatInterface like ChatMessage, and "source info" buttons, cannot be styled with modifiers. I think it's a bug in Panel and I need to dig to identify the root cause. Eventually every widget should be modifiable using modifiers.


pierrotsmnrd:
> For these, could you use the functionality I introduced https://github.com/Quansight/ragna/pull/257?
Yes that's what I'm doing, see [here](https://github.com/Quansight/ragna/blob/200afca4837443dd88f0b8b3781026492dcc53b1/ragna/deploy/_ui/central_view.py#L304)

> I believe this was fixed in https://github.com/holoviz/panel/pull/6304. Should we wait for this to be released before we merge this PR?
I confirm this fixes the issue. Yes, we will have to wait for it to be released before merging.

ragna/deploy/_ui/central_view.py
```diff
@@ -429,50 +420,11 @@ def chat_interface(self):
...
-                            },
-                        ),
-                        (
-                            ":host input[type='text']:focus",
-                            {
-                                "box-shadow": "0px 0px 8px 0px rgba(0, 0, 0, 0.3)",
-                            },
-                        ),
-                    ),
+                    # css_classes is forced to chat-interface-input-widget by Panel
```
pmeier, line 423: Should we report this? Can't we add an additional CSS class with whatever name we want?

ragna/deploy/_ui/styles.py
```diff
@@ -11,109 +11,69 @@ def divider():
...
-                            in some cases.
-                            */
-                            :host(.source-content) p:nth-of-type(1)  {
-                                margin-top: 0px;
-                            }
-                        """
-        ]
-    }
+    add_modifier(
+        pn.layout.Accordion, " :host(.source-accordion) { height: 100%; width: 100%; } "
```
pmeier, line 49: Should this also go into a CSS file? If not, let's use our helper function.

ragna/deploy/_ui/styles.py
```diff
@@ -11,109 +11,69 @@ def divider():
...
-    # add here calls to other design modifiers,
-    #   group them per UI component
+    apply_design_modifiers_auth_page()
+    apply_design_modifiers_central_view()
+    apply_design_modifiers_chat_info()
+    apply_design_modifiers_chat_interface()
+
+
+def add_modifier(
+    modifier_class: Type[Any], modifications: Any, property: str = "stylesheets"
```
pmeier, line 23: Are we ever passing something for `property`? If not can we just hardcode it and maybe rename the function `add_stylesheets_modifier`?

ragna/deploy/_ui/styles.py
```diff
@@ -11,109 +11,69 @@ def divider():
...
+    if property not in pn.theme.fast.Fast.modifiers[modifier_class]:
+        pn.theme.fast.Fast.modifiers[modifier_class] = {property: [modifications]}
+    else:
+        pn.theme.fast.Fast.modifiers[modifier_class][property].append(modifications)
+
+
+def apply_design_modifiers_global():
+    add_modifier(
+        pn.widgets.TextInput,
+        """ .bk-input {border-color: var(--neutral-color) !important;} """,
```
pmeier, line 37: CSS file or helper function.

ragna/deploy/_ui/styles.py
```diff
@@ -11,109 +11,69 @@ def divider():
...
+    )
+    add_modifier(
+        pn.widgets.Select,
+        """ .bk-input {border-color: var(--neutral-color) !important;} """,
```
pmeier, line 41: CSS file or helper function.

pierrotsmnrd:
> I see the styles.py file still has massive CSS stylesheets. Can they be moved to CSS files as well?
The goal is to get rid of all the CSS currently in python files and move that in proper CSS files. So yes, they'll be moved.

> Since we now only rarely need to use ui.stylesheets (basically only if the style is only known at runtime), should we simplify the helper function? I'm thinking of it only accepting one class selector at a time.
Totally ok with the concept, but I suggest to name it `ui.css`, it's shorter but conveys the same meaning.

ragna/deploy/_ui/app.py
```diff
@@ -24,7 +24,7 @@
 
 
 HERE = Path(__file__).parent
-# CSS = HERE / "css"
+CSS = HERE / "css"
```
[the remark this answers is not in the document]
pierrotsmnrd: I committed your suggested change. Earlier in the project, having these paths in variables made sense as things were still prone to change, but now we can definitely do that way.

ragna/deploy/_ui/central_view.py
```diff
@@ -161,24 +161,36 @@ def __init__(
         self._stylesheets.extend(message_stylesheets)
 
     def _copy_and_source_view_buttons(self) -> pn.Row:
+        # I have tried to style this button using modifiers, but it doesn't work.
```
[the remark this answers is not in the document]
pierrotsmnrd: Fixed now

ragna/deploy/_ui/styles.py
pierrotsmnrd: For the sake of keeping as much CSS as possible out of the python code, this should move into CSS files. I was hesitating at first because that's a very short snippet, but it's actually the right thing to do considering the goal. 
I'll also move the raw CSS from `apply_design_modifiers_global` into CSS files, that's the same pattern.

ragna/deploy/_ui/styles.py
pierrotsmnrd: We're not yet, but changes are high that we will. Modifiers are very powerful, we can override any property of a Panel widget/viewable object. `stylesheets`, but also `css_classes`, or `margin`, or whatever. I'm unsure which we'll use in the future, as I really heavily on `stylesheets` as it's the most versatile, but I'd like to keep this option.

ragna/deploy/_ui/styles.py
pierrotsmnrd: will move to CSS file as indicated here https://github.com/Quansight/ragna/pull/312#discussion_r1481478256

ragna/deploy/_ui/styles.py
pierrotsmnrd: will move to CSS file as indicated here https://github.com/Quansight/ragna/pull/312#discussion_r1481478256

ragna/deploy/_ui/central_view.py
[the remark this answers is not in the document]
pmeier: Well, yeah, our side is fixed now. But this makes this PR dependent on `panel==1.4`. Let's leave it open until we can have that as well.

pierrotsmnrd:
Current state of things : 

- 💥 **All the CSS is now out of the python code**  (except for a few rare places where it makes sense. I've left a comment each time)
- 🚀  I have added a mechanism to preload all the CSS files. This fixes a flashy behavior where unstyled components appears for a brief second when switching between chats. It's a bit hacky but couldn't find alternative. it's commented and explained in `app.py`
- ⚙️ I have simplified and factorized the code to load all the CSS files of modifiers. it's now leaner, and it's a result of the need for the CSS preload. I have also changed `ui.stylesheets` into `ui.css` for short.
- ❌ I noticed a few differences before/after. I'm going to fix them and then the PR will be ready for a complete review.

ragna/deploy/_ui/central_view.py
pierrotsmnrd: yes, that's worth reporting to Panel.
if we define the input widget of the chat interface like this : 
```python
# ...
widgets=[
                pn.widgets.TextInput(
                    placeholder="Ask a question about the documents",
                    css_classes=["my_custom_class"]
                )
            ],
# ...
```
The resulting input widget has its css classes forced to `chat-interface-input-widget` instead of adding it.
it's not a problem in our case, but could be for other people trying to style the ChatInterface.

ragna/deploy/_ui/app.py
```diff
@@ -23,32 +23,45 @@
...
+        # Apply the design modifiers to the panel components
+        # It returns all the CSS files of the modifiers
+        self.css_filepaths = ui.apply_design_modifiers()
         self.url = url
         self.api_url = api_url
         self.origins = origins
 
     def get_template(self):
+        # A bit hacky, but works.
+        # we need to preload the css files to avoid a flash of unstyled content, especially when switching between chats.
```
pmeier, line 39: Should we report this to panel as well? Seems like a useful thing to have in general.

ragna/deploy/_ui/central_view.py
```diff
@@ -109,41 +59,22 @@ def __init__(
...
-                        "border-width": "1.2px",
-                        "border-radius": "5px",
-                    },
-                )
-            ),
-        )
+        if role == "user":
+            object.css_classes.append("message-content-no-border")
+        else:
+            object.css_classes.append("message-content-border")
```
pmeier, line 77:
```suggestion
        object.css_classes.append(
            "message-content-no-border"
            if role == "user"
            "message-content-border"
        )
```

ragna/deploy/_ui/central_view.py
```diff
@@ -158,27 +89,24 @@ def __init__(
             show_copy_icon=False,
             css_classes=[f"message-{role}"],
         )
-        self._stylesheets.extend(message_stylesheets)
+        self._stylesheets.append("css/chat_interface/chatmessage.css")
 
     def _copy_and_source_view_buttons(self) -> pn.Row:
+        source_info_button = pn.widgets.Button(
```
pmeier, line 95: Any reason why we need to define before we create the row? If not, can we move it back into the `pn.Row`?

ragna/deploy/_ui/left_sidebar.py
```diff
@@ -79,98 +62,30 @@ def __panel__(self):
         self.chat_buttons = []
         for chat in self.chats:
             button = pn.widgets.Button(
-                name=chat["metadata"]["name"], button_style="outline"
+                name=chat["metadata"]["name"],
+                button_style="outline",
```
pmeier, line 66: Shouldn't that also be part of the CSS rather than mixing panel functionality with our custom CSS?

ragna/deploy/_ui/main_page.py
```diff
@@ -141,6 +141,8 @@ def __panel__(self):
                             """.replace(
                     "{new_chat_btn_name}", new_chat_button_name
                 ).strip(),
+                # This is not really styling per say, it's just a way to hide from the page the HTML item of this hack.
```
pmeier, line 144: But it is worth to use the CSS helper :upside_down_face:

ragna/deploy/_ui/modal_configuration.py
```diff
@@ -224,19 +221,19 @@ def advanced_config_ui(self):
                         self.config.param.chunk_size,
                         name="Chunk Size",
                         bar_color=ui.MAIN_COLOR,
-                        stylesheets=[ui.SS_LABEL_STYLE],
+                        css_classes=["modal_configuration_int_slider"]
+                        + (["disabled"] if disabled_source_storage else []),
```
pmeier, line 225: Nit: I really dislike adding lists. Feel free to ignore.

```suggestion
                        css_classes=["modal_configuration_int_slider" ,*["disabled"] if disabled_source_storage else []],
```

ragna/deploy/_ui/modal_configuration.py
```diff
@@ -224,19 +221,19 @@ def advanced_config_ui(self):
...
                         width_policy="max",
                         disabled=disabled_source_storage,
-                        css_classes=["disabled"] if disabled_source_storage else [],
                     ),
                     pn.widgets.IntSlider.from_param(
                         self.config.param.chunk_overlap,
                         name="Chunk Overlap",
                         bar_color=ui.MAIN_COLOR,
-                        stylesheets=[ui.SS_LABEL_STYLE],
+                        css_classes=["modal_configuration_int_slider"]
```
pmeier, line 233: Same as above. Since it has two more occurrences below, would it make sense to define a 

```python
maybe_disabled = ["disabled"] if disabled_source_storage else []
```

on top and just use it here?

ragna/deploy/_ui/modal_configuration.py
```diff
@@ -251,18 +248,18 @@ def advanced_config_ui(self):
                     pn.widgets.IntSlider.from_param(
                         self.config.param.max_context_tokens,
                         bar_color=ui.MAIN_COLOR,
-                        stylesheets=[ui.SS_LABEL_STYLE],
+                        css_classes=["modal_configuration_int_slider"]
```
pmeier, line 251: Same as above.

ragna/deploy/_ui/modal_configuration.py
```diff
@@ -251,18 +248,18 @@ def advanced_config_ui(self):
...
+                        + (["disabled"] if disabled_source_storage else []),
                         width_policy="max",
                         disabled=disabled_source_storage,
-                        css_classes=["disabled"] if disabled_source_storage else [],
                     ),
                     pn.widgets.IntSlider.from_param(
                         self.config.param.max_new_tokens,
                         bar_color=ui.MAIN_COLOR,
-                        stylesheets=[ui.SS_LABEL_STYLE],
+                        css_classes=["modal_configuration_int_slider"]
```
pmeier, line 259: Same as above.

ragna/deploy/_ui/styles.py
```diff
@@ -1,155 +1,86 @@
...
-                        div.bk-panel-models-markup-HTML.markdown {
-                            margin-left: 15px;
-                            margin-right: 15px;
-                            margin-top:0px;
-                        }
+def apply_design_modifiers():
+    css_filepaths = []
+    for dir, classes in css_modifiers.items():
+        for cls in classes:
+            css_filename = cls.__name__.lower().split(".")[-1] + ".css"
```
pmeier, line 57: A class name can never have a period `.` in it. Should this just be

```suggestion
            css_filename = f"{cls.__name__.lower()}.css"
```

?

ragna/deploy/_ui/styles.py
```diff
@@ -1,155 +1,86 @@
...
-                        """
-        ]
-    }
+def add_modifier(
+    modifier_class: pn.viewable.Viewable,
+    modifications: Dict[str, Any],
+    property: str = "stylesheets",
+):
+    if modifier_class not in pn.theme.fast.Fast.modifiers:
+        pn.theme.fast.Fast.modifiers[modifier_class] = {}
```
pmeier, line 70: Just a little syntax sugar

```suggestion
    properties = pn.theme.fast.Fast.modifiers.setdefault(modifier_class, {})
    property_modifications = properties.setdefault(property, [])
    property_modifications.append(modifications)
```

This also supersedes the `if` / `else` below, but GH doesn't let me suggest it for both.

ragna/deploy/_ui/styles.py
```diff
@@ -159,133 +90,42 @@ def stylesheets(
...
-    border-color: darkgray;
-}
-
-
-.noUi-target[disabled='true'] .noUi-connect {
-    background-color: darkgray !important;
-}
+"""
+CSS and UI Helpers
+"""
```
pmeier, line 103: I don't think we need that "header" here

```suggestion
```

ragna/deploy/_ui/styles.py
```diff
@@ -159,133 +90,42 @@ def stylesheets(
...
 
 
-:host(.disabled) div.bk-slider-title {
-    color: darkgray !important;
-}
-
+def divider():
+    return pn.layout.Divider(styles={"padding": "0em 1em"})
```
pmeier, line 107: Should we use `stylesheets` here as well and for that the helper below? IIRC, `styles` is the old way of styling panel widgets.

ragna/deploy/_ui/styles.py
```diff
@@ -159,133 +90,42 @@ def stylesheets(
...
 
-"""
 
+def css(
```
pmeier, line 110: Please apply the refactor suggestion from https://github.com/Quansight/ragna/pull/312#pullrequestreview-1867178050.

pmeier (commented):
I've reviewed the code. I'll pull the PR later today and check if the styling is indeed the same as before. Thanks a ton for the work @pierrotsmnrd!

ragna/deploy/_ui/app.py
pierrotsmnrd: Yes. I'd say the feature request is "add a placeholder for custom HTML inside the <head>"

ragna/deploy/_ui/central_view.py
pierrotsmnrd: I changed this because that was more convenient while debugging, I needed to check properties on the button itself. I'm moving it back into `pn.Row`

ragna/deploy/_ui/styles.py
pierrotsmnrd: Better than that : I'll use a modifier.

ragna/deploy/_ui/styles.py
pierrotsmnrd: Done

pmeier (commented):
I didn't review everything yet, but I already found a few differences in the UI styling:

## Create chat

### `main`

Screenshot from 2024-02-12 15-57-16

### PR

Screenshot from 2024-02-12 15-56-02

## Message feed

### `main`

Screenshot from 2024-02-12 15-57-40

The divider is black on the PR while it is a lighter gray on `main`.

### PR

Screenshot from 2024-02-12 15-56-43

The messages on the PR are missing the border / background color.

ragna/deploy/_ui/styles.py
pierrotsmnrd: It clearly separates pure CSS constants from non constants. But feel free to replace/delete it.

ragna/deploy/_ui/modal_configuration.py
```diff
@@ -251,18 +251,17 @@ def advanced_config_ui(self):
...
+                        css_classes=source_storage_css_classes,
                         width_policy="max",
                         disabled=disabled_source_storage,
-                        css_classes=["disabled"] if disabled_source_storage else [],
                     ),
                     pn.widgets.IntSlider.from_param(
                         self.config.param.max_new_tokens,
                         bar_color=ui.MAIN_COLOR,
-                        stylesheets=[ui.SS_LABEL_STYLE],
+                        css_classes=["modal_configuration_int_slider"]
```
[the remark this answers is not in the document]
pierrotsmnrd: I don't understand the question

pierrotsmnrd:
There was indeed a problem with the separator in the configuration modal. 
I don't reproduce your screenshot of the chat feed though. I guess it's a matter of emptying cache. Can you try that ?

ragna/deploy/_ui/modal_configuration.py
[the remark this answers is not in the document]
pmeier: The other sliders right above this one have

```python

css_classes=source_storage_css_classes,
```

while this one has the configuration manually. Is this intentional?

ragna/deploy/_ui/central_view.py
pmeier: See #379.

pmeier:
@pierrotsmnrd panel==1.4.0 (tracked in #379) also introduced new ways of styling the chat messages: https://blog.holoviz.org/posts/panel_release_1.4/#other-enhancements Can this be adopted here to simplify stuff?

ragna/deploy/_ui/modal_configuration.py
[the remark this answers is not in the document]
pmeier: Should we also add a `assistant_css_classes` to make it more obvious?

ragna/deploy/_ui/app.py
pmeier: Did we do that yet?

pmeier:
image

"Source info" button looks different from the copy button. This has nothing to do with `panel==1.4` as it is visible in https://github.com/Quansight/ragna/pull/312#pullrequestreview-1875418845 as well.

pierrotsmnrd:
Capture d’écran 2024-05-13 à 12 40 10

This is what I see. Can you make sure you empty cache before refresh ?

ragna/deploy/_ui/central_view.py
pmeier: Let's do it in a follow-up PR @pierrotsmnrd

ragna/deploy/_ui/app.py
pmeier: @pierrotsmnrd will do it afterwards.

pmeier (approved):
Thanks a ton Pierre! This looks awesome!

PR #316, opened 2024-02-12
re-add Authentication documentation

### LINKED ISSUE #279: Documentation for authentification objects is missing
#205 moved the `Authentication` object from `ragna.core` to `ragna.deploy`. However, the latter is not part of our API documentation:

https://github.com/Quansight/ragna/blob/38f75f6d88e74f72c030ef50045315b81eb203aa/docs/references/python-api.md?plain=1#L1-L7

In addition, `ragna.deploy` is also missing the block in the `__init__.py` file that hides our internal structure from the documentation. We can just copy the one from `ragna.core`

https://github.com/Quansight/ragna/blob/38f75f6d88e74f72c030ef50045315b81eb203aa/ragna/core/__init__.py#L57-L62

pmeier:
Fixes #279.

### DIFF
--- a/docs/references/python-api.md
+++ b/docs/references/python-api.md
@@ -5,3 +5,5 @@
 ::: ragna.source_storages
 
 ::: ragna.assistants
+
+::: ragna.deploy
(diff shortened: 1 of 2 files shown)


pavithraes (approved):
LGTM, thank you :)

PR #317, opened 2024-02-12
only send sources on first chunk when streaming

pmeier:
I've set out to do this from the start in https://github.com/Quansight/ragna/pull/215#issuecomment-1860626030, but I forgot about it.


nenb (approved):
LGTM! This makes a lot of sense. I'll have a little time over the next few days, so I'll also push a small test to prevent regressions in this in the future. (Pushing a test myself will help me to understand any potential edge cases here, which I may have missed from eyeballing the code in the review!)

pmeier:
> I'll have a little time over the next few days, so I'll also push a small test to prevent regressions in this in the future. (Pushing a test myself will help me to understand any potential edge cases here, which I may have missed from eyeballing the code in the review!)

Thanks a ton :pray: I know I have been way to lax on adding tests. I'll get on that in the next couple of weeks.

However, let's add the tests in a follow-up PR. I need this one for #89.

PR #319, opened 2024-02-14
restrict accepted files in file uploader

### LINKED ISSUE #165: Limit allowed files in file selector
We are currently setting the text of the upload area to include the supported documents

https://github.com/Quansight/ragna/blob/3d8541f15ca187b6ef01d49e40db8f33fc9d7ec4/ragna/_ui/modal_configuration.py#L163-L169

image

However, the file selection window that pops up allows all files. Meaning, if users don't read the text properly (and they won't; I don't this is why I'm reporting this), they can select documents that will refused directly afterwards.

Since we have the supported suffixes available, we should limit the "allowed documents" in the file selector as well.

### LINKED ISSUE #271: Don't reformat allowed document types in upload area
### Feature description

In the file selection area of the UI, we are currently dropping the leading period and uppercasing the extension of the allowed documents:

https://github.com/Quansight/ragna/blob/20ecade8992ecf458575a668a944b229e45e6f11/ragna/deploy/_ui/modal_configuration.py#L167-L169

Meaning, we are turning `.pdf` into `PDF`. This is in line with our [Figma designs](https://www.figma.com/file/9J5Vax4ENZCPQQMl8WamBq/Ragna?type=design&node-id=550-1345&mode=design&t=pCq14XC2jL09lG71-0).

Looking at this screenshot taken from #270

image

I think TXT and PDF are clear enough, but I "stumbled" over `MD` as I have never seen that before. Had it been `.md`, I think it would be instantly clear for someone familiar with Markdown. Even if not, they could potentially perform the match when looking at the extension of the file in the file picker.

I want us to look at the design again and see if this was intentional or not. If no, I would like us to just use the plain extension including the leading period.

cc @smeragoel 

### Value and/or benefit

Less confusion about the possible file formats for the user. This becomes even more relevant if we add more formats in the future, which is planned.

### Anything else?

_No response_

pmeier:
Fixes #165, fixes #271.


ragna/deploy/_ui/components/file_uploader.py
```diff
@@ -151,14 +144,16 @@ def perform_upload(self, event=None, after_upload_callback=None):
...
+                        Allowed files: ${allowed_documents_str}
                     </div>
                     <input  type="file" 
                             name="fileUpload"
                             class="fileUpload" 
                             id="fileUpload" 
                             multiple="multiple" 
-                            onchange="${script('file_input_on_change')}" /> 
+                            onchange="${script('file_input_on_change')}" 
+                            accept="${allowed_documents_str}"
```
pmeier, line 155: This is not working for some reason. When looking at it with the browser console, I see

image

`accept=""` is empty. Curiously, using the exact same syntax above to display the allowed files works just fine

image

Not sure what is going on here.

pmeier: Fix in d01ec2da2062cdb03cd18543c0bbcabe8a405248 by @pierrotsmnrd.

pierrotsmnrd (approved):
Looks good to me

PR #320, opened 2024-02-14
remove memory option for state db

### LINKED ISSUE #213: Remove the memory option for database URL
After #205, the config is no longer in use for the Python API. Thus, we can just use a local sqlite instance as default to have persistence. Users can still select an ephemeral in-memory database by using `sqlite://` as option.

Here are some places where changes need to happen

- Config wizard
  https://github.com/Quansight/ragna/blob/2c1e5c407c4e8afd7f2976021bd5f6551ff57af0/ragna/deploy/_cli/config.py#L254-L265
- REST API
  https://github.com/Quansight/ragna/blob/2c1e5c407c4e8afd7f2976021bd5f6551ff57af0/ragna/deploy/_api/core.py#L105-L108
- REST API tests
  https://github.com/Quansight/ragna/blob/2c1e5c407c4e8afd7f2976021bd5f6551ff57af0/tests/deploy/api/test_e2e.py#L12-L15

pmeier:
Fixes #213.


ragna/deploy/_config.py
```diff
@@ -62,7 +63,8 @@ class ApiConfig(ConfigBase):
     url: str = "http://127.0.0.1:31476"
     # FIXME: this needs to be dynamic for the UI url
     origins: list[str] = ["http://127.0.0.1:31477"]
-    database_url: str = "memory"
+    # FIXME: this needs to be dynamic from the local cache root
```
nenb, line 66: What does this FIXME mean?

ragna/deploy/_cli/config.py
```diff
@@ -252,19 +252,6 @@ def _wizard_common() -> Config:
         qmark=QMARK,
     ).unsafe_ask()
 
-    if questionary.confirm(
```
nenb, line 255: Does this mean that it is no longer possible to configure the db location from the wizard? If so, what is the rationale for this?

nenb (approved):
LGTM! I had a clarification question over the wizard that you might want to double-check.

ragna/deploy/_config.py
pmeier: When you create a config in python and set only the `local_cache_root`, e.g.

```

config = Config(local_cache_root="/my/custom/path")
```

the default database URL will still be in `~/.cache/ragna`. Optimally, the default database URL should be `sqlite:////{local_cache_root}/ragna.db` rather than hardcoding the directory. I'll open an issue about that.

pmeier: See #326.

PR #322, opened 2024-02-14
use FastAPI test client for testing

pmeier:
Per title. Documentation is [here](https://fastapi.tiangolo.com/tutorial/testing/). This has at least three upsides:

1. We don't need to maintain code to actually starting and shutting down the API.
2. Since the test client doesn't actually start the API, this is quite a bit faster.
3. This allows the API code to be tracked in coverage without any shenanigans regarding coverage in subprocesses.


tests/deploy/api/test_e2e.py
```diff
@@ -1,47 +1,40 @@
...
+    with TestClient(app(config)) as client:
         username = default_user()
         token = (
             client.post(
                 "/token",
                 data={
                     "username": username,
                     "password": os.environ.get(
-                        "AI_PROXY_DEMO_AUTHENTICATION_PASSWORD", username
+                        "RAGNA_DEMO_AUTHENTICATION_PASSWORD", username
```
pmeier, line 37: This was a driveby since I was looking at this function. After that I `grep`ed for the wrong string and found one more usage that I also fixed.

tests/deploy/api/test_e2e.py
```diff
@@ -1,47 +1,40 @@
...
+from fastapi.testclient import TestClient
 
-from ragna._utils import timeout_after
 from ragna.core._utils import default_user
 from ragna.deploy import Config
-from tests.utils import ragna_api
+from ragna.deploy._api import app
 
 
-@pytest.mark.parametrize("database", ["memory", "sqlite"])
```
pmeier, line 14: Removing the memory test here for two reasons:

1. #320 will remove it from our special cases. Meaning, this is no longer part of our core API, but rather a special case.
2. For some reason, in-memory databases don't play nice with the `TestClient` (tiangolo/fastapi#3906). The proposed solution, i.e. using `poolclass=StaticPool` does indeed work, but will cause the streaming test to crash with encode/starlette#1315. I figured it wasn't worth the effort to dive any deeper than that.

nenb (approved):
LGTM! Makes sense.

PR #323, opened 2024-02-15
increase prettier tab width to 4 for proper documentation rendering

pmeier:
This PR looks massive, but it is mostly automatic whitespace changes. The only manual change is in `.prettierrc`. I'll flag it below.

PR #324, opened 2024-02-15
use truly-sane-lists plugin

pmeier:
Alternative to #323. We add a dev dependency, but I think that is worth it to avoid a lot of extra whitespace by increasing the tab width.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -4,6 +4,14 @@ title: Home
 
 # ![Ragna logo](assets/brand/logo-lockup-horizontal/logo-lockup-horizontal.png)
 
+- Here is a nested list with a tab width of 2
+  - First item
+  - Second item
+- Here is a nested list with a tab width of 4
+    - First item
+    - Second item
+
+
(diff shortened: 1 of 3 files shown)


docs/index.md
```diff
@@ -4,6 +4,14 @@ title: Home
 
 # ![Ragna logo](assets/brand/logo-lockup-horizontal/logo-lockup-horizontal.png)
 
+- Here is a nested list with a tab width of 2
+  - First item
+  - Second item
+- Here is a nested list with a tab width of 4
+    - First item
+    - Second item
```
pmeier, line 12: Lint CI is also triggering now, but this format with the extension gives this result

image


Since we don't want to use a tab width of 4, this is fine. Of course we need to revert 06a71a8ed7cb24135814b42b5aee5fcba56c56af before merge.

pavithraes (approved):
@pmeier - Thanks, I prefer this one too. The PR looks good, I'll approve so we can revert the testing commit and merge. :)

PR #325, opened 2024-02-15
Remove duplicate call to answer method when not streaming

nenb:
`await core_chat.answer(...)` is also called again inside the non-streaming block of code.


pmeier (approved):
Thanks!

PR #328, opened 2024-02-19
Config refactor

### LINKED ISSUE #268: make deploying Ragna easier
## Status quo

When deploying Ragna, we have the following configuration parameters:

https://github.com/Quansight/ragna/blob/20ecade8992ecf458575a668a944b229e45e6f11/ragna/deploy/_config.py#L55-L61

https://github.com/Quansight/ragna/blob/20ecade8992ecf458575a668a944b229e45e6f11/ragna/deploy/_config.py#L65-L70

Both the API and the UI have an `url` and an `origins` parameter.

### `url`

This parameter actually contains three bits of information: scheme, host, and port.

#### API

https://github.com/Quansight/ragna/blob/20ecade8992ecf458575a668a944b229e45e6f11/ragna/deploy/_cli/core.py#L83-L91

- We ignore the scheme, e.g. `http://`, since `uvicorn` is an HTTP server and cannot support any other protocol anyway (not that we want that)
- We bind the server to the given `host` and `port`
- If the URL has no `port`, we default to `31476`

#### UI

https://github.com/Quansight/ragna/blob/20ecade8992ecf458575a668a944b229e45e6f11/ragna/deploy/_ui/app.py#L117-L130

- We ignore the scheme, since `panel` is an HTTP server
- We ignore the host. This is most likely and oversight. We should pass it as `address`
- We bind the server to the `localhost:{port}`

### `origins`

This parameter is a list of other URLs that are allowed to make CORS requests to the deployment. It is clear that a browser, i.e. the client that makes the CORS requests, needs to do so for the UI. For the API it is not as clear on first sight. Due some limitations of the `tornado` server (underlying technology of `bokeh`, which itself drives `panel`), we needed to write our custom upload logic in JS. And since that is running in the browser, uploading documents through the UI results in CORS requests to the API.

Since CORS is fairly strict by default, `localhost` and `127.0.0.1` are not interchangeable there. This is unexpected for people unfamiliar with the topic and thus we have a small helper function [`handle_localhost_origins`](https://github.com/Quansight/ragna/blob/20ecade8992ecf458575a668a944b229e45e6f11/ragna/_utils.py#L56-L73) which does nothing but add the respective other if only one of `localhost` and `127.0.0.1` are added to the origins.

#### API

https://github.com/Quansight/ragna/blob/20ecade8992ecf458575a668a944b229e45e6f11/ragna/deploy/_api/core.py#L40-L46

#### UI

https://github.com/Quansight/ragna/blob/20ecade8992ecf458575a668a944b229e45e6f11/ragna/deploy/_ui/app.py#L137

### What is the problem?

The UI is built on top of the API. For this, the UI needs to be aware of the API URL. Superficially, this is not an issue, since we have access to the config which carries this information. However, this logic is flawed: `config.api.url` is where the API server is bound to, while the UI needs to know the URL where it can contact. When doing everything locally, the world is still fine because the API is bound to `localhost:31476` and also contacted under the same address.

Here is a scenario where this breaks down:

For PyData NYC and Global we deployed Ragna on AWS. Both the API and UI where running from separate docker containers (which used the same image). After a lot of trial and error we realized that we are hitting the issue I described above. Our solution was to set `config.api.url` on the API container to `http://0.0.0.0:31476` and on the UI container to `https://api.ragna.quansight.dev`. Since we had two separate containers we were able to abuse the same variable to create the split of binding and contact address. Even for us as core maintainers this was unexpected and non-obvious. Meaning, just because the problem has a workaround for this specific use case, we should strife to make this easier.

While working on #267, we hitting CORS issues again and my hunch says this is exactly the same issue. I'll update here with my findings.

### LINKED ISSUE #326: Use local_cache_root in defaults for other config options
In #320 we are removing the `memory` option for the state database. By default, we'll then have a SQLite database. This should default to `sqlite:///{config.local_cache_root}/ragna.db`, but is hardcoded to `sqlite:///~/.cache/ragna/ragna.db` for now.

Basically, we want the behavior that we already have for the config wizard

https://github.com/Quansight/ragna/blob/2591c763e124a3c31bcdc3e789843da32c538225/ragna/deploy/_cli/config.py#L260-L264

This should be doable with `pydantic`.

### LINKED ISSUE #327: make origins in config dynamic
Sister issue to #326 and can probably be fixed in the same PR. Right now we already have a FIXME for this

https://github.com/Quansight/ragna/blob/2591c763e124a3c31bcdc3e789843da32c538225/ragna/deploy/_config.py#L63-L64

https://github.com/Quansight/ragna/blob/2591c763e124a3c31bcdc3e789843da32c538225/ragna/deploy/_config.py#L73-L74

The idea is if someone specifies a different URL than the default, the origins should be set automatically. Meaning, if I set `config Config(ui={"url": "http://127.0.0.1:12345"})`, we should have `config.api.origins == config.ui.origins == ["http://127.0.0.1:12345"]`.

pmeier:
Closes #268, closes #326, and closes #327.

This PR groups a few changes:

- Replace the `config.api.url` and `config.ui.url` settings with a `hostname` and `port` pair. See #268 for the rationale here. `config.api.url` is retained to now indicate the URL the API is served and can be contacted from the UI.
- Rename the `config.local_cache_root` to `config.local_root` to align with our Python API
  https://github.com/Quansight/ragna/blob/b3cc95ab5b4228b03af9937afc9dc842ccf25432/ragna/_utils.py#L15
- Remove the `components` section and merge its contents into the top-level document. Originally, the components section was called `rag`, but this was changed when removed the config from the RAG Python API in #205. I initially thought it makes sense to have the RAG components in a separate section, but I'm walking back on that here. We can have the source storages and assistants next to the document and authentication. Note that this is not an additional BC break since #205 is not yet released either.
- Support dynamic default values that depend on the rest of the config being validated first:
  - `config.api.database_url` depends on `config.local_root`
  - `config.api.origins` and `config.ui.origins` depend on `config.ui.hostname` and `config.ui.port`

I'm aware these are quite a few breaking changes. But I figured if we have a BC break anyway (and we have that already for `0.2.0`), let's fix a few other issues along the way. Better to break BC once than breaking it multiple times (this is **not** a promise it won't happen again before the 1.0 release).


tests/core/test_config.py
pmeier, line 1: The file was not deleted, but rather moved from `tests/core/test_config.py` to `tests/deploy/test_config.py`. GH doesn't recognize this, because this PR added a bunch of tests in this module and thus the differ significantly.

ragna/__init__.py
```diff
@@ -14,3 +14,12 @@
 
 from . import assistants, core, deploy, source_storages
 from .core import Rag
+
```
nenb, line 17: nitpick: why not include `__version__` here?

nenb (approved):
LGTM!

There are a bunch of changes here that could break things in unexpected ways (as you mentioned in the PR description), so it would be good to have others test as well.

As a comment, `deploy/_config.py` now has a high degree of complexity. This makes it powerful, but it has definitely made it more challenging (personal experience) for debugging. I would personally not want to add any more complexity here.

ragna/__init__.py
pmeier: Oversight. Good catch!

pmeier:
> As a comment, `deploy/_config.py` now has a high degree of complexity. This makes it powerful, but it has definitely made it more challenging (personal experience) for debugging. I would personally not want to add any more complexity here.

In general I agree with that. However, we currently don't know what other stuff we might want to do in the future. If we find a simple solution, I'm all for it. But ultimately, the complexity of the code is only a secondary goal if we decided to implement a feature. Of course the complexity should be taken into account when making the decision. But if that is done, we can only try our best to keep the complexity down.

Issue #337, opened 2024-03-03
Switching between chats is slow


nenb:
### Bug description

If I switch between chats multiple times, the UI performance becomes slower and slower. This effect is more extreme depending on the number of messages in a chat. For chats of over 20 messages, toggling between chats can make the UI unusable quite quickly on lower-end machines.

### How to reproduce the bug?

Create two chats of twenty+ messages. Toggle between the two chats a few times. It should take longer and longer to load the chat interface each time. The memory consumed should also be seen to grow steadily.

### Versions and dependencies used

`main` branch.

### Anything else?

The culprit is the `_stylesheets` [class variable](https://github.com/Quansight/ragna/blob/c1c159d196db99cb0372cddd4abf8ecee3ce255d/ragna/deploy/_ui/central_view.py#L161) in `RagnaChatMessage`. 

For each message created, this class variable grows in size. Multiple messages are (re-)created when toggling between chats, causing this class variable to quickly explode in size. This is causing `panel` to do a bunch of work under the hood, which quickly degrades the UI performance.

This is likely best dealt with in #312, to make sure that similar effects don't creep in elsewhere.

(Also happy to open a small PR resolving this if this is easier.)

pmeier:
This was first reported in #239 although without any indication of what is wrong. Thanks for debugging @nenb!

> This is likely best dealt with in #312, to make sure that similar effects don't creep in elsewhere.
> 
> (Also happy to open a small PR resolving this if this is easier.)

#312 hinges on `panel==1.4.0`, which is not released yet (https://github.com/Quansight/ragna/pull/312#discussion_r1481492621). Happy to review a workaround in the meantime.

nenb:
> Happy to review a workaround in the meantime.

I've opened a quick PR in #341. This **does not** prevent future regressions, as I didn't have time to think of a way to do this. This is potentially relevant to #312.

PR #341, opened 2024-03-04
Prevent creation of large _stylesheet class variable

nenb:
This PR fixes a bug (#337) whereby large stylesheets objects were passed to `panel`/`bokeh` to parse and render. The issue was that every time a new message was created in a chat, the same bunch of CSS rules were added to a class variable (`_stylesheets`), causing rapid growth in the size of the variable.


pmeier (commented):
To asses the situation correctly:

- `ChatMessage._stylesheets` is a class variable coming from `panel` so there is little we can do about it
- It is filled with some default values and we just want to append ours
- Our style sheets are static and thus could also be appended statically

Is that correct? If so, can't we just do

```python
class RagnaChatMessage(pn.chat.ChatMessage):
    _stylesheets = [
        *pn.chat.ChatMessage._stylesheets,
        *message_stylesheets,
    ]
```

? No need to ever touch them at runtime.

nenb:
> Is that correct? If so, can't we just do
```
class RagnaChatMessage(pn.chat.ChatMessage):
    _stylesheets = [
        *pn.chat.ChatMessage._stylesheets,
        *message_stylesheets,
    ]
```
> ? No need to ever touch them at runtime.

Makes sense to me (although I haven't tested it). Feel free to push this to my branch and use it instead.

pmeier (approved):
Added and tested my suggestion from https://github.com/Quansight/ragna/pull/341#pullrequestreview-1915214284.

PR #309, opened 2024-02-02
move BC policy into documentation

pmeier:
Closes #266.

PR #343, opened 2024-03-04
fix doc links in README

pmeier:
Follow-up to #89 and #309. These links are broken as of now, since they point to the stable documentation, which will only include the resources after the 0.2.0 release.


pavithraes (approved):
Thanks, looks good. I'll approve, but as you noted, we can merge this after the release.

pmeier:
No, we'll have to merge before, because the README is part of the release. Otherwise the links will not work on PyPI. I'll merge this right before cutting the release.

PR #345, opened 2024-03-06
add option to not open the browser when starting the web UI

pmeier:
@aktech With this you can pass `--no-open-browser`.


aktech (approved):
Looks good, verified it works.

PR #346, opened 2024-03-06
use root_path in api.url

pmeier:
Supersedes #305. After an offline discussion with @aktech we agreed that it preferable to include the `config.api.root_path` in `config.api.url`. While this means we have this we have this information duplicated in the default configuration, this has two upsides:

1. We don't need to join the URL and the root path everywhere in our code base.
2. It opens up the possibility to route an app deployed with a root path to a top level domain without one.


aktech (approved):
Looks good to me, verified it works.

PR #349, opened 2024-03-11
add test for a streaming assistant

pmeier:
Previously, we indeed tested `stream=True`, but we used the demo assistant to do it which does not support streaming. Meaning, we only ever got a single chunk from the API.

Test for this was proposed in https://github.com/Quansight/ragna/pull/317#pullrequestreview-1886007856.


nenb (approved):
Nice work, can see this being a very useful safety feature in the future 🚀 . 

Sorry I didn't follow through on this.

PR #350, opened 2024-03-11
improve error handling for builtin assistants

pmeier:
I spend almost an hour today trying to find a bug that wasn't there, because we didn't raise a proper error when our API call failed.

### DIFF
--- a/tests/assistants/test_api.py
+++ b/tests/assistants/test_api.py
@@ -6,6 +6,7 @@ from ragna import assistants
 from ragna._compat import anext
 from ragna.assistants._api import ApiAssistant
 from ragna.core import RagnaException
+from tests.utils import skip_on_windows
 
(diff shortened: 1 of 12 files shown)


tests/assistants/test_api.py
```diff
@@ -0,0 +1,28 @@
...
+API_ASSISTANTS = [
+    assistant
+    for assistant in assistants.__dict__.values()
+    if isinstance(assistant, type)
+    and issubclass(assistant, ApiAssistant)
+    and assistant is not ApiAssistant
+]
+
+
+@skip_on_windows
```
pmeier, line 20: I tried multiple things now, but got nothing to work. All data points to the test code being the problem rather than the code that is tested here. Let's ignore this for now. Maybe someone else has more insights later.

nenb (approved):

Issue #256, opened 2023-12-19
"managed" Ragna


pmeier:
It's been roughly two months since the initial release of Ragna. While the overall feedback is positive, we had feature requests (I'll link stuff below) for a "managed" version of Ragna. By "managed" I don't mean managed by the Ragna team as a paid service, but rather managed by an admin team inside an organization for regular users inside the same organization. If someone has a better name for this, please come forward.

Let's have a look at how Ragna currently works and why this is the case, what exactly is requested, and what ultimately needs to change to make these requests a reality.

## Status quo

Ragnas original goal was to make the research case, i.e. evaluating all different RAG components, easy. We made the following assumptions:

- The user not only has access to the documents they want to test, but also has them on disk.
- The user potentially wants to (and in turn is allowed to) tweak everything.

With that we made the chat the "atomic unit" of Ragna. Each chat can be about different documents and all components can be configured freely.

## What is missing?

As explained above, the use case that we build Ragna for requires a large degree of agency of the user. In an actual production setting, the assumptions we made above might not hold:

- The user might not have the documents on disk, but they are stored in some remote "database" (potentially an actual database, a wiki, ...)
- The user might not have any clue what all the parameters mean and they "just want to get my questions answered"

Thus, Ragna currently falls flat for this use case or at least makes it difficult. This is evident by bunch of feature requests and comments we received. I'll try to link everything here. If I missed something, feel free to follow up on this comment.

- Instead of uploading documents for each chat, there is a (potentially large) corpus of documents that either exists in a remote storage (https://github.com/Quansight/ragna/discussions/202#discussioncomment-7622056, https://github.com/Quansight/ragna/pull/251, https://github.com/Quansight/ragna/discussions/245#discussion-5949885) or will be uploaded once by an administrator (https://github.com/Quansight/ragna/issues/176#issuecomment-1804119516). The uploaded documents are than chunked, embedded and stored in the source storage.
- When starting a chat, the user can optionally select a subset of the documents, e.g. by selecting tags attached by the administrator (https://github.com/Quansight/ragna/issues/246). If nothing is selected, they use the full corpus. Afterwards they potentially select an assistant and its parameters and can start without any time-consuming chunking and embedding.

Basically, this uncouples the `Chat.prepare` and `Chat.answer` methods. The former will be executed by the administrator and the latter by the users.

Cc @peachkeel, @nenb, @NetDevAutomate who contributed the requests linked above.

## What needs to change?

[...]

Issue #217, opened 2023-11-27
Allow arbitrary component parameters in the web UI


pmeier:
### Feature description

One of the core features of Ragna is its extension mechanism that allows users (or third-party projects) to integrate well with it. The flexibility happens on two levels:

1. The user can select which components to use.
2. Each component can declare its parameters that can be set through a unified interface.

Point 1. is fully supported by the Python and REST API as well as the web UI. Point 2. however is only properly supported by the Python and REST API. The web UI currently hardcodes the available parameters:

https://github.com/Quansight/ragna/blob/2c1e5c407c4e8afd7f2976021bd5f6551ff57af0/ragna/deploy/_ui/modal_configuration.py#L214-L280

This relies on the fact that almost all builtin source storages and assistants rely on the same set of parameters. However, we already got bitten by it ourselves, see #155. This will not become any easier in the future when we actually go for #191.

My proposal here is to add functionality to the

1. Python API,
2. REST API, and
3. web UI,

to be able to support any custom component without hardcoding it.

Fortunately, we already have most of the things done for 1. and 2.

1. We check the `**kwargs` of the `ragna.core.Chat` construction with `pydantic`. We are able to do so by dynamically building a model for each function that is part of the Ragna protocol and merging them together for all components that are used for a chat. Right now we only do this on the type. However, we could optionally allow additional metadata the same way that FastAPI and typer do: instead just declaring `foo: int` we could do `foo: typing.Annotated[int, <metadata>]`. The `<metadata>` could just be a `pydantic.FieldInfo` that can have some additional information like `gt` (greater than), `title`, and `description`.
2. We are using the information from 1. in the `/components` endpoint. We return the name of the component as well as the JSON schema from the model we build in 1. Meaning, if we implement 1. as suggested above, the new metadata is automatically included and we don't need to change anything else.

For example:

```python
import pydantic
import pprint


class AssistantAnswer(pydantic.BaseModel):
    max_new_tokens: int = pydantic.Field(
        default=256,
        gt=0,
        title="Maximum new tokens",
        description=(
            "Maximum new tokens to generate. "
            "If you experience truncated answers, increase this value. "
            "However, be aware that longer answers also incur a higher cost."
        ),
    )
```

```
{'properties': {'max_new_tokens': {'default': 256,
                                   'description': 'Maximum new tokens to '
                                                  'generate. If you experience '
                                                  'truncated answers, increase '
                                                  'this value. However, be '
                                                  'aware that longer answers '
[...]

PR #354, opened 2024-03-12
Embeddings Model and Chunking Engine (Preliminary PR for evaluation purposes only)

Tengal-Teemo:
This attempts to add an `EmbeddingModel` and `ChunkingModel` with the same overall architecture (using components) as the rest of ragna allowing for, hopefully, automated generation of UI components.  

This PR is for preliminary evaluation only.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,7 @@ __pycache__/
 
 # Ignore my entrypoint
 test.py
+doctest/
 
 # C extensions
 *.so
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,9 @@ __pycache__/
 *.py[cod]
 *$py.class
 
+# Ignore my entrypoint
+test.py
+
 # C extensions
 *.so
 
--- a/ragna/core/_rag.py
+++ b/ragna/core/_rag.py
@@ -48,7 +48,7 @@ class Rag(Generic[C]):
         self._components: dict[Type[C], C] = {}
 
     def _load_component(
-        self, component: Union[Type[C], C], *, ignore_unavailable: bool = False, **kwargs
+        self, component: Union[Type[C], C], *, ignore_unavailable: bool = False
     ) -> Optional[C]:
         cls: Type[C]
         instance: Optional[C]
@@ -71,7 +71,7 @@ class Rag(Generic[C]):
                     raise RagnaException(
                         "Component not available", name=cls.display_name()
                     )
-                instance = cls(**kwargs)
+                instance = cls()
 
             self._components[cls] = instance
 
@@ -168,7 +168,7 @@ class Chat:
 
         self.documents = self._parse_documents(documents)
         self.source_storage = cast(
-            SourceStorage, self._rag._load_component(source_storage, embedding_dimensions=self.embedding_model.get_embedding_dimensions())
+            SourceStorage, self._rag._load_component(source_storage)
         )
         self.assistant = cast(Assistant, self._rag._load_component(assistant))
 
--- a/ragna/core/_rag.py
+++ b/ragna/core/_rag.py
@@ -25,6 +25,9 @@ from ._components import Assistant, Component, Message, MessageRole, SourceStora
 from ._document import Document, LocalDocument
 from ._utils import RagnaException, default_user, merge_models
 
+from ragna.source_storages._embedding_model import GenericEmbeddingModel
+from ragna.source_storages._chunking_model import GenericChunkingModel
+
 T = TypeVar("T")
 C = TypeVar("C", bound=Component)
 
@@ -45,7 +48,7 @@ class Rag(Generic[C]):
         self._components: dict[Type[C], C] = {}
 
     def _load_component(
-        self, component: Union[Type[C], C], *, ignore_unavailable: bool = False
+        self, component: Union[Type[C], C], *, ignore_unavailable: bool = False, **kwargs
     ) -> Optional[C]:
         cls: Type[C]
         instance: Optional[C]
@@ -68,7 +71,7 @@ class Rag(Generic[C]):
                     raise RagnaException(
                         "Component not available", name=cls.display_name()
                     )
-                instance = cls()
+                instance = cls(**kwargs)
 
             self._components[cls] = instance
 
@@ -80,6 +83,8 @@ class Rag(Generic[C]):
         documents: Iterable[Any],
         source_storage: Union[Type[SourceStorage], SourceStorage],
         assistant: Union[Type[Assistant], Assistant],
+        embedding_model: Union[Type[GenericEmbeddingModel], GenericEmbeddingModel],
+        chunking_model: Union[Type[GenericChunkingModel], GenericChunkingModel],
         **params: Any,
     ) -> Chat:
         """Create a new [ragna.core.Chat][].
@@ -89,6 +94,8 @@ class Rag(Generic[C]):
                 [ragna.core.LocalDocument.from_path][] is invoked on it.
             source_storage: Source storage to use.
             assistant: Assistant to use.
+            embedding_model: Embedding Model to use
+            chunking_model: Chunking Model to use
             **params: Additional parameters passed to the source storage and assistant.
         """
         return Chat(
@@ -96,6 +103,8 @@ class Rag(Generic[C]):
             documents=documents,
             source_storage=source_storage,
             assistant=assistant,
+            embedding_model=embedding_model,
+            chunking_model=chunking_model,
             **params,
         )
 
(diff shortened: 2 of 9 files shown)


Tengal-Teemo:
@pmeier See this draft pull request.

pmeier:
Thanks for the PR @Tengal-Teemo! I'll have a look soon.

>  as the rest of ragna allowing for, hopefully, automated generation of UI components.

Well, #217 :smiling_face_with_tear:

ragna/core/_rag.py
```diff
@@ -188,7 +202,14 @@ async def prepare(self) -> Message:
                 detail=RagnaException.EVENT,
             )
 
-        await self._run(self.source_storage.store, self.documents)
+        if list[Document] in inspect.signature(self.source_storage.store).parameters.values():
```
pmeier, line 205: 1. We should have this kind of logic on as class attribute on the `SourceStorage` itself. Otherwise, how are we going to communicate this to the REST API / web UI? This needs to be known, because it makes no sense to force the user to select an embedding model when it is unused by the backend.
2. This check needs to be more strict. We should only check the first argument rather than the whole signature.

ragna/core/_rag.py
```diff
@@ -188,7 +202,14 @@ async def prepare(self) -> Message:
...
+            await self._run(self.source_storage.store, self.documents)
+        else:
+            # Here we need to generate the list of embeddings
+            chunks = self.chunking_model.chunk_documents(self.documents)
+            embeddings = self.embedding_model.embed_chunks(chunks)
```
pmeier, line 210: The source storage should also be able to just request chunks. Meaning, we have three distinct cases and cannot group these two. However, if we split this PR as suggested above, this distinction will only come in the follow-up PR.

ragna/core/_rag.py
```diff
@@ -218,7 +239,10 @@ async def answer(self, prompt: str, *, stream: bool = False) -> Message:
 
         self._messages.append(Message(content=prompt, role=MessageRole.USER))
 
-        sources = await self._run(self.source_storage.retrieve, self.documents, prompt)
+        if list[Document] in inspect.signature(self.source_storage.store).parameters.values():
```
pmeier, line 242: This hits a point that I didn't consider before: we are currently passing the documents again to the retrieve function. See the part about BC in https://github.com/Quansight/ragna/issues/256#issuecomment-1954057313 for a reason why. This will likely change when we implement #256. However, in the mean time we need to decide if we want the same "input switching" here as for store. I think this is ok, but want to hear your thoughts.

ragna/source_storages/_chroma.py
```diff
@@ -25,6 +28,9 @@ def __init__(self) -> None:
 
         import chromadb
 
+        self._embedding_model = MiniLML6v2()
+        self._chunking_model = NLTKChunkingModel()
```
pmeier, line 32: Why would the source storage need any of these?

ragna/source_storages/_chroma.py
```diff
@@ -33,58 +39,59 @@ def __init__(self) -> None:
             )
         )
 
+        self._tokens = 0
+        self._embeddings = 0
```
pmeier, line 43: These will change with every `.store` call. Why are they instance attributes rather than local variables?

ragna/core/_rag.py
pmeier: Ooh, I only in hindsight understand the logic here. Of course we need to be able to embed the prompt. So this correct.

ragna/source_storages/_embedding_model.py
```diff
@@ -0,0 +1,47 @@
...
+    chunk: Chunk
+
+    def __init__(self, embedding: list[float], chunk: Chunk):
+        super().__init__()
+        self.embedding = embedding
+        self.chunk = chunk
+
+
+class GenericEmbeddingModel(Component, ABC):
+    _EMBEDDING_DIMENSIONS: int
```
pmeier, line 24: Shouldn't this be an attribute on the `Embedding` class? Or maybe we can eliminate it all together since it is just `len(Embedding().embedding)`?

ragna/source_storages/_embedding_model.py
```diff
@@ -0,0 +1,47 @@
...
+
+    @abstractmethod
+    def embed_chunks(self, chunks: list[Chunk]) -> list[Embedding]:
+        raise NotImplementedError
```
pmeier, line 28: Nit: `@abstractmethods` usually just use `...` (this is what Ragna does) or `pass` in the body as the decorator alone prevents the class to be instantiated when the method is not overridden. Meaning, it is impossible to hit this error.

ragna/source_storages/_embedding_model.py
```diff
@@ -0,0 +1,47 @@
...
+
+    def embed_text(self, text: str) -> list[float]:
+        raise NotImplementedError
```
pmeier, line 31: Is there ever a use case to do that? Aren't we always going to embed chunks?

ragna/source_storages/_embedding_model.py
```diff
@@ -0,0 +1,47 @@
+from abc import ABC, abstractmethod
+
+from sentence_transformers import SentenceTransformer
+import torch
```
pmeier, line 4: PyTorch is a massive dependency that we cannot pull in by default. This has to be optional.

ragna/source_storages/_embedding_model.py
```diff
@@ -0,0 +1,47 @@
...
+
+from ragna.core._components import Component
+
+from ragna.source_storages._vector_database import Chunk
+
+device = 'cuda' if torch.cuda.is_available() else 'cpu'
```
pmeier, line 10: 1. We cannot hardcode this. What if the user wants to run the embedding on their CPU even if they have a GPU? This being a default is fine though.
2. What is the performance when running this model on CPU? How many tokens per second can we expect here? I'd rather not include a model that basically requires a GPU to be somewhat usable.

ragna/source_storages/_lancedb.py
```diff
@@ -1,10 +1,14 @@
...
 
+from typing import cast
+
 import ragna
 from ragna.core import Document, PackageRequirement, Requirement, Source
 
 from ._vector_database import VectorDatabaseSourceStorage
 
+from ._embedding_model import Embedding
+import pyarrow as pa
```
pmeier, line 11: Import this locally where needed to avoid a hard dependency.

ragna/source_storages/_embedding_model.py
```diff
@@ -0,0 +1,47 @@
...
+    embedding: list[float]
+    chunk: Chunk
+
+    def __init__(self, embedding: list[float], chunk: Chunk):
+        super().__init__()
+        self.embedding = embedding
+        self.chunk = chunk
+
+
+class GenericEmbeddingModel(Component, ABC):
```
pmeier, line 23: This class as well as the `Embedding` class above should be moved into `ragna.core._components` given that we want to separate them from the source storages.

.gitignore
```diff
@@ -6,6 +6,10 @@ __pycache__/
 *.py[cod]
 *$py.class
 
+# Ignore my entrypoint
+test.py
+doctest/
```
pmeier, line 11: Flag to revert them before merge.

pmeier (commented):
IMO, we should split this PR into two parts. One that adds the embedding models and one that adds chunking. Otherwise this will be really hard to review. Assuming we want to do embedding models first, I would totally be ok with `EmbeddingModel` receiving documents and performing the chunking internally as it is done currently in the source storages. In the follow-up PR we can factor it out into the chunking.

Tengal-Teemo:
@pmeier I've read all your insights, and I'll be updating the PR accordingly.

pmeier:
@Tengal-Teemo I took the liberty and implemented the automatic type discovery that I pitched in https://github.com/Quansight/ragna/pull/354#discussion_r1526215146 in 34fd3ee07fec3dd9d3ac15b9197b7cfb66bbffbe. Now every `SourceStorage` will have a `__ragna_input_type__` class attribute that holds either `Document` or `Embedding`. We can use this to switch the input type in the `Chat` object.

ragna/source_storages/_chroma.py
Tengal-Teemo: These variables represent the average length of a chunk within the source storage, It is aggregated across calls to store.  I'm not sure what scope you're referring to but I don't think they can be local for them to work.

ragna/source_storages/_embedding_model.py
Tengal-Teemo: No, because it specified the embedding dimensions of the model.  Models create an embedding of N dimensions.  But yes I will just remove it entirely.

ragna/source_storages/_chroma.py
Tengal-Teemo: It doesn't, and these have been removed.

ragna/core/_rag.py
Tengal-Teemo: So something like?
``` python
        if type(self.source_storage).__ragna_input_type__ == Document:
            await self._run(self.source_storage.store, self.documents)
        else:
            chunks = self.chunking_model.chunk_documents(documents=self.documents)
            if type(self.source_storage).__ragna_input_type__ == Chunk:
                await self._run(self.source_storage.store, chunks)
            else:
                embeddings = self.embedding_model.embed_chunks(chunks)
                await self._run(self.source_storage.store, embeddings)
```

Tengal-Teemo:
@pmeier I've created two new pull requests, one for embedding and one for chunking.  This PR shall be left open for now as a record of the comments you made.

pmeier:
We don't need to keep it open to still have access to the comments. Since we are not planning to add anything here, I'm going to close it. We can always re-open later if needs be.

Tengal-Teemo:
Sorry yea I meant I won't delete the branch

ragna/source_storages/_embedding_model.py
Tengal-Teemo: No, because when you prompt the database you must convert your prompt into an embedding when you query.  Either we turn this prompt into a chunk which I don't like, or just separate the logic like this.

PR #357, opened 2024-03-12
use JSONL streaming over SSE

### LINKED ISSUE #356: Use JSONL for streaming rather then SSE
Looking at the builtin assistants, there are currently three ways of streaming:

1. Using SSE
   - [Anthropic](https://github.com/Quansight/ragna/blob/4a667f9aced1970f8eab8aa9e01aaf8c67b05db0/ragna/assistants/_anthropic.py#L38)
   - [OpenAI](https://github.com/Quansight/ragna/blob/4a667f9aced1970f8eab8aa9e01aaf8c67b05db0/ragna/assistants/_openai.py#L38)
2. Using streaming responses
   1. with one object per line (JSONL):
      - [Cohere](https://github.com/Quansight/ragna/blob/4a667f9aced1970f8eab8aa9e01aaf8c67b05db0/ragna/assistants/_cohere.py#L59-L60)
   2. with streaming JSON: 
      - [Google](https://github.com/Quansight/ragna/blob/4a667f9aced1970f8eab8aa9e01aaf8c67b05db0/ragna/assistants/_google.py#L64)

Out of those, I think 2.ii. is by far the worst as it is the hardest to parse. When we implemented streaming for Ragna, I went with 1., since that is what I was familiar with and seen other providers to as well. At that point, I was not aware of streaming JSONL (2.i.) as an alternative.

2.i. has the upside, that one does not need an extra library to handle as it is "required" for the other two cases (of course we could write the logic ourselves as well).

Due to its simplicity I would actually prefer JSONL streaming. We should make a decision here before we release streaming with `0.2.0`.

pmeier:
Closes #356.

### DIFF
--- a/docs/examples/gallery_streaming.py
+++ b/docs/examples/gallery_streaming.py
@@ -145,41 +145,38 @@ chat = (
 client.post(f"/chats/{chat['id']}/prepare").raise_for_status()
 
 # %%
-# Streaming the response is performed with
-# [server-sent events (SSE)](https://en.wikipedia.org/wiki/Server-sent_events).
+# Streaming the response is performed with [JSONL](https://jsonlines.org/). Each line
+# in the response is valid JSON and corresponds to one chunk.
 
-import httpx_sse
 import json
 
-chunks = []
-with httpx_sse.connect_sse(
-    client,
+
+with client.stream(
     "POST",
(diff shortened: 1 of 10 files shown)


aktech (approved):
Looks good to me.

PR #358, opened 2024-03-13
[DOC] Add tutorial for adding your own components

smokestacklightnin:
This is a work in progress

This PR adds a tutorial to help users or developers add assistants and source storages for LLMs to Ragna that are not currently supported


docs/tutorials/gallery_adding_an_assistant.py
```diff
@@ -0,0 +1,177 @@
...
+Ragna has builtin support for [several assistants][ragna.assistants], but there may be cases where you 
+want to use one that is not currently supported.
+
+This tutorial walks you through the basics of creating an assistant for LLMs that are not
+currently supported. 
+
+!!! note
+
+    This tutorial assumes that our TutorialAssistant class (shown below) is located in the
+    file ragna/assistants/_tutorial.py.
```
pmeier, line 13: We cannot assume that. The point of this tutorial is to explain how users can implement their own assistant without forking Ragna and putting their stuff into the source code.

docs/tutorials/gallery_adding_an_assistant.py
```diff
@@ -0,0 +1,177 @@
...
+# %%
+# ### Step 0: Import Necessary Modules
+
+# %%
+# First start by importing some necessary components. Ragna uses the `typing` library:
+
+# %%
+# ```python
+# from typing import Iterator
+# ```
```
pmeier, line 88: This is not needed for two reasons:

1. The `Iterator` annotation is not necessary for this to function.
2. If one wishes to use type annotations, they will be familiar with the `typing` module. Since it is part of the standard library, it also comes pre-installed.

docs/tutorials/gallery_adding_an_assistant.py
```diff
@@ -0,0 +1,177 @@
...
+#          )
+# ```
+
+
+# %%
+# !!! note
+#     While including the `_default_answer` method seems to obfuscate the code, it is meant
+#     to demonstrate that you can add multiple methods to your assistant that help compute
+#     the response. Examples include calling an external API or preparing the sources to be
+#     used by the model.
```
pmeier, line 148: I don't understand the reasoning. We can assume that someone who wants to write their own assistant, comes with enough programming knowledge to know that you can call a function from another function. Thus, there is no reason for us to obfuscate our code here. Let's keep it as simple as possible.

docs/tutorials/gallery_adding_an_assistant.py
```diff
@@ -0,0 +1,177 @@
...
+
+# %%
+# ### Step 2: Include the Assistant in Ragna
+
+# %%
+# Once you have created your assistant, you must add it to the system so that Ragna recognizes it.
+# To do this, add your custom assistant to the `__all__` list in the file
+# `ragna/assistants/__init__.py`, and import it in the same file. An example is shown below:
```
pmeier, line 156: Please have a look at https://ragna.chat/en/latest/references/config/#referencing-python-objects on how to use components defined in external files.

docs/tutorials/gallery_adding_an_assistant.py
```diff
@@ -0,0 +1,177 @@
...
+        """Answer a prompt given some sources.
+
+        Args:
+            prompt: Prompt to be answered.
+            sources: Sources to use when answering answer the prompt.
+
+        Returns:
+            Answer.
+        """
+        yield self._default_answer(prompt, sources)
```
pmeier, line 47: We should definitely mention that you can stream form the assistant by `yield`ing multiple times. Not sure whether we do it as code comment here or as a note down below. We can also link to the [streaming example](https://ragna.chat/en/latest/generated/examples/gallery_streaming/).

pmeier (changes requested):
I've added some high level comments. I'll only review the actual text when the structure is a place where we can move forward.

Apart from this, there is one thing that I'm missing: a usage example. We should have one for Python and REST API here and maybe have a screenshot for the UI (not sure about that yet).

You can have a look at the [Python API](https://ragna.chat/en/latest/generated/tutorials/gallery_python_api/) and [REST API](https://ragna.chat/en/latest/generated/tutorials/gallery_rest_api/) tutorials on how to do that. The [streaming example](https://ragna.chat/en/latest/generated/examples/gallery_streaming/#rest-api) includes some code on how to integrate a custom assistant when using the REST API in the documentation.

docs/tutorials/gallery_adding_an_assistant.py
```diff
@@ -0,0 +1,110 @@
...
+    def answer(self, prompt, sources):
+        """Answer a prompt given some sources.
+
+        Args:
+            prompt: Prompt to be answered.
+            sources: Sources to use when answering answer the prompt.
+
+        Returns:
+            Answer.
+        """
```
pmeier, line 37: Maybe we shouldn't add a docstring here given that we cannot link to `Source` or the like. Maybe we should just explain the method and its inputs in a block before we show the implementation?

docs/tutorials/gallery_adding_an_assistant.py
```diff
@@ -0,0 +1,110 @@
...
+#     The code snippet below only includes up to Step 1. To actually include your source
+#     storage in Ragna, you must still perform Step 2.
+
+from ragna.core import Assistant
+
+
+class TutorialAssistant(Assistant):
+    """
+    This is a basic assistant created for didactic purposes
+    """
```
pmeier, line 26: This adds little value but introduces vertical space that makes it harder to grasp the full code.

```suggestion
```

docs/tutorials/gallery_adding_an_assistant.py
```diff
@@ -0,0 +1,110 @@
...
+#
+#     For organizational purposes, this tutorial is divided into steps, but you may wish to perform
+#     Step 2 during Step 1 for manual testing or debugging.
+
+
+# %%
+# ### Step 0: Import Necessary Modules
+
+# %%
+# First start by importing some necessary components.
```
pmeier, line 60: You misunderstood my comment in https://github.com/Quansight/ragna/pull/358#discussion_r1526158304. Type hints are really useful for tutorials. Please put them back. What I meant is that you don't need to explain the `typing` import.

docs/tutorials/gallery_adding_an_assistant.py
```diff
@@ -0,0 +1,110 @@
...
+
+
+# %%
+# ## The Explanation
+
+# %%
+# !!! tip
+#
+#     For organizational purposes, this tutorial is divided into steps, but you may wish to perform
+#     Step 2 during Step 1 for manual testing or debugging.
```
pmeier, line 53: I feel doing this in steps is harder than just doing it in one go. You now have to replicate the code in two locations.

docs/tutorials/gallery_adding_an_assistant.py
```diff
@@ -0,0 +1,110 @@
...
+# ```
+
+# %%
+# !!! note
+#     Ragna also supports streaming responses from the assistant. See the
+#     [example how to use streaming responses](../../generated/examples/gallery_streaming.md)
+#     for more information.
+
+# %%
+# See the documentation on [how to include external objects in Ragna](../../references/config.md#referencing-python-objects)
```
pmeier, line 107: Although the general process is explained there, we should make a concrete example here. Something along the lines of: "assuming you put the implementation above into a `tutorial.py` module, you can now add it to your configuration file like this: ..."

smokestacklightnin:
Closing this PR and continuing it on PR #368

PR #361, opened 2024-03-13
fix release notes link in project metadata

pmeier:
1. We renamed the file.
2. We should point at the release notes of the current stable release rather than the dev version.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -49,7 +49,7 @@ dynamic = ["version"]
 [project.urls]
 Homepage = "https://ragna.chat"
 Documentation = "https://ragna.chat"
-Changelog = "https://ragna.chat/en/latest/references/changelog/"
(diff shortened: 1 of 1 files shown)


smokestacklightnin (commented):
The link works for me but it the page doesn't render properly for me. It gives a 404 not found response and the rendering looks broken

pmeier:
Sorry for not providing enough context. This is similar to #343. The link will work after the 0.2.0 release. Right now the correct link is https://ragna.chat/en/latest/references/release-notes/. When the release is published, what currently is in `/latest` will be moved to `/stable` and thus making the link in this PR valid.

However, in contrast to #343, this link is only in the `pyproject.toml` rather than in the very visible README. Thus, we can merge this already and don't need to wait until the last second.

smokestacklightnin (approved):

Issue #191, opened 2023-11-10
embedding model component?


pmeier:
Currently, the `SourceStorage` is responsible for extracting the text from the document, embedding it and finally storing it. In this issue, I want to discuss whether it makes sense for us to factor out the extraction and embedding logic into a separate `EmbeddingModel` component on the same level as `SourceStorage` and `Assistant`. Here is what would change:

- The new `EmbeddingModel` class would have an `.embed()` method that takes a list of documents as inputs and returns a list of `Embedding`s.
- An `Embedding` is a dataclass that holds two things:
  1. The chunk, i.e. the actual text, the number of tokens, the document it was created from, and if available the location of the chunk inside said document
  2. The numerical values (probably `list[float]`) of the embedded text
- The `SourceStorage.store()` method would no longer take a list of documents, but rather a list of embeddings as input.
- The chunking parameters like `chunk_size` and `chunk_overlap` would move from the `SourceStorage` to the `EmbeddingModel`. That means that for all currently implemented source storages, there is no configuration anymore. This is not a bad thing, just pointing it out.

We have not implemented it that way from the start to allow source storages where the user has no control over the extraction and embedding. For example [AWS Kendra](https://aws.amazon.com/kendra/) needs to take the full document and does the extraction and embedding in the background. Meaning, it will not **not** fit the schema above.

That being said, AWS Kendra will under current policy not land in Ragna, since it requires additional infrastructure. And for custom solution it is always possible to fall back to a custom `class NoEmbedding(Embedding)` that doesn't do any embedding, but just stores the document so the source storage can do its thing.

So the question here is: do we want to have a separate `EmbeddingModel` component? It would make it easier for users to try new things, but makes it potentially harder on source storages that don't operate like a "regular" vector database.

dharhas:
So on another project, we were looking into a more mature RAG pipeline that has been in development for a while and the approach that was implemented is actually even more fine grained. I don't have the chart in front of me but document to embedding to storage was a multi-step pipeline with configurable components including a summarization and metadata extraction step to annotate the chunks. To be fair, they were also looking at supporting extracting structure from powerpoint and tables etc.

Overall, I think it make sense to go this direction. 

Would it enale the embed once and use multiple LLM's workflow or is that orthogonal?

pmeier:
> Would it enale the embed once and use multiple LLM's workflow or is that orthogonal?

It is orthogonal. The point is to make the embedding model a "first class citizen". Right now, everything is hardcoded to

[...]

PR #364, opened 2024-03-15
Add tutorial for adding your own source storages

smokestacklightnin:
This is a work in progress

This PR adds a tutorial to help users or developers add assistants for source storages to Ragna that are not currently supported

### DIFF
--- a/docs/tutorials/gallery_adding_source_storages.py
+++ b/docs/tutorials/gallery_adding_source_storages.py
@@ -43,7 +43,9 @@ class TutorialSourceStorage(SourceStorage):
             for document in documents
         ]
 
-    def retrieve(self, sources: list[Source], prompt: str) -> list[Source]:
+    def retrieve(
+        self, documents: list[Document], prompt: str, *, chat_id: int
+    ) -> list[Source]:
         """Retrieve sources for a given prompt.
 
         Args:
@@ -53,7 +55,7 @@ class TutorialSourceStorage(SourceStorage):
         Returns:
             Matching sources for the given prompt ordered by relevance.
         """
-        return sources
+        return self._storage[chat_id]
 
 
 # %%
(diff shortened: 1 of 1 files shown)


docs/tutorials/gallery_adding_source_storages.py
```diff
@@ -0,0 +1,168 @@
...
+
+
+class TutorialSourceStorage(SourceStorage):
+    def __init__(self):
+        # import database api
+
+        # set up database
+        ...
+
+    def store(self, sources: list[Source]) -> None:
```
smokestacklightnin, line 37: The `store()` and `retrieve()` methods of the `SourceStorage` ABC both have lists of `Document` types as parameters, but `Document` is part of `ragna.core._document`, which is a private module. I have instead replaced `Document` in the tutorial with `Source`, which is public. I did this (1) because I didn't want to make reference to any private parts of the codebase, and (2) because I wanted to be able to link to documentation of the types in the arguments to `store()` and `retrieve()`. @pmeier, do you have thoughts or input on this decision?

pmeier: > The `store()` and `retrieve()` methods of the `SourceStorage` ABC both have lists of `Document` types as parameters, but `Document` is part of `ragna.core._document`, which is a private module.

That is true, but they are exposed under `ragna.core`

https://github.com/Quansight/ragna/blob/5a9bdf69de6b4234831532c0d13797a7015a61bf/ragna/core/__init__.py#L34-L35

In general, we expose all the functionality that users need either under `ragna.core` or `ragna.deploy`. Anything below that is private. This has two upsides for us

1. We expose only exactly what we want given that we need put it in explicitly.
2. We can freely change module structure or the like in our implementation and only need to keep the upper namespace BC compliant.

> I have instead replaced `Document` in the tutorial with `Source`, which is public.

The two are two completely different objects and cannot be used interchangeably:
- A `Document` is the starting point that we extract information from. Right now this is handled by source storage directly, but that might change soon (see #191).
- A `Source` represents the information that we have extracted earlier.

PR #368, opened 2024-03-19
[DOC] Add tutorial for adding your own objects

smokestacklightnin:
This is a work in progress.

This is a continuation of PR #358. It was created by rebasing PR #358 to include the merged commits from PR #362 that removes `max_input_size`. 

This PR adds a tutorial to help users or developers add assistants and source storages for LLMs to Ragna that are not currently supported.

TODO: Add usage example with the Python API


docs/tutorials/gallery_adding_components.py
```diff
@@ -0,0 +1,125 @@
+"""
+# Adding Components to Ragna
+
+Ragna has builtin support for several [assistants][ragna.assistants] and 
+[source storages][ragna.source_storages], but there may be cases where you 
+want to use one that is not currently supported.
```
pmeier, line 6: I would phrase this a little different. Ragna was built with the goal in mind that you bring your own components. We only provide very basic ones inside the library. For everything else you'll have to implement your own.

docs/tutorials/gallery_adding_components.py
```diff
@@ -0,0 +1,125 @@
+"""
+# Adding Components to Ragna
```
pmeier, line 2: The title is visible in the side bar and thus should be as concise as possible.

```suggestion
# Custom Components
```

docs/tutorials/gallery_adding_components.py
```diff
@@ -0,0 +1,125 @@
...
+Ragna has builtin support for several [assistants][ragna.assistants] and 
+[source storages][ragna.source_storages], but there may be cases where you 
+want to use one that is not currently supported.
+
+This tutorial walks you through the basics of adding components
+that are not currently officially supported. 
+"""
+
+# %%
+# ## Adding an LLM Assistant
```
pmeier, line 13: 1. While probably true in practice, the `Assistant` doesn't have to be an LLM. We can put this as text before the actual implementation.
2. Let's keep the header concise.

```suggestion
# ## Assistant
```

docs/tutorials/gallery_adding_components.py
```diff
@@ -0,0 +1,125 @@
...
+from typing import Iterator
+
+from ragna.core import Assistant, Source
+
+
+class TutorialAssistant(Assistant):
+    def answer(self, prompt: str, sources: list[Source]) -> Iterator[str]:
+        yield (
+            f"This is a default answer. There were {len(sources)} sources."
+            ""
```
pmeier, line 24: We can have a second look when we have the usage example and actually see the output, but I think we can get rid of this blank line.
```suggestion
```

docs/tutorials/gallery_adding_components.py
```diff
@@ -0,0 +1,125 @@
...
+# %%
+# The main thing to do is to implement the [`answer()`][ragna.core.Assistant.answer] abstract method.
+# The [`answer()`][ragna.core.Assistant.answer] method is where you put the logic to access your LLM.
+# This could call an API directly, call other member functions of your assistant that call an API,
+# or call a local LLM. Ragna is designed to give you that flexibility.
+
+# %%
+# Your [`answer()`][ragna.core.Assistant.answer] method should take a prompt in the form of a
+# string, and a list of [`Source`][ragna.core.Source]s, in addition to whatever other arguments
+# necessary for your particular assistant. The return type is an [`Iterator`](https://docs.python.org/3/library/stdtypes.html#typeiter) of strings.
```
pmeier, line 39: I would move these explanations before the implementation.

docs/tutorials/gallery_adding_components.py
```diff
@@ -0,0 +1,125 @@
...
+
+# %%
+# The main thing to do is to implement the [`answer()`][ragna.core.Assistant.answer] abstract method.
+# The [`answer()`][ragna.core.Assistant.answer] method is where you put the logic to access your LLM.
+# This could call an API directly, call other member functions of your assistant that call an API,
+# or call a local LLM. Ragna is designed to give you that flexibility.
+
+# %%
+# Your [`answer()`][ragna.core.Assistant.answer] method should take a prompt in the form of a
+# string, and a list of [`Source`][ragna.core.Source]s, in addition to whatever other arguments
```
pmeier, line 38: We should add a quick explanation what `Source`s actually are. Recall that this is likely the first time a user of Ragna actually comes in touch with them.

docs/tutorials/gallery_adding_components.py
```diff
@@ -0,0 +1,125 @@
...
+            f"The prompt was"
+            f"{prompt}"
+        )
+
+
+# %%
+# The main thing to do is to implement the [`answer()`][ragna.core.Assistant.answer] abstract method.
+# The [`answer()`][ragna.core.Assistant.answer] method is where you put the logic to access your LLM.
+# This could call an API directly, call other member functions of your assistant that call an API,
+# or call a local LLM. Ragna is designed to give you that flexibility.
```
pmeier, line 34:
```suggestion
# For example, this could call an API directly or call a local LLM.
```

docs/tutorials/gallery_adding_components.py
```diff
@@ -0,0 +1,125 @@
...
+# necessary for your particular assistant. The return type is an [`Iterator`](https://docs.python.org/3/library/stdtypes.html#typeiter) of strings.
+
+# %%
+# !!! note
+#     Ragna also supports streaming responses from the assistant. See the
+#     [example how to use streaming responses](../../generated/examples/gallery_streaming.md)
+#     for more information.
+
+# %%
+# ## Adding a Source Storage
```
pmeier, line 48: We should move this before the assistant, because

1. the source storage is used before the assistant in the RAG workflow
2. the source storage creates `Source`s that will be consumed by the assistant. If we explain what they are here, we can keep it short in the assistant example.

docs/tutorials/gallery_adding_components.py
```diff
@@ -0,0 +1,125 @@
...
+
+from ragna.core import Document, Source, SourceStorage
+
+
+class TutorialSourceStorage(SourceStorage):
+    def __init__(self):
+        # import database api here
+
```
pmeier, line 56:
```suggestion
```

docs/tutorials/gallery_adding_components.py
```diff
@@ -0,0 +1,125 @@
...
+# %%
+# ## Adding an LLM Assistant
+
+from typing import Iterator
+
+from ragna.core import Assistant, Source
+
+
+class TutorialAssistant(Assistant):
```
pmeier, line 20: Before we actually put the code, we need to explain what the `TutorialAssistant` will do. Something along the lines of "for simplicity, we are not going to implement an actual LLM here, but rather a demo assistant that just mirrors back the inputs. This is similar as the RagnaDemoAssistant".

docs/tutorials/gallery_adding_components.py
```diff
@@ -0,0 +1,125 @@
...
+    def __init__(self):
+        # import database api here
+
+        # set up database
+        self._storage: dict[int, list[Source]] = {}
+
+    def store(self, documents: list[Document], chat_id: int) -> None:
+        self._storage[chat_id] = [
+            Source(
+                document=document,
```
pmeier, line 63: Lose the trailing comma for the autoformatter to inline this

```suggestion
                document=document
```

docs/tutorials/gallery_adding_components.py
```diff
@@ -0,0 +1,125 @@
...
+
+
+class TutorialSourceStorage(SourceStorage):
+    def __init__(self):
+        # import database api here
+
+        # set up database
+        self._storage: dict[int, list[Source]] = {}
+
+    def store(self, documents: list[Document], chat_id: int) -> None:
```
pmeier, line 60: There are three things here:

1. The `chat_id` is not an `int`, but rather a `uuid.UUID`.
2. So far we have not explained what `chat_id` is and where it comes from. It is not part of the standard API:

    https://github.com/Quansight/ragna/blob/c6bdb0f9f5f71b37c0c312280839ab58d3059e34/ragna/core/_components.py#L102-L103

    https://github.com/Quansight/ragna/blob/c6bdb0f9f5f71b37c0c312280839ab58d3059e34/ragna/core/_components.py#L111-L112

    This deserves a full section below. Meaning at this point, we need to tell the reader "we need a way to identify the chat that we are storing documents for. We can use the unique `chat_id` for this. To get access to it, just add it as parameter to method signature. We'll explain how this works in section X".
3. It is convention to separate the non-standard parameters to be keyword-only.
    ```suggestion
        def store(self, documents: list[Document], *, chat_id: int) -> None:
    ```

docs/tutorials/gallery_adding_components.py
```diff
@@ -0,0 +1,125 @@
...
+            for document in documents
+        ]
+
+    def retrieve(
+        self, documents: list[Document], prompt: str, *, chat_id: int
+    ) -> list[Source]:
+        return self._storage[chat_id]
+
+
+# %%
```
pmeier, line 74: Same as for the assistant. Move this before the implementation.

docs/tutorials/gallery_adding_components.py
```diff
@@ -0,0 +1,125 @@
...
+# [`store()`][ragna.core.SourceStorage.store] takes a list of [`Source`][ragna.core.Source]s
+# as an argument and places them in the database that you are using to hold them. This is
+# different for each different source storage implementation.
+
+# %%
+# [`retrieve()`][ragna.core.SourceStorage.retrieve] returns sources matching
+# the given prompt in order of relevance.
+
+# %%
```
pmeier, line 87: The usage example for the Python API should be added before this section.

docs/tutorials/gallery_adding_components.py
```diff
@@ -0,0 +1,125 @@
...
+# ## Including External Python Objects
```
pmeier, line 88: We should rename this to reflect that you need to do this if you want to access your custom components from the REST API or web UI.

docs/tutorials/gallery_adding_components.py
```diff
@@ -0,0 +1,125 @@
...
+# %%
+# If the external object is already in your
+# [`PYTHONPATH`](https://docs.python.org/3/using/cmdline.html#envvar-PYTHONPATH), e.g.
+# already in your virtual environment, then the above `#!bash export` command is not necessary.
+
+# %%
+# Once the module(s) containing your objects is(are) in your
+# [`PYTHONPATH`](https://docs.python.org/3/using/cmdline.html#envvar-PYTHONPATH), you can
+# you can include them in Ragna by setting their corresponding
+# [environment variables](../../references/config.md#environment-variables):
```
pmeier, line 109: While correct, the regular way is for users to add the components to their configuration file, e.g. https://ragna.chat/en/stable/references/config/#referencing-python-objects.

pmeier (dismissed):

docs/tutorials/gallery_adding_components.py
pmeier: Re 2.: Here is how it works:

When adding new parameters to the signature of the protocol methods

https://github.com/Quansight/ragna/blob/c6bdb0f9f5f71b37c0c312280839ab58d3059e34/ragna/core/_components.py#L99-L100

https://github.com/Quansight/ragna/blob/c6bdb0f9f5f71b37c0c312280839ab58d3059e34/ragna/core/_components.py#L210-L213

we automatically pick up on this and add the parameters to the model of the component

https://github.com/Quansight/ragna/blob/c6bdb0f9f5f71b37c0c312280839ab58d3059e34/ragna/core/_components.py#L73-L76

When creating a new `Chat`, we check if the additional parameters satisfy the additional parameters of the selected components:

https://github.com/Quansight/ragna/blob/c6bdb0f9f5f71b37c0c312280839ab58d3059e34/ragna/core/_rag.py#L251

You don't need to understand this in detail. The mental model that you can use is that the extra `params` that you can pass to the `Chat`

https://github.com/Quansight/ragna/blob/c6bdb0f9f5f71b37c0c312280839ab58d3059e34/ragna/core/_rag.py#L77-L84

are automatically checked for completeness and distributed to the components at runtime. Meaning, every parameter you define without a default, must be passed when creating the chat. If you add a default value, this can be overridden by passing a value to the chat, but you don't have to.

So why didn't we need to pass the `chat_id` although it didn't have a default? `chat_id` is a special parameter alongside `chat_name` and `user`:

https://github.com/Quansight/ragna/blob/c6bdb0f9f5f71b37c0c312280839ab58d3059e34/ragna/core/_rag.py#L103-L108

They are automatically set by Ragna and can always be added to a signature without the need for the user to supply them. Same as the other parameters they can be overridden by passing other values when creating the chat. This is what we do for the REST API:

https://github.com/Quansight/ragna/blob/c6bdb0f9f5f71b37c0c312280839ab58d3059e34/ragna/deploy/_api/core.py#L204-L207

docs/tutorials/gallery_adding_components.py
```diff
@@ -0,0 +1,311 @@
...
+# %%
+# ## Using the Web UI with Custom Objects
+
+# %%
+# To include our custom objects in the Web UI, make sure they are in your
+# [`PYTHONPATH`](https://docs.python.org/3/using/cmdline.html#envvar-PYTHONPATH),
+# and list them in your [configuration file](../../references/config.md).
+
+# %%
+# Here is an example configuration file in the case that you would like to use one:
```
pmeier, line 166: We don't need a full configuration file here. We can point to the reference in the documentation and just put the important configuration here. We don't want to drown the useful information in the noise.

docs/tutorials/gallery_adding_components.py
```diff
@@ -0,0 +1,311 @@
...
+import documentation_helpers
+
+from ragna.deploy import Config
+
+import httpx
+
+config = Config()
+
+config.source_storages = [TutorialSourceStorage]
+config.assistants = [TutorialAssistant]
```
pmeier, line 228:
```suggestion
config = Config(source_storages=[TutorialSourceStorage], assistants=[TutorialAssistant])
```

docs/tutorials/gallery_adding_components.py
```diff
@@ -0,0 +1,311 @@
...
+    "/token",
+    data={"username": username, "password": password},
+).raise_for_status()
+token = response.json()
+
+# %%
+# We set the API token on our HTTP client so we don't have to manually supply it
+# for each request below.
+
+client.headers["Authorization"] = f"Bearer {token}"
```
pmeier, line 249: This will give you a fully authenticated client

```suggestion
client = rest_api.start(config, authenticate=True)
```

docs/tutorials/gallery_adding_components.py
```diff
@@ -0,0 +1,311 @@
...
+
+# %%
+# Uploading the documents takes place in two steps. See
+# [the REST API tutorial](../../generated/tutorials/gallery_rest_api.md#step-3-uploading-documents)
+# for more details.
```
pmeier, line 254: Since we have the disclaimer above on where to find detailed information, we can cut this short here. We don't need to explain in detail the parts in this tutorial that are necessary, but not interesting to drive the narrative.

```suggestion
# Next, we upload the documents.
```

docs/assets/images/ragna-tutorial-components.png
pmeier, line 1: This needs to be a `git lfs` pointer rather than the actual image. See https://ragna.chat/en/stable/community/contribute/#view-add-or-update-images. I don't have push access to your fork, so I wasn't able to resolve this.

pmeier (commented):
I've fixed most of my own comments. There is some slight cleanup needed, but this is good for a first round of review from someone else.

docs/tutorials/gallery_custom_components.py
```diff
@@ -0,0 +1,415 @@
...
+response = client.post(
+    f"/chats/{chat['id']}/answer",
+    json={"prompt": "What is Ragna?"},
+).raise_for_status()
+answer = response.json()
+print(json.dumps(answer, indent=2))
+
+# %%
+# Let's stop the REST API and have a look at what would have printed in the terminal if
+# we had started it the regular way.
```
pavithraes, line 226: I think we should clearly state what the _regular way_ is (i.e. say that instead of regular), since it's unclear to a reader here (e.g. me!).

Do you mean through the CLI's `api` command here? If so, let's state that here.

docs/tutorials/gallery_custom_components.py
```diff
@@ -0,0 +1,415 @@
...
+# !!! warning
+#
+#     See the section about [custom parameters for the web UI](#web-ui_1) for some
+#     limitations using custom components in the web UI.
+
+# %%
+# ## Custom parameters
+#
+# Ragna supports passing parameters to the components on a per-chat level. The extra
+# parameters are defined by simply adding them to the signature of the method. For
```
pavithraes, line 250:
```suggestion
# parameters are defined by adding them to the signature of the method. For
```

docs/tutorials/gallery_custom_components.py
```diff
@@ -0,0 +1,415 @@
...
+# example, let's define a more elaborate assistant. The `my_required_parameter` has to
+# be passed and has to be an `int`, while the `my_optional_parameter` can be passed, but
+# still has to be a string.
```
pavithraes, line 253:
```suggestion
# example, let's define a more elaborate assistant that accepts the following arguments:
#
# * `my_required_parameter` must be passed and has to be an `int`
# * `my_optional_parameter` may be passed and has to be a `str` if passed
```

docs/tutorials/gallery_custom_components.py
```diff
@@ -0,0 +1,415 @@
...
+        print(f"Running {type(self).__name__}().answer()")
+        yield (
+            f"I was given {my_required_parameter=} and {my_optional_parameter=}."
+        )
+
+
+# %%
+# ### Python API
+#
+# To pass custom parameters to the components, simply pass them as keyword arguments
```
pavithraes, line 274:
```suggestion
# To pass custom parameters to the components, pass them as keyword arguments
```

docs/tutorials/gallery_custom_components.py
```diff
@@ -0,0 +1,415 @@
...
+#
+#     When using Ragnas builtin components, all source storages cover parameters 1. to
+#     3. and all assistant cover parameter 4.
+
+# %%
+# ## Sync or `async`?
+#
+# So far we have `def`ined all of the abstract methods of the components as regular,
+# i.e. synchronous methods. However, this is *not* a requirement. If you want to run
+# `async` code inside the methods, just define them with `async def` and Ragna will
```
pavithraes, line 385:
```suggestion
# `async` code inside the methods, define them with `async def` and Ragna will
```

docs/tutorials/gallery_custom_components.py
```diff
@@ -0,0 +1,415 @@
...
+
+# %%
+# ### Web UI
+#
+# !!! warning
+#     Unfortunately, Ragnas web UI currently does **not** support arbitrary custom
+#     parameters. This is actively worked on and progress is tracked in
+#     [#217](https://github.com/Quansight/ragna/issues/217). Until this is resolved,
+#     the source storage and the assistant together need to take the following four
+#     parameters:
```
pavithraes, line 370:
```suggestion
#     Unfortunately, Ragna's web UI currently does **not** support arbitrary custom
#     parameters. This is actively worked on and progress is tracked in
#     [#217](https://github.com/Quansight/ragna/issues/217). Until this is resolved,
#     the source storage and the assistant together take the following four
#     parameters:
```

docs/tutorials/gallery_custom_components.py
```diff
@@ -0,0 +1,415 @@
...
+#
+#     1. `chunk_size: int`
+#     2. `chunk_overlap: int`
+#     3. `num_tokens: int`
+#     4. `max_new_tokens: int`
+#
+#     When using Ragnas builtin components, all source storages cover parameters 1. to
```
pavithraes, line 377:
```suggestion
#     When using Ragna's builtin components, all source storages cover parameters 1. to
```

docs/tutorials/gallery_custom_components.py
```diff
@@ -0,0 +1,415 @@
...
+# %%
+# ### Web UI
+#
+# The setup for the web UI is exactly the same as for the [REST API](#rest-api). If you
+# have your configuration set up, you can start the web UI and use the custom
+# components. See the [web UI tutorial](../../generated/tutorials/gallery_rest_api.md)
+# for details.
+#
+#
+# ![Create-chat-modal of Ragnas web UI with the TutorialSourceStorage and TutorialAssistant as selected options](../../assets/images/ragna-tutorial-components.png)
```
pavithraes, line 239:
```suggestion
# ![Create-chat-modal of Ragna's web UI with the TutorialSourceStorage and TutorialAssistant as selected options](../../assets/images/ragna-tutorial-components.png)
```

pavithraes (changes requested):
Thanks for working on this! Most of the suggestions are related to language being used, specifically [avoiding phrases like simply or just](https://developers.google.com/style/tone#some-things-to-avoid-where-possible) and using [active voice](https://developers.google.com/style/voice).

I have a note on my end to improve the existing tutorials to also use second person in the tutorials (i.e. "you" rather than "we") on this project, so not including those as review suggestions here.

docs/tutorials/gallery_custom_components.py
```diff
@@ -0,0 +1,415 @@
...
+
+# %%
+# ### REST API
+#
+# To be able to use custom components in Ragnas REST API or web UI, two things need to
+# happen:
+#
+# 1. We need move the code into a Python module or package, e.g. `tutorial.py` that is
+#    on the
+#    [`PYTHONPATH`](https://docs.python.org/3/using/cmdline.html#envvar-PYTHONPATH),
```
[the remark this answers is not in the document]
pmeier: > let's suggest `sys.path` here?

That is not possible. `sys.path` is only available from inside Python, but when starting the API or UI, the user never runs Python directly. They have to make their custom components importable "from the outside". Let me try to rephrase this to make it more clear what needs to happen.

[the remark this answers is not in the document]
pmeier: @pavithraes In 737e322e5039974f9c0cdb1973173b93e5f92833 I have shortened this section to only say that the code needs to be on the search path and included a link to our config reference where this is explained in more detail. Please have another look.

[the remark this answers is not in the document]
pavithraes: Thanks, looks good :)

pavithraes (approved):
Thanks for the updates, looks good to me :)

PR #369, opened 2024-03-19
Embedding pr

Tengal-Teemo:
@pmeier 
This is the first of the two ongoing pull requests that will represent the embedding and chunking models.  All suggested changes have been implemented.


pmeier:
The title says it is the "Embedding pr", but I see quite a bit code for the chunking as well. Can we remove that to keep the review manageable?

Tengal-Teemo:
You are correct, I forgot to push a rebase from yesterday that dropped the chunking commits.  It should be fixed now.

ragna/core/_components.py
```diff
@@ -76,6 +89,29 @@ def _protocol_model(cls) -> Type[pydantic.BaseModel]:
         return merge_models(cls.display_name(), *cls._protocol_models().values())
 
 
+# Just for demo purposes. We need to move the actual class here.
+# See https://github.com/Quansight/ragna/pull/354#discussion_r1526235318
+class Embedding:
```
pmeier, line 94: Let's make this a dataclass if we have no logic on it. We should do the same for `Source` below. We don't need a Pydantic model there.

ragna/core/_components.py
```diff
@@ -76,6 +89,29 @@ def _protocol_model(cls) -> Type[pydantic.BaseModel]:
...
+    embedding: list[float]
+    chunk: Chunk
+
+    def __init__(self, embedding: list[float], chunk: Chunk):
+        super().__init__()
+        self.embedding = embedding
+        self.chunk = chunk
+
+
+class GenericEmbeddingModel(Component, ABC):
```
pmeier, line 104: Let's drop the "Generic" prefix. We don't have it for `Assistant` and `SourceStorage` as well

```suggestion
class EmbeddingModel(Component, ABC):
```

ragna/core/_components.py
```diff
@@ -76,6 +89,29 @@ def _protocol_model(cls) -> Type[pydantic.BaseModel]:
...
+    _EMBEDDING_DIMENSIONS: int
```
pmeier, line 105: See https://github.com/Quansight/ragna/pull/354#discussion_r1529570625

ragna/core/_components.py
```diff
@@ -76,6 +89,29 @@ def _protocol_model(cls) -> Type[pydantic.BaseModel]:
...
+
+    @abstractmethod
+    def embed_chunks(self, chunks: list[Chunk]) -> list[Embedding]:
+        ...
+
+    def embed_text(self, text: str) -> list[float]:
+        ...
```
pmeier, line 112: See https://github.com/Quansight/ragna/pull/354#discussion_r1526230507

ragna/core/_components.py
```diff
@@ -76,6 +89,29 @@ def _protocol_model(cls) -> Type[pydantic.BaseModel]:
...
+        super().__init__()
+        self.embedding = embedding
+        self.chunk = chunk
+
+
+class GenericEmbeddingModel(Component, ABC):
+    _EMBEDDING_DIMENSIONS: int
+
+    @abstractmethod
+    def embed_chunks(self, chunks: list[Chunk]) -> list[Embedding]:
```
pmeier, line 108: If we actually don't need the `embed_text` as I reckon below, we can probably name this just

```suggestion
    def embed(self, chunks: list[Chunk]) -> list[Embedding]:
```

?

ragna/source_storages/__init__.py
```diff
@@ -2,11 +2,15 @@
     "Chroma",
     "LanceDB",
     "RagnaDemoSourceStorage",
+    "MiniLML6v2",
+    "VectorDatabaseSourceStorage",
 ]
 
+from ._vector_database import VectorDatabaseSourceStorage
```
pmeier, line 9: Why did this become public? This is an internal helper class that we can probably remove entirely after the chunking PR is landed as well. Please make it private again.

ragna/source_storages/_embedding.py
pmeier, line 1: We need a new `ragna.embedding_models` namespace for this. We shouldn't group this under `ragna.source_storages`.

ragna/core/__init__.py
pmeier, line 1: We need to add `Embedding` and `EmbeddingModel` here so they can be accessed from `ragna.core`.

ragna/source_storages/_embedding.py
```diff
@@ -0,0 +1,16 @@
+from ragna.core import Chunk
+
+from ragna.core._components import Embedding, GenericEmbeddingModel
+
+
+class MiniLML6v2(GenericEmbeddingModel):
```
pmeier, line 6: How is this different from the one that comes prepackaged with `chromadb`? IIUC, they are the same: https://github.com/chroma-core/onnx-embedding. We should prefer that, because

1. Installing `chromadb` will also unlock Chroma as source storage
2. Compared to `sentence_transformers` `chromadb` is actually a lightweight dependency.

ragna/source_storages/_embedding.py
```diff
@@ -0,0 +1,16 @@
...
+
+    def __init__(self):
+        from sentence_transformers import SentenceTransformer
```
pmeier, line 9: See the other components on how you specify optional dependencies, e.g. 

https://github.com/Quansight/ragna/blob/aa2a70263b373baaf40b4c41701c850e40873ad9/ragna/source_storages/_vector_database.py#L51-L64

ragna/source_storages/_chroma.py
```diff
@@ -32,59 +32,62 @@ def __init__(self) -> None:
                 anonymized_telemetry=False,
             )
         )
+        self._tokens = 0
+        self._embeddings = 0
 
     def store(
         self,
-        documents: list[Document],
+        documents: list[Embedding],
```
pmeier, line 40: This should be 

```suggestion
        embeddings: list[Embedding],
```

now

ragna/source_storages/_chroma.py
```diff
@@ -32,59 +32,62 @@ def __init__(self) -> None:
...
             documents=texts,
+            embeddings=embeddings,
             metadatas=metadatas,  # type: ignore[arg-type]
         )
 
     def retrieve(
         self,
         documents: list[Document],
-        prompt: str,
+        prompt: list[float],
```
pmeier, line 79: Maybe

```suggestion
        embedded_prompt: list[float],
```

?

ragna/core/_rag.py
```diff
@@ -188,7 +196,17 @@ async def prepare(self) -> Message:
                 detail=RagnaException.EVENT,
             )
 
-        await self._run(self.source_storage.store, self.documents)
+        from ragna.core import Document
```
pmeier, line 199: We have that imported already

```suggestion
```

ragna/core/_rag.py
```diff
@@ -188,7 +196,17 @@ async def prepare(self) -> Message:
...
+        if type(self.source_storage).__ragna_input_type__ == Document:
```
pmeier, line 200: I would prefer something like

```suggestion
        if issubclass(self.source_storage.__ragna_input_type__, Document):
```

ragna/core/_rag.py
```diff
@@ -188,7 +196,17 @@ async def prepare(self) -> Message:
...
+            await self._run(self.source_storage.store, self.documents)
+        else:
+            # Here we need to generate the list of embeddings
+            chunks = []
+            for document in self.documents:
+                chunks += self.source_storage._chunk_pages(document.extract_pages(), document_id=document.id, chunk_size=500, chunk_overlap=250)
```
pmeier, line 206: This logic needs to be moved to the `MiniLML6v2` class. We can no longer put this on the source storage.

ragna/core/_rag.py
pmeier, line 200: Especially with the upcoming change for the chunking as well, I would prefer something like

```python
input = self.documents
if not issubclass(self.source_storage.__ragna_input_type__, Document):
    input = self.chunking_model(input)
if not issubclass(self.source_storage.__ragna_input_type__, Document):
    input = self.embedding_model(input)
await self._run(self.source_storage.store, input)
```

ragna/core/_rag.py
```diff
@@ -80,6 +83,7 @@ def chat(
         documents: Iterable[Any],
         source_storage: Union[Type[SourceStorage], SourceStorage],
         assistant: Union[Type[Assistant], Assistant],
+        embedding_model: Union[Type[GenericEmbeddingModel], GenericEmbeddingModel],
```
pmeier, line 86: We cannot make this required. Depending on the source storage, the user might not need to supply an embedding model.

```suggestion
        embedding_model: Optional[Union[Type[GenericEmbeddingModel], GenericEmbeddingModel]] = None,
```

And later in the `__init__` we can do something like

```python
if embedding_model is None and issubclass(source_storage.__ragna_input_type__, Embedding):
    raise RagnaException
```

pmeier (changes requested):
Thanks @Tengal-Teemo for moving forward with this. I did a first pass mostly about moving all the pieces into the right place. I think we are making good progress here. When everything is resolved here, I'll do another pass.

One thing that I don't really understand yet is https://github.com/Quansight/ragna/pull/354#discussion_r1529468766. But we'll come to that later.

ragna/core/_components.py
Tengal-Teemo: If we don't separate out the embed_text function, then you need to create a chunk holding the prompt in order to query the database.  I dislike this, so I would prefer to keep them separated.

pmeier: Right, makes sense.

ragna/source_storages/_chroma.py
Tengal-Teemo: There are some issues with doing this since they both stem from the same class which specifies a pydantic BaseModel.  calling them different names causes AttributeErrors.  Perhaps they could be called `data` or something more generic?  Forgive me if there is a simpler way of doing this, I am not very familiar with pydantic.

Tengal-Teemo:
@pmeier I believe I've either implemented or commented on all your changes.

ragna/core/_components.py
```diff
@@ -76,6 +96,119 @@ def _protocol_model(cls) -> Type[pydantic.BaseModel]:
         return merge_models(cls.display_name(), *cls._protocol_models().values())
 
 
+# Just for demo purposes. We need to move the actual class here.
+# See https://github.com/Quansight/ragna/pull/354#discussion_r1526235318
```
pmeier, line 100:
```suggestion
```

ragna/core/_components.py
```diff
@@ -76,6 +96,119 @@ def _protocol_model(cls) -> Type[pydantic.BaseModel]:
...
+@dataclass
+class Embedding:
+    embedding: list[float]
+    chunk: Chunk
+
+    def __init__(self, embedding: list[float], chunk: Chunk):
+        super().__init__()
+        self.embedding = embedding
+        self.chunk = chunk
```
pmeier, line 109: This is automated by the `@dataclass`

```suggestion
```

ragna/core/_rag.py
```diff
@@ -148,13 +153,19 @@ def __init__(
...
 
+        if embedding_model is None and issubclass(source_storage.__ragna_input_type__, Embedding):
+            raise RagnaException
+
+        self.embedding_model = cast(EmbeddingModel, self._rag._load_component(embedding_model))
+
         self.documents = self._parse_documents(documents)
         self.source_storage = cast(
-            SourceStorage, self._rag._load_component(source_storage)
+            VectorDatabaseSourceStorage, self._rag._load_component(source_storage)
```
pmeier, line 168: The VDB class is private

```suggestion
            SourceStorage, self._rag._load_component(source_storage)
```

ragna/core/_rag.py
pmeier: You have resolved this, but the line is still there.

ragna/core/_rag.py
```diff
@@ -188,7 +199,17 @@ async def prepare(self) -> Message:
...
-        await self._run(self.source_storage.store, self.documents)
+        from ragna.core import Document
+        if issubclass(self.source_storage.__ragna_input_type__, Document):
+            await self._run(self.source_storage.store, self.documents)
+        else:
+            # Here we need to generate the list of embeddings
+            chunks = []
+            for document in self.documents:
+                chunks += self.embedding_model._chunk_pages(document.extract_pages(), document_id=document.id, chunk_size=500, chunk_overlap=250)
+            embeddings = self.embedding_model.embed_chunks(chunks)
```
pmeier, line 210: I've commented this before: this logic cannot be here. It should be hardcoded in the embed method of MiniLML6v2. Of course will not stay there when we do the chunking refactor. But for now, we just want to preserve the status quo. `self.embedding_model._chunk_pages` is not something we can depend on here given that it is not part of the standard API of the `EmbeddingModel`

ragna/core/_components.py
```diff
@@ -76,6 +96,119 @@ def _protocol_model(cls) -> Type[pydantic.BaseModel]:
...
+                    ", ".join(map(str, range_int))
+                    if len(range_int) < 3
+                    else f"{range_int[0]}-{range_int[-1]}"
+                )
+                range_int = []
+
+        return ", ".join(ranges_str)
+
+    @classmethod
+    def _take_sources_up_to_max_tokens(
```
pmeier, line 189: Shouldn't this stay on the VDB base class? An `EmbeddingModel` has no touching points with the `Source`s.

ragna/source_storages/_chroma.py
pmeier: I'm confused. Could you please post the error? Not sure how Pydantic is involved here. The name of the variable should be arbitrary.

ragna/embedding_models/_embedding.py
```diff
@@ -0,0 +1,14 @@
...
+    def __init__(self):
+        super().__init__()
+        from chromadb.utils import embedding_functions
+        self.model = embedding_functions.DefaultEmbeddingFunction()
+
+    def embed_chunks(self, chunks: list[Chunk]) -> list[Embedding]:
+        return [Embedding(embed_chunk[0], embed_chunk[1]) for embed_chunk in zip(self.embed_text([chunk.text for chunk in chunks]), chunks)]
+
+    def embed_text(self, text: list[str]) -> list[float]:
+        return self.model(text)
```
pmeier, line 14: The input here is annotated as `list[str]` and it is passed in this way through `embed_chunks`. However in `Chat.answer` we just pass the string. One of them cannot be right. IMO we should just take `str` here and refactor the `embed_chunks` method.

ragna/embedding_models/_embedding.py
```diff
@@ -0,0 +1,14 @@
...
+
+
+class MiniLML6v2(EmbeddingModel):
+    def __init__(self):
+        super().__init__()
+        from chromadb.utils import embedding_functions
+        self.model = embedding_functions.DefaultEmbeddingFunction()
+
+    def embed_chunks(self, chunks: list[Chunk]) -> list[Embedding]:
+        return [Embedding(embed_chunk[0], embed_chunk[1]) for embed_chunk in zip(self.embed_text([chunk.text for chunk in chunks]), chunks)]
```
pmeier, line 11:
```suggestion
        return [Embedding(embedding=self.embed_text(chunk), chunk=chunk) for chunk in chunks]
```

ragna/embedding_models/_embedding.py
```diff
@@ -0,0 +1,14 @@
+from ragna.core import Chunk, Embedding, EmbeddingModel
+
+
+class MiniLML6v2(EmbeddingModel):
+    def __init__(self):
+        super().__init__()
+        from chromadb.utils import embedding_functions
```
pmeier, line 7: This needs to be listed in the `requirement` classmethod.

ragna/embedding_models/_embedding.py
```diff
@@ -0,0 +1,14 @@
...
+        self.model = embedding_functions.DefaultEmbeddingFunction()
```
pmeier, line 8: Let's make this private given that users do not need to interact with it.

```suggestion
        self._model = embedding_functions.DefaultEmbeddingFunction()
```

ragna/source_storages/_vector_database.py
```diff
@@ -61,78 +17,4 @@ def requirements(cls) -> list[Requirement]:
             # requirements. We use this as base for all of our Vector DBs.
             PackageRequirement("chromadb>=0.4.13"),
             PackageRequirement("tiktoken"),
```
pmeier, line 19: They shouldn't be needed anymore given that this now happens on the `EmbeddingModel`.

pmeier (commented):
@Tengal-Teemo I did another pass. Still mostly moving stuff around. I'll come to the meaty part after that. All open comments new and old still need your attention.

ragna/core/_rag.py
Tengal-Teemo: I see what you mean, I misunderstood.  I will fix it.

Tengal-Teemo:
@pmeier The error message from changing the chromadb retrieve function is as follows:
```
  File "/home/###/miniconda3/envs/ragna-dev/lib/python3.9/asyncio/runners.py", line 44, in run
    return loop.run_until_complete(main)
  File "/home/###/miniconda3/envs/ragna-dev/lib/python3.9/asyncio/base_events.py", line 647, in run_until_complete
    return future.result()
  File "/home/###/Work/ragna-dev-embedding/ui_entrypoint.py", line 28, in main
    async with rag.chat(
  File "/home/###/Work/ragna-dev-embedding/ragna/core/_rag.py", line 100, in chat
    return Chat(
  File "/home/###/Work/ragna-dev-embedding/ragna/core/_rag.py", line 178, in __init__
  File "/home/###/Work/ragna-dev-embedding/ragna/core/_rag.py", line 293, in _unpack_chat_params
    chat_params = ChatModel.model_validate(params, strict=True).model_dump(
  File "/home/###/miniconda3/envs/ragna-dev/lib/python3.9/site-packages/pydantic/main.py", line 503, in model_validate
    return cls.__pydantic_validator__.validate_python(
  File "/home/###/miniconda3/envs/ragna-dev/lib/python3.9/site-packages/pydantic/_internal/_mock_val_ser.py", line 41, in __getattr__
    val_ser = self._attempt_rebuild()
  File "/home/###/miniconda3/envs/ragna-dev/lib/python3.9/site-packages/pydantic/_internal/_mock_val_ser.py", line 73, in attempt_rebuild_validator
    if cls.model_rebuild(raise_errors=False, _parent_namespace_depth=5) is not False:
  File "/home/###/miniconda3/envs/ragna-dev/lib/python3.9/site-packages/pydantic/main.py", line 457, in model_rebuild
    _model_construction.unpack_lenient_weakvaluedict(cls.__pydantic_parent_namespace__) or {}
  File "/home/###/miniconda3/envs/ragna-dev/lib/python3.9/site-packages/pydantic/_internal/_model_construction.py", line 215, in __getattr__
    raise AttributeError(item)
AttributeError: __pydantic_parent_namespace__
python-BaseException
```
Apologies for my lack of experience with pydantic.

Tengal-Teemo:
Additionally @pmeier , I will finish up the rest of the comments tomorrow and attempt to solve why dropping commit [ed4f91a](https://github.com/Quansight/ragna/pull/369/commits/ed4f91af0e42d2f546319af661a7c256b1fd2d75) causes the following error:
```
  File "<frozen importlib._bootstrap>", line 1007, in _find_and_load
  File "<frozen importlib._bootstrap>", line 986, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 680, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 850, in exec_module
  File "<frozen importlib._bootstrap>", line 228, in _call_with_frames_removed
  File "/home/###/Work/ragna-dev-embedding/ragna/__init__.py", line 15, in <module>
    from . import assistants, core, deploy, source_storages
  File "/home/###/Work/ragna-dev-embedding/ragna/source_storages/__init__.py", line 7, in <module>
    from ._chroma import Chroma
  File "/home/###/Work/ragna-dev-embedding/ragna/source_storages/_chroma.py", line 13, in <module>
    class Chroma(VectorDatabaseSourceStorage):
  File "/home/###/miniconda3/envs/ragna-dev/lib/python3.9/abc.py", line 106, in __new__
    cls = super().__new__(mcls, name, bases, namespace, **kwargs)
  File "/home/###/Work/ragna-dev-embedding/ragna/core/_components.py", line 233, in __init_subclass__
    valid_input_types = get_args(get_type_hints(cls)["__ragna_input_type__"])
  File "/home/###/miniconda3/envs/ragna-dev/lib/python3.9/typing.py", line 1459, in get_type_hints
    value = _eval_type(value, base_globals, localns)
  File "/home/###/miniconda3/envs/ragna-dev/lib/python3.9/typing.py", line 292, in _eval_type
    return t._evaluate(globalns, localns, recursive_guard)
  File "/home/###/miniconda3/envs/ragna-dev/lib/python3.9/typing.py", line 554, in _evaluate
    eval(self.__forward_code__, globalns, localns),
  File "<string>", line 1, in <module>
NameError: name 'Union' is not defined
python-BaseException
```
I'm not 100% sure, but I believe it's caused by the introduction of the `__ragna_input_type__` variable somehow.

pmeier:
Re https://github.com/Quansight/ragna/pull/369#issuecomment-2019507390: This is on me. In contrast to what I said earlier, we do actually check for the name:

https://github.com/Quansight/ragna/blob/a5e4795532b4ad96256a695fbd45ff83df8356be/ragna/core/_components.py#L52-L70

Leave it as is. This is something I wanted to refactor anyway. I'll fix it in a follow-up PR.

pmeier:
Re https://github.com/Quansight/ragna/pull/369#issuecomment-2019516426: Fixed in 1d5d4a87152c7c44c6593c810155abedc3869499. Note that this also includes a few automatic formatting changes. They have to be applied before merge anyway, so I didn't mess with the pre-commit hooks for this.

ragna/source_storages/_vector_database.py
Tengal-Teemo: Yes this entire class can be removed now actually, if you don't mind.  It doesn't really have to wait until the chunking PR.

pmeier: Note that there is https://github.com/Quansight/ragna/pull/369#discussion_r1537222442. I don't mind making that a standalone function in a `_utils` module though.

ragna/embedding_models/_embedding.py
Tengal-Teemo: I would rather make it optional and handle the case of `str` being passed, because batching these should in theory make it much faster.

pmeier: Sorry for my ignorance here: are you saying that passing `list[str]` runs all items of that list in parallel? If so, I totally agree that we should do this. Can we assert this assumption though? A quick benchmark will do.

ragna/source_storages/_chroma.py
pmeier: Let's not for now in the light of https://github.com/Quansight/ragna/pull/369#issuecomment-2019756910.

ragna/source_storages/_lancedb.py
```diff
@@ -32,84 +39,81 @@ def __init__(self) -> None:
...
-        chunk_size: int = 500,
         num_tokens: int = 1024,
     ) -> list[Source]:
         table = self._db.open_table(str(chat_id))
 
         # We cannot retrieve source by a maximum number of tokens. Thus, we estimate how
         # many sources we have to query. We overestimate by a factor of two to avoid
         # retrieving to few sources and needed to query again.
-        limit = int(num_tokens * 2 / chunk_size)
+        limit = int(num_tokens * 2 / self._tokens * self._embeddings)
```
pmeier, line 105: IIUC, we only have these instance variables to compute the approximate number of sources that we want to pull out, right? If so we can probably remove this and put a hardcoded limit here:

1. This is a micro-optimization. Later we are taking the right number anyway so this just limits the number of rows we _initially_ query. Without evidence that this actually improves performance significantly, I don't think we should complicate our code for that.
2. The estimation does not really work for Chroma and we hardcode a limit already

    https://github.com/Quansight/ragna/blob/a5e4795532b4ad96256a695fbd45ff83df8356be/ragna/source_storages/_chroma.py#L88-L100
3. Most importantly, after the chunking rework, I'm not even sure we can support this. If the chunking is not based on tokens, and that likely will be true for all chunking models except the one we have right now, we might not even have the chunking information available.

Meaning, I'm ok with hardcoding `100` here same as for Chroma and revisit on the chunking work.

ragna/source_storages/_lancedb.py
```diff
@@ -32,84 +39,81 @@ def __init__(self) -> None:
...
+        _schema = pa.schema(
             [
                 pa.field("id", pa.string()),
                 pa.field("document_id", pa.string()),
                 pa.field("page_numbers", pa.string()),
                 pa.field("text", pa.string()),
                 pa.field(
                     self._VECTOR_COLUMN_NAME,
-                    pa.list_(pa.float32(), self._embedding_dimensions),
+                    pa.list_(pa.float32(), len(documents[0].embedding)),
```
pmeier, line 66: Since that might not be super obvious at first glance, can we create a variable for this above the block

```
dimensionality = len(documents[0].embedding))
_schema = ...
```

ragna/embedding_models/_all_minilm_l6_v2.py
```diff
@@ -0,0 +1,103 @@
...
+
+    def embed_documents(self, documents: list[Document]) -> list[Embedding]:
+        return [
+            Embedding(embedding=self.embed_text(chunk.text), chunk=chunk)
+            for document in documents
+            for chunk in self._chunk_pages(
+                document.extract_pages(),
+                document_id=document.id,
+                chunk_size=500,
+                chunk_overlap=250,
```
pmeier, line 67: These two need to become parameters of this function so users can still specify their values like they could before. Ultimately, we need to integrate this method into the Ragna protocol so that this can happen through the regular interface.

I can take that over when the PR is otherwise finished, because this will go deep into the Ragna internals. But LMK if you want me to guide you instead.

ragna/embedding_models/_all_minilm_l6_v2.py
```diff
@@ -0,0 +1,103 @@
...
+            )
+        ]
+
+    def embed_text(
```
pmeier, line 71: After looking at the PR in more detail, I again have my doubts about this function. As discussed before, this only exists for our benefit of embedding the prompt. However it comes at the cost of

1. We need to mess with the type annotations quite a bit, because the typing system doesn't really support output types based on the input, e.g. `str -> str` and `list[str] -> list[str]`.
2. Users have to implement two methods instead of one.
3. We need to integrate two very similar methods into the Ragna protocol.

I think the costs outweigh the benefit here.

Right now `embed_documents` takes `Document`s. But this will change with the chunking work to take `Chunk`s. If we move the chunking into `Chat.prepare` for now to emulate the future case for the embedding model, couldn't we create a private `_embed_text` method stuffs the text into a dummy `Chunk`, calls the user defined `embed_documents` and returns the embedding? That way we should be able to eliminate all the downsides above while keeping the benefit.

Thoughts?

ragna/core/_components.py
```diff
@@ -76,6 +92,22 @@ def _protocol_model(cls) -> Type[pydantic.BaseModel]:
         return merge_models(cls.display_name(), *cls._protocol_models().values())
 
 
+@dataclass
+class Embedding:
+    embedding: list[float]
```
pmeier, line 97: Should we call this

```suggestion
    values: list[float]
```

`embedding.embedding` always feels wrong.

ragna/core/_components.py
```diff
@@ -76,6 +92,22 @@ def _protocol_model(cls) -> Type[pydantic.BaseModel]:
...
+    chunk: Chunk
+
+
+class EmbeddingModel(Component, ABC):
```
pmeier, line 101: Open for opinions here. Chroma calls this `EmbeddingFunction` should we align to that? Are there other references that we should look at?

pmeier (commented):
Thanks Toby, this going pretty well. I took the liberty and pushed a few cleanup commits about some minor stuff that would just slow us down. Mostly moving the parts into their final position as well as linting. Now we can finally get to the interesting part.

ragna/source_storages/_lancedb.py
```diff
@@ -32,84 +39,81 @@ def __init__(self) -> None:
...
-                            )[0],
-                            "num_tokens": chunk.num_tokens,
-                        }
-                    ]
-                )
+        table = self._db.create_table(name=str(chat_id), schema=_schema)
+
+        for embedding in documents:
+            self._tokens += embedding.chunk.num_tokens
+            self._embeddings += 1
```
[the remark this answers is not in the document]
Tengal-Teemo: Ah I finally understand what you meant.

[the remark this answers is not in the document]
Tengal-Teemo: We can just remove this if we are hardcoding 100, so I will do that.

ragna/embedding_models/_all_minilm_l6_v2.py
Tengal-Teemo: I think, after reading a good analysis you are correct.  I will implement your solution.

Tengal-Teemo: The main thing swaying it for me is that the user (ragna developer) has to implement two different methods.  I think this warrants the roundabout method.

Tengal-Teemo: I'd also like to point out that pretty much no matter what you do, you'll still encounter the issue of passing through a batch vs passing through a single object.  We really want to be able to batch chunks because it is simply more efficient to batch-encode these things.  I will show you the difference after finishing up these comments.

Tengal-Teemo: You can take a look at what I've done in commit [8c9fb84](https://github.com/Quansight/ragna/pull/369/commits/8c9fb8436f79e8622eac51c06ad2a3003f4644a3), I don't particularly like it but it does work.

ragna/core/_components.py
Tengal-Teemo: EmbeddingFunction makes more sense if you're "calling" it on some text/chunks.  The user treats this as a model, so I think it should be named as such.  Although, now that we have embed_chunks as the only function of it, we could change it to be EmbeddingFunction as chromaDB does and make it callable.  The issue with that is that something that simple shouldn't require a chunk input, it should just be text.  I've made the change in a new branch [here](https://github.com/peachkeel/ragna/tree/callable-embedding-models) so you can have a look

ragna/embedding_models/_all_minilm_l6_v2.py
Tengal-Teemo: resolved as for the time being and also permanently and indefinitely this will happen outside of the `EmbeddingModel`

Tengal-Teemo:
@pmeier Sorry I've been away for a while, I believe I've caught up with your comments and made the suggested changes.  Let me know when you get time to do another passthrough.

Tengal-Teemo:
@pmeier Any chance of reviving this?  I'm happy to continue with it.

pmeier:
I'm really sorry this takes so long. I didn't expect to get so swamped the past few weeks. TBH, this PR is already in a fairly good state. Instead of going through yet another round of review, let me just finish it later today and merge. From there on, you can continue with the chunking PR if you like.

Tengal-Teemo:
@pmeier Ok no problem, I'm happy with that.

PR #376, opened 2024-03-23
[ENH] Add support for Ollama assistants

smokestacklightnin:
This PR adds support for Ollama assistants.


ragna/assistants/_ollama.py
```diff
@@ -0,0 +1,68 @@
...
+
+    def _make_system_content(self, sources: list[Source]) -> str:
+        instruction = (
+            "You are an helpful assistants that answers user questions given the context below. "
+            "If you don't know the answer, just say so. Don't try to make up an answer. "
+            "Only use the following sources to generate the answer."
+        )
+        return instruction + "\n\n".join(source.content for source in sources)
+
+    async def _call_api(
```
pmeier, line 24: We don't need this extra layer of abstraction here. Just use `async def answer`. `ApiAssistant._call_api` is an artifact from a previous implementation that wasn't cleaned up. We can do that when we get to #381. But let's not replicate this mistake here.

ragna/assistants/_ollama.py
```diff
@@ -0,0 +1,68 @@
...
+        self,
+        prompt: str,
+        sources: list[Source],
+        *,
+        max_new_tokens: int,
+        api_url: str = "http://localhost:11434/api/chat",
```
pmeier, line 30: This is not a runtime parameter, given that this URL doesn't change. Let's put it into `__init__`. We should also have a way to set this through an environment variable, e.g.

```python
def __init__(self, url="http://localhost:11434/api/chat"):
    self._url = os.environ.get("RAGNA_ASSISTANTS_OLLAMA_URL", url)
```

ragna/assistants/_ollama.py
```diff
@@ -0,0 +1,68 @@
...
+
+            async for chunk in response.aiter_lines():
+                # This part modeled after https://github.com/ollama/ollama/blob/06a1508bfe456e82ba053ea554264e140c5057b5/examples/python-loganalysis/readme.md?plain=1#L57-L62
+                if chunk:
+                    json_data = json.loads(chunk)
+
+                    if not json_data["done"]:
+                        yield cast(str, json_data["message"]["content"])
+                else:
+                    raise RagnaException("The response was empty.")
```
pmeier, line 64: In the example above, there is no error raised in case the chunk is empty. Maybe an empty chunk is a keep-alive chunk? Did you somehow confirm that this is an error case?

ragna/assistants/_ollama.py
```diff
@@ -0,0 +1,68 @@
...
+
+
+class Gemma2B(OllamaApiAssistant):
```
pmeier, line 67: Not sure about naming here. `vLLM` also supports Gemma. Meaning, without any namespacing, these two cannot coexist under `ragna.assistants`. Maybe we should starting to namespace the builtin assistants with the organization, e.g. `OllamaGemma`, `VllmGemma`, `OpenaiGpt4`, etc?

Let's open an issue for this. I'm ok with merging with the current name if we address the above before next release.

ragna/assistants/_api.py
```diff
@@ -11,12 +11,10 @@
...
 
 
-class ApiAssistant(Assistant):
-    _API_KEY_ENV_VAR: str
-
+class _ApiAssistant(Assistant):
     @classmethod
     def requirements(cls) -> list[Requirement]:
-        return [EnvVarRequirement(cls._API_KEY_ENV_VAR), *cls._extra_requirements()]
+        return []
```
pmeier, line 17: Should we make the accessibility of the API an availability check? Does the API have an endpoint that we could call to indicate whether the API is ready?

I would put that into 

```
@classmethod
def is_available():
```

similar to what we did [here](https://github.com/Quansight/ragna/blob/1d4cd568bdbd1335da39743743d5f2f48d05ca26/ragna/assistants/_airoboros.py#L19-L27).

ragna/assistants/_ollama.py
smokestacklightnin: I feel better changing it to `OllamaGemma2B` from `Gemma2B` and opening an issue.

smokestacklightnin: Discussed in Issue #384

ragna/assistants/_ollama.py
```diff
@@ -0,0 +1,112 @@
...
+        if not requirements_available:
+            return False
+
+        try:
+            response = httpx.get("http://localhost:11434/")
+            response.raise_for_status()
+        except httpx.HTTPError:
+            return False
+
+        return response.is_success
```
pmeier, line 39:
```suggestion
        try:
            return httpx.get("http://localhost:11434/").raise_for_status().is_success
        except httpx.HTTPError:
            return False
```

ragna/assistants/_ollama.py
```diff
@@ -0,0 +1,138 @@
...
+
+            async for chunk in response.aiter_lines():
+                # This part modeled after https://github.com/ollama/ollama/blob/06a1508bfe456e82ba053ea554264e140c5057b5/examples/python-loganalysis/readme.md?plain=1#L57-L62
+                if chunk:
+                    json_data = json.loads(chunk)
+
+                    if "error" in json_data:
+                        raise RagnaException(json_data["error"])
+                    if not json_data["done"]:
+                        yield cast(str, json_data["message"]["content"])
```
pmeier, line 96: @nenb This violates "same response schema" part of https://github.com/Quansight/ragna/pull/425#issuecomment-2134490726 :(

nenb (approved):
This looks fine to me.

I've pulled the branch locally and confirmed that it works for OpenAI models and for a Llamafile that I had locally.

Issue #375, opened 2024-03-22
ApiAssistant abstract base class for assistants that don't require an API key


smokestacklightnin:
### Feature description

Previously, we only had assistants hosted nonlocally. Now, adding Ollama, we will be supporting local assistants. The remote assistants all required API keys, but local ones like Ollama do not. That is really the only substantial difference; Ollama still is accessed via a REST API. Does it make sense to others to refactor `ragna.assistants._api.ApiAssistant` to not require an API key and then subclass it with the newly created subclass requiring the key?

An alternative is subclassing `ragna.core.Assistant` from "scratch" and not requiring an API key. This option seems suboptimal to me because all the behavior is otherwise the same as `ragna.assistants._api.ApiAssistant`.

Another alternative is to override the requirements to remove the API key requirement.

smokestacklightnin:
@pmeier Do you have an opinion?

pmeier:
> Now, adding Ollama, we will be supporting local assistants

To give some context here: this is not decided yet. We are exploring this an option.

> Does it make sense to others to refactor `ragna.assistants._api.ApiAssistant` to not require an API key and then subclass it with the newly created subclass requiring the key?

So far the new Ollama assistant is the only outlier. Thus, I wouldn't break our abstraction just yet, given that a workaround is available (see below). 

In case we encounter 1 or 2 more exceptions (maybe [vLLM](https://github.com/vllm-project/vllm)) to our current rule, I indeed agree it would make sense to refactor the class. I would move the API key functionality to a `class AuthorizedApiAssistant(ApiAssistant)` (name TBD) subclass.

> An alternative is subclassing `ragna.core.Assistant` from "scratch" and not requiring an API key. This option seems suboptimal to me because all the behavior is otherwise the same as `ragna.assistants._api.ApiAssistant`.
> 
> Another alternative is to override the requirements to remove the API key requirement.

Of those two, I would vote for the latter. That is a simple enough workaround that can easily be removed if we go for the more general solution explained above in the future.

PR #381, opened 2024-04-02
Refactor `ApiAssistant` to `AuthenticatedApiAssistant` and `UnauthenticatedApiAssistant`

smokestacklightnin:
In preparation for adding local assistants that do not require API keys, this PR refactors `ragna.assistants._api.ApiAssistant`. 

This resolves Issue #375. 

If anyone has strong opinions about the chosen naming conventions, please share them. 

IMPORTANT: This PR should be merged after PR #380 is merged and before PR #376 is merged.

### DIFF
--- a/ragna/assistants/_api.py
+++ b/ragna/assistants/_api.py
@@ -1,7 +1,6 @@
 import abc
 import contextlib
 import json
-import os
 from typing import AsyncIterator
 
 import httpx
@@ -12,11 +11,9 @@ from ragna.core import Assistant, EnvVarRequirement, RagnaException, Requirement
 
 
 class ApiAssistant(Assistant):
-    _API_KEY_ENV_VAR: str
-
     @classmethod
     def requirements(cls) -> list[Requirement]:
-        return [EnvVarRequirement(cls._API_KEY_ENV_VAR), *cls._extra_requirements()]
+        return []
 
     @classmethod
     def _extra_requirements(cls) -> list[Requirement]:
@@ -27,7 +24,6 @@ class ApiAssistant(Assistant):
             headers={"User-Agent": f"{ragna.__version__}/{self}"},
             timeout=60,
         )
-        self._api_key = os.environ[self._API_KEY_ENV_VAR]
 
     async def answer(
         self, prompt: str, sources: list[Source], *, max_new_tokens: int = 256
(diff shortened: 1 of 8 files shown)


ragna/assistants/_api.py
```diff
@@ -58,3 +55,20 @@ async def _assert_api_call_is_success(self, response: Response) -> None:
...
+    @classmethod
+    def requirements(cls) -> list[Requirement]:
+        return [EnvVarRequirement(cls._API_KEY_ENV_VAR), *cls._extra_requirements()]
+
+    def __init__(self) -> None:
+        super().__init__()
+        self._api_key = os.environ[self._API_KEY_ENV_VAR]
+
+
+class UnauthenticatedApiAssistant(_ApiAssistant):
```
pmeier, line 72: This class is not needed. We can simply use the superclass for this.

ragna/assistants/_api.py
```diff
@@ -11,12 +11,10 @@
...
-    _API_KEY_ENV_VAR: str
-
+class _ApiAssistant(Assistant):
     @classmethod
     def requirements(cls) -> list[Requirement]:
-        return [EnvVarRequirement(cls._API_KEY_ENV_VAR), *cls._extra_requirements()]
+        return []
 
     @classmethod
     def _extra_requirements(cls) -> list[Requirement]:
```
pmeier, line 20: Not sure if it makes sense to have this on the superclass if for now only the authenticated subclass needs it.

ragna/assistants/_api.py
```diff
@@ -11,12 +11,10 @@
...
 
 
-class ApiAssistant(Assistant):
-    _API_KEY_ENV_VAR: str
-
+class _ApiAssistant(Assistant):
     @classmethod
     def requirements(cls) -> list[Requirement]:
-        return [EnvVarRequirement(cls._API_KEY_ENV_VAR), *cls._extra_requirements()]
+        return []
```
pmeier, line 17: This is the default return if we don't override the method. Meaning, we can just leave it out here.

tests/assistants/test_api.py
```diff
@@ -4,16 +4,22 @@
...
 API_ASSISTANTS = [
     assistant
     for assistant in assistants.__dict__.values()
     if isinstance(assistant, type)
-    and issubclass(assistant, ApiAssistant)
-    and assistant is not ApiAssistant
+    and issubclass(assistant, _ApiAssistant)
+    and assistant is not _ApiAssistant
+    and assistant is not AuthenticatedApiAssistant
+    and assistant is not UnauthenticatedApiAssistant
```
pmeier, line 22:
```suggestion
    and issubclass(assistant, AuthenticatedApiAssistant)
    and assistant is not AuthenticatedApiAssistant
```

pmeier (commented):
We misunderstood each other when spoke offline yesterday. What I meant is that we should simply not subclass the `OllamaAssistant` from the `ApiAssistant` for now. My original reasoning from #375 holds. I don't want to introduce another abstraction just for a single assistant. We can revisit this later. Until then, I'll close this PR.

Issue #385, opened 2024-04-08
MosaicML API is either down or removed


pmeier:
First test failure observed here: https://github.com/Quansight/ragna/actions/runs/8594065268/job/23546541544?pr=380

I cannot find any documentation about their API anymore. Maybe it was removed? We need to look into this and potentially remove the assistants.

PR #387, opened 2024-04-12
Remove Mosaic Assistant

smokestacklightnin:
This PR addresses Issue #385. 

Mosaic ML seems to have removed or deprecated their REST API. As a result, including this assistant was causing tests to fail (see the attached issue).

It is not necessary to support deprecated APIs, so this PR removes it from Ragna.


smokestacklightnin:
The typing linter [seems to fail](https://github.com/Quansight/ragna/actions/runs/8657272637/job/23739176767?pr=387) on GitHub but passes locally for me. I am skeptical that the failure has anything to do with the changes in this PR

pmeier:
> The typing linter [seems to fail](https://github.com/Quansight/ragna/actions/runs/8657272637/job/23739176767?pr=387) on GitHub but passes locally for me. I am skeptical that the failure has anything to do with the changes in this PR

It has not. See #390.

I'm waiting for one other confirmation that the API is indeed gone. If I have that, the PR is good to go.

pmeier (approved):

PR #396, opened 2024-04-27
Add instructions to install from conda-forge.

kklein:
ragna has recently been brought to conda-forge; see [this feestock](https://github.com/conda-forge/ragna-feedstock) and the [package information](https://anaconda.org/conda-forge/ragna).

This PR
- addresses a TODO in the root's readme to add installation instructions via conda
- adds a button to access find the package on conda-forge, imitating the PyPI button

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -28,7 +28,7 @@ And, upgrade to latest versions with:
 pip install --upgrade ragna
 ```
 
-Alternatively, you can install ``ragna`` with conda from conda-forge:
+Alternatively, you can install `ragna` with conda from conda-forge:
 
 ```bash
 conda install ragna -c conda-forge
--- a/docs/install.md
+++ b/docs/install.md
@@ -28,7 +28,11 @@ And, upgrade to latest versions with:
 pip install --upgrade ragna
 ```
 
-<!-- Add conda and conda-forge if/when available -->
+Alternatively, you can install ``ragna`` with conda from conda-forge:
+
+```bash
+conda install ragna -c conda-forge
+```
 
 ## Verify your installation
 
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@ A RAG orchestration framework.
          alt="PyPI" />
         </a>
 		<a href="https://anaconda.org/conda-forge/ragna">
-			<img srg="https://img.shields.io/conda/vn/conda-forge/ragna?logoColor=white&logo=conda-forge" alt="conda-forge">
+			<img src="https://img.shields.io/conda/vn/conda-forge/ragna?logoColor=white&logo=conda-forge" alt="conda-forge">
 		</a>
     </td>
     <td>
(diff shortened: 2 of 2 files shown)


kklein:
Hi @pmeier ! I attended your workshop at PyData Berlin last week - thank you!
I wanted to get started with ragna and realized it wasn't on conda-forge yet. Hence I brought it to conda-forge. This is a PR to add some information on ragna's availability on conda-forge.

pmeier (approved):
Ah nice, thanks Kevin! We had a try earlier at this (conda-forge/staged-recipes#24706), but hit some blockers.

One thing that I'm wondering though: our recommendation is to `pip install ragna[all]` and only go for `pip install ragna` if one wants fine grained control. For conda this is "flipped" given that there are no optional dependencies. Meaning, users will always install the minimal version and have no way to automatically install the one with batteries included.

That sucks from a UX perspective. Let me ask around if there is a good solution to the problem. I'm sure we are not the first ones to hit this.

docs/install.md
```diff
@@ -28,7 +28,11 @@ And, upgrade to latest versions with:
 pip install --upgrade ragna
 ```
 
-<!-- Add conda and conda-forge if/when available -->
+Alternatively, you can install ``ragna`` with conda from conda-forge:
```
pmeier, line 31: MD not RST :stuck_out_tongue: 

```suggestion
Alternatively, you can install `ragna` with conda from conda-forge:
```

kklein:
@pmeier thanks for your response!

> That sucks from a UX perspective. Let me ask around if there is a good solution to the problem. I'm sure we are not the first ones to hit this.

I understand. Iirc the way this is usually done is to have two outputs in the conda recipe. For instance, the [matplotlib recipe](https://github.com/conda-forge/matplotlib-feedstock/blob/main/recipe/meta.yaml#L17) has one output called `matplolib`, which includes all dependencies as well as `matplotlib-base`, which includes a barebones set of dependencies.

Should I look into setting this up, tag you in the feedstock PR, and create a follow-up ragna PR once this is settled?

pmeier:
> Iirc the way this is usually done is to have two outputs in the conda recipe. For instance, the [matplotlib recipe](https://github.com/conda-forge/matplotlib-feedstock/blob/main/recipe/meta.yaml#L17) has one output called `matplolib`, which includes all dependencies as well as `matplotlib-base`, which includes a barebones set of dependencies.

That is what I got from my sources as well.

> Should I look into setting this up, tag you in the feedstock PR, and create a follow-up ragna PR once this is settled?

Yes, that would be awesome. TBH, we should probably have a discussion about the same for our PyPI releases. Not only will it be awkward if conda and pip installs diverge, but also `pip install ragna` should give you what you need without looking up installation instructions.

pmeier:
> Should I look into setting this up, tag you in the feedstock PR, and create a follow-up ragna PR once this is settled?

Maybe hold off on this one. I'll start an issue soon on how we want to do this. I got some conflicting opinions from people that know conda / conda-forge a lot better than me.

PR #401, opened 2024-05-03
Fix #399 : eager loading of Chats docs, msgs, srcs

### LINKED ISSUE #399: Eager loading of ORM relationships
### Feature description

@nenb reported offline that loading chats through the API is really slow. He tracked it down to

https://github.com/Quansight/ragna/blob/04168aba0e151e431b4a5a359e9c8eed475d4894/ragna/deploy/_api/database.py#L101

This function looks really innocent at first glance. The slow part is the attribute access on the ORM relationships. In this example these are

https://github.com/Quansight/ragna/blob/04168aba0e151e431b4a5a359e9c8eed475d4894/ragna/deploy/_api/database.py#L104

https://github.com/Quansight/ragna/blob/04168aba0e151e431b4a5a359e9c8eed475d4894/ragna/deploy/_api/database.py#L119

https://github.com/Quansight/ragna/blob/04168aba0e151e431b4a5a359e9c8eed475d4894/ragna/deploy/_api/database.py#L123

Under the hood, SQLAlchemy is performing a separate `SELECT` for each individual item. This is not acceptable.

Instead, we should eagerly load these when loading the chat. SQLAlchemy has [multiple strategies](https://docs.sqlalchemy.org/en/20/orm/queryguide/relationships.html) to do that. I think at the scales that we are currently talking about, `lazy="joined"` is probably best, because it creates the least amounts of queries. But I'm ok with either if reasoning is explained.

This shouldn't be limited to the chats, but should be expanded to all of our ORM objects. Since we always we have the ORM -> Schema conversion on the way out of the database layer, we will always access all attributes anyway.

### Value and/or benefit

Way faster loading times for Ragna deployments at scale.

### Anything else?

I've tried to naively add `lazy="joined"` to all relationships. Running the test suite with this, got me to an error that our newly added JSON type for some reason got a `None` input here:

https://github.com/Quansight/ragna/blob/04168aba0e151e431b4a5a359e9c8eed475d4894/ragna/deploy/_api/orm.py#L25-L30

I have no idea how this happened, given that this should only ever happen on the way out of the DB and thus there should be data. Maybe this happens because we are trying to eagerly load objects that have the JSON datatype, but simply don't have data yet. For example, eagerly loading a `orm.Document` also loads an "associated" `orm.Chat`, but when starting from a fresh state, we might not have a single row in the chat table yet. Anyway, adding

```python
if value is None:
     return None
```

sidestepped this issue. Afterwards, I was hit with another issue that is explained in [this SO thread](https://stackoverflow.com/questions/74179020/the-unique-method-must-be-invoked-on-this-result-exception-raised-after-sqlalc). I've tried to add `.unique()` to the selects, but the disambiguitation did not work properly.

This is as far as I have looked before writing up this issue.

pierrotsmnrd:
This is an attempt at solving issue #399 

Using `lazy="joined"` on all the relationships was one way of doing it. I think we should use the alternative `joinedload` for more flexibility. It's kind of the same but we have the control over when we want to use it. 

I'd like your feedback on the performances in the various contexts you're using it. I've tested it locally, adding `echo=True` to `create_engine`. I could see that before this fix, there was a query to retrieve all chats, and then queries to get documents, messages and sources for each chat. After the fix, I see only one query to them all.

Unit tests are passing locally, except 2 but they seem related to my local setup

### DIFF
--- a/ragna/deploy/_api/database.py
+++ b/ragna/deploy/_api/database.py
@@ -140,7 +140,7 @@ def get_chats(session: Session, *, user: str) -> list[schemas.Chat]:
     return [
         _orm_to_schema_chat(chat)
         for chat in session.execute(
-            select(orm.Chat)
+            select(orm.Chat)  # type: ignore[attr-defined]
             .options(
                 joinedload(orm.Chat.messages).joinedload(orm.Message.sources),
                 joinedload(orm.Chat.documents),
--- a/ragna/deploy/_api/database.py
+++ b/ragna/deploy/_api/database.py
@@ -6,7 +6,7 @@ from typing import Any, Callable, Optional, cast
 from urllib.parse import urlsplit
 
 from sqlalchemy import create_engine, select
-from sqlalchemy.orm import Session
+from sqlalchemy.orm import Session, joinedload
 from sqlalchemy.orm import sessionmaker as _sessionmaker
 
 from ragna.core import RagnaException
@@ -140,9 +140,15 @@ def get_chats(session: Session, *, user: str) -> list[schemas.Chat]:
     return [
         _orm_to_schema_chat(chat)
         for chat in session.execute(
-            select(orm.Chat).where(orm.Chat.user_id == _get_user_id(session, user))
+            select(orm.Chat)
+            .options(
+                joinedload(orm.Chat.messages).joinedload(orm.Message.sources),
+                joinedload(orm.Chat.documents),
+            )
+            .where(orm.Chat.user_id == _get_user_id(session, user))
         )
         .scalars()
+        .unique()
         .all()
     ]
 
--- a/ragna/deploy/_api/database.py
+++ b/ragna/deploy/_api/database.py
@@ -136,16 +136,23 @@ def _orm_to_schema_chat(chat: orm.Chat) -> schemas.Chat:
     )
 
 
+def _select_chat(*, eager: bool = False) -> Any:
+    selector = select(orm.Chat)
+    if eager:
+        selector = selector.options(  # type: ignore[attr-defined]
+            joinedload(orm.Chat.messages).joinedload(orm.Message.sources),
+            joinedload(orm.Chat.documents),
+        )
+    return selector
+
+
 def get_chats(session: Session, *, user: str) -> list[schemas.Chat]:
     return [
         _orm_to_schema_chat(chat)
         for chat in session.execute(
-            select(orm.Chat)  # type: ignore[attr-defined]
-            .options(
-                joinedload(orm.Chat.messages).joinedload(orm.Message.sources),
-                joinedload(orm.Chat.documents),
+            _select_chat(eager=True).where(
+                orm.Chat.user_id == _get_user_id(session, user)
             )
-            .where(orm.Chat.user_id == _get_user_id(session, user))
         )
         .scalars()
         .unique()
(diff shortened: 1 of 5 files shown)


ragna/deploy/_api/database.py
```diff
@@ -140,9 +140,15 @@ def get_chats(session: Session, *, user: str) -> list[schemas.Chat]:
     return [
         _orm_to_schema_chat(chat)
         for chat in session.execute(
-            select(orm.Chat).where(orm.Chat.user_id == _get_user_id(session, user))
+            select(orm.Chat)  # type: ignore[attr-defined]
+            .options(
+                joinedload(orm.Chat.messages).joinedload(orm.Message.sources),
+                joinedload(orm.Chat.documents),
```
pmeier, line 146: Why are the first two chained together, i.e. separated by a dot `.`, while the latter is another input argument, i.e. separated by a comma `,`?

pmeier (commented):
Thanks @pierrotsmnrd this looks like what I wanted to do yesterday, but I guess my brain was too fried to get there. Some minor comments below. Other than them, we need to wait for @nenb to test and confirm that this indeed solves the performance issue. But I'm confident about that.

---

We need the same eager behavior for `get_chat` (mind the missing `s` at the end) below. Since that calls `_get_orm_chat` internally, I suggest we add a `eager=False` parameter to it and set `eager=True` in `get_chat`.

ragna/deploy/_api/database.py
pierrotsmnrd: Each first-level `joinedload` (comma-separated) makes joins on Message and Document from Chat.
The second-level `joinedload` (dot separated) makes a join from Message to Source.

That is because we have : 
```
1 Chat ➞ N Documents
1 Chat ➞ N Messages
        1 Message ➞ N Sources
```

pmeier: Got it, thanks!

pmeier:
@pierrotsmnrd I've fixed my own comment. However, I'm now hitting an issue with `.unique()` where the disambiguiation doesn't work. Could you have a look at the test failures?

nenb:
@pmeier I noticed in the tests (and probably in the code) that we are relying on the `Chat.messages` to be ordered by `message.timestamp`. I've added this to the ORM relationship, and this has all but one (Windows) of the test-suites passing.

The issue with the Windows test-suite is that it is producing identical timestamps for the `User` and `Assistant` messages (when clearly this should not be the case, as they are created separately in the code). The ordering occasionally fails in this case which is not desirable (eg an `Assistant` message will be returned before a `User` message in the UI, even though the `User` asked the initial question). I have tried to introduce extra ordering on the `Enum`, but wasn't able to get this to work without larger changes to the code.

I have checked a database where I am using `ragna` on a Windows machine, and I can find no evidence of this identical timestamp issue, which is somewhat reassuring. I honestly have no idea why it's happening though - it's an extremely frustrating bug! 

I'm going to leave this PR for now. Let me know if you are able to determine what is going on in Windows.

pmeier:
Somewhat good news: I can reproduce the timestamp issue on `main` (see #403). Meaning, this PR is not the culprit and we can go ahead and merge. Still, we need to figure out what is happening here.

pmeier:
@nenb TL;DR The issue is solved by adding a small `time.sleep` into `.answer()` method. 

For some reason, `utcnow()` on Windows seem to be cached if called multiple times in rapid succession. Or we are hitting a resolution issue.

Here are some debug logs from #403 on Ubuntu:

```
COUNT=0, timestamp=datetime.datetime(2024, 5, 6, 10, 59, 51, 646422), timestamp2=datetime.datetime(2024, 5, 6, 10, 59, 51, 646424, tzinfo=datetime.timezone.utc)
COUNT=1, timestamp=datetime.datetime(2024, 5, 6, 10, 59, 51, 659666), timestamp2=datetime.datetime(2024, 5, 6, 10, 59, 51, 659668, tzinfo=datetime.timezone.utc)
COUNT=2, timestamp=datetime.datetime(2024, 5, 6, 10, 59, 51, 661967), timestamp2=datetime.datetime(2024, 5, 6, 10, 59, 51, 661969, tzinfo=datetime.timezone.utc)
```

The `COUNT` indicates a new call to making a timestamp when creating a new message. Each timestamp is different even if its just by single digit microseconds. Here is the same print on Windows

```
COUNT=0, timestamp=datetime.datetime(2024, 5, 6, 11, 2, 17, 762144), timestamp2=datetime.datetime(2024, 5, 6, 11, 2, 17, 762144, tzinfo=datetime.timezone.utc)
COUNT=1, timestamp=datetime.datetime(2024, 5, 6, 11, 2, 17, 793388), timestamp2=datetime.datetime(2024, 5, 6, 11, 2, 17, 793388, tzinfo=datetime.timezone.utc)
COUNT=2, timestamp=datetime.datetime(2024, 5, 6, 11, 2, 17, 793388), timestamp2=datetime.datetime(2024, 5, 6, 11, 2, 17, 793388, tzinfo=datetime.timezone.utc)
```

Two requests back to back, i.e. `timestamp=` and `timestamp2=` will always have the same timestamp. And it seems we are hitting the same issue again for `COUNT=1` and `COUNT=2`.

The reason you are not seeing this during the actual operation is that the time it takes for a real assistant to compute the answer is long enough to give us a new timestamp and not trigger the behavior above. Meaning, if we simulate a "real" assistant in our tests by adding a small delay, we are good.

pmeier (approved):
Thanks Pierre and Nick! @nenb I'll leave it up to you to merge when tested on your side.

PR #402, opened 2024-05-03
mypy for sqlalchemy

pmeier:
While looking into the lint issues of #401, I realized we never activated the SQLAlchemy mypy plugin. This PR makes up for that.

This surfaces some errors. When columns are defined without `nullable=False`, the value of the field is always `Optional[T]`. Since we don't have `None` values and won't have them for the currently available columns in the future, I've set `nullable=False` everywhere.

### DIFF
--- a/ragna/deploy/_api/database.py
+++ b/ragna/deploy/_api/database.py
@@ -211,7 +211,7 @@ def update_chat(session: Session, user: str, chat: schemas.Chat) -> None:
     orm_chat = _get_orm_chat(session, user=user, id=chat.id)
 
     orm_chat.prepared = chat.prepared
-    orm_chat.messages = [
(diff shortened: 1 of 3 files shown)


ragna/deploy/_api/database.py
```diff
@@ -211,7 +211,7 @@ def update_chat(session: Session, user: str, chat: schemas.Chat) -> None:
     orm_chat = _get_orm_chat(session, user=user, id=chat.id)
 
     orm_chat.prepared = chat.prepared
-    orm_chat.messages = [
+    orm_chat.messages = [  # type: ignore[assignment]
```
pmeier, line 214: mypy incorrectly infers `orm.Chat.messages` to be of type `orm.Message` rather than the correct `list[orm.Message]`. Thus, we just silence it here.

PR #404, opened 2024-05-06
Add endpoint for batch uploading document metadata

nenb:
Add multiple documents to the database in a single operation (`add_all`). This is important when there is non-negligible latency associated with each call to the database.


pmeier (approved):
Thanks Nick! Could you open an issue for the TODO so we don't forget?

PR #411, opened 2024-05-12
Fix CopyToCliboard functionality in UI

### LINKED ISSUE #408: CopytoClipboardButton does not work when streaming
### Bug description

When I try to copy a response to the clipboard when streaming, at most 1 letter seems to be selected.

It appears to work correctly when using the demo assistants (where there is no streaming).

### How to reproduce the bug?

Select an assistant that involves streaming and try to copy the response with the button.

_No response_

### Anything else?

The logic is contained in `central_view.py`. A ReactiveHTML component is created for `CopyToClipboardButton`. I think that when we are streaming, we are not passing the full value to this component.

nenb:
Fixes #408 .

### DIFF
--- a/ragna/deploy/_ui/central_view.py
+++ b/ragna/deploy/_ui/central_view.py
@@ -110,7 +110,7 @@ class RagnaChatMessage(pn.chat.ChatMessage):
         on_click_source_info_callback: Optional[Callable] = None,
         timestamp=None,
         show_timestamp=True,
-        toolbar_visible=True,
+        toolbar_visible=True,  # hide the toolbar during streaming
     ):
         css_class = f"message-content-{self.role}"
         self.content_pane = pn.pane.Markdown(
@@ -124,6 +124,7 @@ class RagnaChatMessage(pn.chat.ChatMessage):
             ),
         )
 
+        # we make this available on the instance so that we can update the value later
         self.clipboard_button = CopyToClipboardButton(
             value=self.content_pane.object,
             title="Copy",
@@ -132,6 +133,7 @@ class RagnaChatMessage(pn.chat.ChatMessage):
             ],
         )
 
+        # we make this available on the instance so that we can toggle the visibility
         self.toolbar = pn.Row(
             self.clipboard_button,
             pn.widgets.Button(
--- a/ragna/deploy/_ui/central_view.py
+++ b/ragna/deploy/_ui/central_view.py
@@ -67,9 +67,6 @@ class RagnaChatMessage(pn.chat.ChatMessage):
         self.clipboard_button = CopyToClipboardButton(
             value=self.content_pane.object,
             title="Copy",
-            stylesheets=[
-                ui.CHAT_INTERFACE_CUSTOM_BUTTON,
-            ],
         )
 
         # we make this available on the instance so that we can toggle the visibility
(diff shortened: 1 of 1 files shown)


ragna/deploy/_ui/central_view.py
```diff
@@ -123,12 +124,37 @@ def __init__(
...
+        self.clipboard_button = CopyToClipboardButton(
+            value=self.content_pane.object,
+            title="Copy",
+            stylesheets=[
+                ui.CHAT_INTERFACE_CUSTOM_BUTTON,
+            ],
+        )
+
+        # we make this available on the instance so that we can toggle the visibility
+        self.toolbar = pn.Row(
```
pmeier, line 137: I'm not against the toolbar name, but should we make it more obvious that this only applies to assistant messages?

ragna/deploy/_ui/central_view.py
```diff
@@ -110,6 +110,7 @@ def __init__(
         on_click_source_info_callback: Optional[Callable] = None,
         timestamp=None,
         show_timestamp=True,
+        toolbar_visible=True,  # hide the toolbar during streaming
```
pmeier, line 113: Should we make this

```suggestion
        toolbar_visible: Optional[bool] = None,  # hide the toolbar during streaming
```

and enforce a `None` value for all roles except `"assistant"`? I mean this is not user facing so not blocking in any way.

ragna/deploy/_ui/central_view.py
```diff
@@ -138,15 +164,17 @@ def __init__(
...
-                    {"background-color": "rgb(243, 243, 243) !important"}
-                    if role == "user"
-                    else {
-                        "background-color": "none",
-                        "border": "rgb(234, 234, 234)",
-                        "border-style": "solid",
-                        "border-width": "1.2px",
-                        "border-radius": "5px",
-                    },
+                    (
```
pmeier, line 167: This will create merge conflicts with #312 that we'll merge soon. But I think this is only formatting anyway so you can just accept whatever we might have upstream.

ragna/deploy/_ui/central_view.py
```diff
@@ -389,11 +396,16 @@ async def chat_callback(
...
                 sources=answer["sources"],
                 on_click_source_info_callback=self.on_click_source_info_wrapper,
+                toolbar_visible=False,
             )
             yield message
 
+            content = answer["content"]
             async for chunk in answer_stream:
                 message.content_pane.object += chunk["content"]
+                content += chunk["content"]
```
pmeier, line 406: Instead doing a second repeated string concatenation here, can't we just do `message.clipboard_button.value = message.content_pane.object` after the loop?

pmeier (approved):
Few minor comments, but otherwise LGTM! Thanks Nick.

ragna/deploy/_ui/central_view.py
pmeier: I went ahead and fixed them.

ragna/deploy/_ui/central_view.py
nenb: Renamed as `assistant_toolbar`.

ragna/deploy/_ui/central_view.py
nenb: The way the current code is designed, the `toolbar` is only used by the `assistant` role. I'll make this more obvious by renaming as `assistant_toolbar_visible`.

PR #413, opened 2024-05-14
UI tests with Playwright

blakerosenthal:
Added some tests for the UI with [Playwright](https://playwright.dev/python/)! Here's a couple decisions I made that could use another gut check or two:

- I chose to spin up an actual uvicorn server for the API rather than mock those endpoints. I also chose not to mock the UI endpoints for the sake of actually testing browser integration.
- I set different ports for the API and the UI so as to not conflict with other servers that might be running on the same machine. 

Questions:

- What's happening with my auth setup? Setting a token either in the browser header or the panel cookies/headers doesn't seem to bypass the main auth page.
- Where should I look for better CSS locators? Playwright codegen offers some wacky identifiers. @pierrotsmnrd maybe you could point me to the right files for this?
- Will this work with GH Actions? Who knows; we'll find out!


tests/deploy/ui/test_ui.py
```diff
@@ -0,0 +1,183 @@
...
+    auth = auth_header(config.api.url)
+    assert len(auth) > len("Bearer ")
+
+    page.set_extra_http_headers({"Authorization": auth})  # this doesn't work
+    pn.state.cookies["auth_token"] = ""  # or this
+    pn.state.headers["Authorization"] = auth  # or this
+
+    index_url = config.ui.origins[0]
+    page.goto(index_url)
+    expect(page.get_by_role("button", name=" New Chat")).to_be_visible()
```
pmeier, line 156: I don't understand this test. Why are you trying to do here? The `auth_token` is a cookie that has to be set by the UI. What you can and should test is that after the login you now have the respective cookie set.

Ultimately you can use this cookie to access the API with bearer authorization and check if the information in the UI is correct. For example, when getting an answer, check with the API whether the displayed sources are actually the ones we also have in the DB.

tests/deploy/ui/test_ui.py
```diff
@@ -0,0 +1,183 @@
...
+from playwright.sync_api import expect, sync_playwright
+
+from ragna.assistants import RagnaDemoAssistant
+from ragna.core._utils import default_user
+from ragna.deploy import ApiConfig, Config, UiConfig
+from ragna.deploy._api import app as api_app
+from ragna.deploy._ui import app as ui_app
+
+TEST_API_PORT = "38769"
+TEST_UI_PORT = "38770"
```
pmeier, line 18: Instead of hardcoding them, can we just make them fully random? See 

https://github.com/Quansight/ragna/blob/dfb434812ad4931f54181e69e7cf04c99351d629/tests/utils.py#L79-L82

(note that this is an old commit and we removed this in #322)

pmeier (commented):
Thanks Blake for the PR. One thing that might simplify the setup quite a bit is the usage of our CLI entrypoint directly. Meaning, instead of starting the API and UI yourself, just run 

https://github.com/Quansight/ragna/blob/0cb205f24ca9fb322ac32a2e1aaec6b34928cd2e/ragna/deploy/_cli/core.py#L100-L101

directly. I'm not 100% sure this works, but it should.

blakerosenthal:
> ... just run
> 
> https://github.com/Quansight/ragna/blob/0cb205f24ca9fb322ac32a2e1aaec6b34928cd2e/ragna/deploy/_cli/core.py#L100-L101
> 
> directly. I'm not 100% sure this works, but it should.

I will try this and see what happens. I think I initially thought you couldn't snag a typer function like that but maybe I'm wrong.

environment-dev.yml
```diff
@@ -10,6 +10,8 @@ dependencies:
       - pytest >=6
       - pytest-mock
       - pytest-asyncio
+      - pytest-playwright
+      - playwright
```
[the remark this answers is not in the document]
blakerosenthal: Good catch 👍

tests/deploy/ui/test_ui.py
```diff
@@ -0,0 +1,183 @@
...
+
+    def start(self):
+        self.proc = Process(target=self.start_server, args=(), daemon=True)
+        self.proc.start()
+
+        timeout = 5
+        while timeout > 0 and not self.server_up():
+            print(f"Waiting for API server to come up on {self.config.api.url}")
+            time.sleep(1)
+            timeout -= 1
```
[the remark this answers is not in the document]
blakerosenthal: Ah, so much better

tests/deploy/ui/test_ui.py
```diff
@@ -0,0 +1,183 @@
...
+        .raise_for_status()
+        .json()
+    )
+
+    return f"Bearer {token}"
+
+
+@pytest.fixture(scope="function")
+def page(config, api_server, headed_mode):
+    server = ui_app(config=config, open_browser=False)
```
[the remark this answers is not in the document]
blakerosenthal: This was one of my questions I forgot to include.

When I had the panel `.serve()` call outside of the `sync_playwright()` scope it would just hang forever. I couldn't figure out why so it ended up in this block.

Initially both the UI server and the API server were just plain functions with no class wrapper, but both MacOS and Windows were struggling to pickle the closures during the pytest setup so I moved the API server to its own class. I will first try to use the CLI func for the UI and fallback to making individual classes for both the UI and API for consistency.

tests/deploy/ui/test_ui.py
blakerosenthal: Oh I see, I was trying to test if having the auth token already in the browser would bypass the login page and route directly to the main index.

Checking whether the displayed sources match the DB is a great test idea and I will add that.

blakerosenthal:
@pmeier (cc @pierrotsmnrd)
Running into a bit of a snag running the two servers in the test env. The tests running in CI are experiencing timeouts waiting for the server to come up, but locally there are some different errors with the various streams getting closed too early. I must not be managing the various sub processes correctly: [pytest_output.txt](https://github.com/Quansight/ragna/files/15406697/pytest_output.txt)

Then, separate issue, clicking on the "start chat" button gives me a pydantic validation error that only shows up in playwright (line 182).

I wasn't seeing the issues with the server in [previous versions of this PR](https://github.com/Quansight/ragna/actions/runs/9085532364/job/24969085758) where I spun up each server separately.

Final thought: given that we're expecting the bokeh/fast-api integration soon-ish, does it make sense to punt on some of these issues until that is merged? At least a couple of these roadblocks might solve themselves.

pmeier:
> the tests running in CI are experiencing timeouts waiting for the server to come up

You've set the timeout to only 5 seconds, which is a little to low. A timeout is there to prevent the test suite to run indefinitely when something goes wrong. Since you have a proper check on when the server is started, I suggest you increase the timeout to something like 30 seconds. Recall that CI machines are small so it might take them 2-3 times to get this server up compared to your local setup.

> but locally there are some different errors with the various streams getting closed too early. 

I've seen these error before when using Chrome / Chromium. That seems to be a bug in `panel` or `bokeh`. However, I've never seen an impact of these on the actual functionality of Ragna. So not sure what is going wrong in the background, but I think we can just ignore them. If we have time at some point, we can report this to panel.

With increasing the timeout to 30 seconds everything runs smoothly locally for me. I'll push this change to your branch to see if CI is happy as well.

tests/deploy/utils.py
```diff
@@ -0,0 +1,40 @@
...
+import time
+
+from fastapi.testclient import TestClient
+
+from ragna.assistants import RagnaDemoAssistant
+from ragna.core._utils import default_user
+
+
+class TestAssistant(RagnaDemoAssistant):
+    def answer(self, prompt, sources, *, multiple_answer_chunks: bool):
```
pmeier, line 11: The pydantic error that happens as part of the error log in is

```
pydantic_core._pydantic_core.ValidationError: 1 validation error for 1d49d534-1b0e-4d37-bbf7-3d922f898376
multiple_answer_chunks
  Field required [type=missing, input_value={'user': '', 'chat_id': U...at 05/22/2024 10:21 AM'}, input_type=dict]
    For further information visit https://errors.pydantic.dev/2.7/v/missing
```

It complains that this assistant requires the `multiple_answer_chunks` parameter since it has no default, but we didn't pass it. Unfortunately, there is no way to do this from the web UI at the moment. See https://ragna.chat/en/latest/generated/tutorials/gallery_custom_components/#web-ui_1 for details.

I'm ok with re-using this `TestAssistant` for the API and UI tests. But to make it work for the UI, we need to give `multiple_answer_chunk: bool = True` a default value. Just add a comment why the default value is there so we can remove it later on when we no longer need it.

.github/workflows/test.yml
```diff
@@ -68,7 +68,19 @@ jobs:
 
       - name: Run unit tests
         id: tests
-        run: pytest --junit-xml=test-results.xml --durations=25
+        run: |
```
pmeier, line 71: We probably should run the UI tests in a different job to be able to parametrize over the browser. I would test Chromium and Firefox on every operating system and additionally Safari (WebKit) on macOS. As for the Python version, I'd say we use the same as we have for docker, i.e.

https://github.com/Quansight/ragna/blob/a45bd9063f106a7728c5fb65c9077387b25c9706/Dockerfile#L1

Meaning, we'll have

```yaml
strategy:
  matrix:
    os:
      - ubuntu-latest
      - windows-latest
      - macos-latest
    browser:
      - chromium
      - firefox
    include:
      - os: macos-latest
        browser: webkit
```

The test command for the "regular" tests will change to `pytest --ignore tests/ui` and for UI tests `pytest tests/ui --browser ${{ matrix.browser }}`.

tests/deploy/api/test_batch_endpoints.py
```diff
@@ -4,7 +4,7 @@
 from ragna.deploy import Config
 from ragna.deploy._api import app
 
-from .utils import authenticate
+from ..utils import authenticate
```
pmeier, line 7: Let's not have relative imports that go beyond the current namespace

```suggestion
from tests.utils import authenticate
```

tests/deploy/utils.py
```diff
@@ -0,0 +1,44 @@
...
+        content = next(super().answer(prompt, sources))
+
+        if multiple_answer_chunks:
+            for chunk in content.split(" "):
+                yield f"{chunk} "
+        else:
+            yield content
+
+
+def authenticate(client: TestClient) -> None:
```
pmeier, line 29: Let's make the name more expressive now that we have this in the common folder.

```suggestion
def authenticate_with_api(client: TestClient) -> None:
```

tests/deploy/api/test_components.py
```diff
@@ -7,7 +7,7 @@
 from ragna.deploy import Config
 from ragna.deploy._api import app
 
-from .utils import authenticate
+from ..utils import authenticate
```
pmeier, line 10: Same as above.

tests/deploy/api/test_e2e.py
```diff
@@ -1,29 +1,12 @@
...
-        # for why this is needed.
-        time.sleep(1e-3)
-        content = next(super().answer(prompt, sources))
-
-        if multiple_answer_chunks:
-            for chunk in content.split(" "):
-                yield f"{chunk} "
-        else:
-            yield content
+from ..utils import TestAssistant, authenticate
```
pmeier, line 9: Same as above.

tests/deploy/ui/test_ui.py
```diff
@@ -0,0 +1,174 @@
...
+
+import httpx
+import panel as pn
+import pytest
+from playwright.sync_api import Page, expect
+
+from ragna._utils import timeout_after
+from ragna.deploy import Config
+
+from ..utils import TestAssistant
```
pmeier, line 14: Same as above.

tests/deploy/ui/test_ui.py
```diff
@@ -0,0 +1,174 @@
...
+
+
+@pytest.fixture(scope="session")
+def api_port():
+    return get_available_port()
+
+
+@pytest.fixture(scope="session")
+def ui_port():
+    return get_available_port()
```
pmeier, line 30: Why are we keeping the same port throughout the same session? Is this a requirement somehow?

tests/deploy/ui/test_ui.py
```diff
@@ -0,0 +1,174 @@
...
+
+
+@pytest.fixture
+def server(config):
+    server = Server(config)
+    try:
+        server.start()
+        yield server
+    finally:
+        server.stop()
```
pmeier, line 97: Nit: this is the textbook example of making `Server` a context manager.

tests/deploy/ui/test_ui.py
```diff
@@ -0,0 +1,174 @@
...
+
+
+def test_health(server, page: Page) -> None:
+    health_url = server.base_url + "/health"
```
pmeier, line 101:
```suggestion
    health_url = f"{server.base_url}/health"
```

tests/deploy/ui/test_ui.py
```diff
@@ -0,0 +1,174 @@
...
+    response = page.goto(health_url)
+    assert response.ok
+
+
+def test_index(server, config, page: Page) -> None:
```
pmeier, line 106: Make the test name more expressive.

```suggestion
def test_start_chat(server, config, page: Page) -> None:
```

pmeier (commented):
Good work Blake! A few minor comments, but this already looks pretty good.

tests/deploy/ui/test_ui.py
blakerosenthal: Ah, that was leftover from when everything was broken and I was trying to reduce the number of variables. Reverted.

.github/workflows/test.yml
```diff
@@ -68,10 +68,76 @@ jobs:
...
+          - python-version: "3.11"
+            os: windows-latest
+          - python-version: "3.10"
+            os: macos-latest
+          - python-version: "3.11"
+            os: macos-latest
+        include:
+          - browser: webkit
+            os: macos-latest
+            python-version: "3.9"
```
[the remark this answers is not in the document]
blakerosenthal: Yeah I agree, let's keep an eye on this. I don't think we need so many python versions in the matrix.

tests/deploy/ui/test_ui.py
```diff
@@ -0,0 +1,172 @@
...
+
+
+@pytest.fixture
+def ui_port():
+    return get_available_port()
+
+
+@pytest.fixture(scope="session")
+def headed_mode(pytestconfig):
+    return pytestconfig.getoption("headed") or False
```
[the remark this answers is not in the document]
blakerosenthal: Good point. This is something playwright uses already and I was thinking I'd pass it on to ragna in the `open_browser` option but think it would be confusing to have multiple browsers opening during tests. I'll remove it.

PR #418, opened 2024-05-16
unify API and UI servers

### LINKED ISSUE #417: serve REST API and web UI from the same server
### Feature description

Per title. 

- Replace the `ragna api` and `ragna ui` commands with `ragna deploy`
- The `ragna deploy` command should still have the option to deploy the the REST API and web UI individually
- The default port should be 31476
- The API should be served under `/api` and the UI under `/ui`

### Value and/or benefit

- Deploying Ragna will be easier, since we longer have two web servers that need to communicate with each other. To avoid increasing the scope here, it is ok that the web UI will still communicate with the REST API through HTTP although they are served together. Fixing this is tracked in #392.
- This is an important step towards refactoring the auth workflow.

### Anything else?

Depends on #416.

pmeier:
Closes #417. This PR mostly moves stuff around there are very few actual functional changes:

- `ragna.deploy._cli` was moved to `ragna._cli`
- The general part of `ragna.deploy._api.core` (error handling, CORS, etc.) was moved into `ragna.deploy._core`

The actual functional changes are all things listed in #417 and a massively simplified config. Since we no longer need a separate hostname / port / ... for the REST API and web UI, I was able to get completely rid of the sections. Meaning, our configuration at this point is flat.

This PR can already be reviewed. The code however is currently defunct due to bokeh/bokeh-fastapi#8.


pmeier:
@pierrotsmnrd To run this you need to 

```shell
$ pip install git+https://github.com/holoviz/panel@main git+https://github.com/bokeh/bokeh-fastapi
```

You also need to create a new config with `ragna init`. Either set a different database URL there or remove any existing from `~/.cache/ragna/ragna.db`. Finally, API and UI are now started at the same time with `ragna deploy`.

blakerosenthal:
> @pierrotsmnrd To run this you need to
> 
> ```shell
> $ pip install git+https://github.com/holoviz/panel@main git+https://github.com/bokeh/bokeh-fastapi@main
> ```
> 
> You also need to create a new config with `ragna init`. Either set a different database URL there or remove any existing from `~/.cache/ragna/ragna.db`. Finally, API and UI are now started at the same time with `ragna deploy`.

I needed to bump python to 3.10 in order to get this to install

ragna/deploy/_config.py
```diff
@@ -76,14 +75,48 @@ def settings_customise_sources(
...
-            if isinstance(value, ConfigBase):
-                value._resolve_default_sentinels(config)
-            elif isinstance(value, AfterConfigValidateDefault):
+            if isinstance(value, AfterConfigValidateDefault):
                 setattr(self, name, value.make_default(config))
 
+    @property
+    def _url(self) -> str:
+        return f"http://{self.hostname}:{self.port}{self.root_path}"
+
```
blakerosenthal, line 119: Should I be able to specify a custom root path in the config file? When I try something like
> root_path = "/test"

I get 404s everywhere. Am I misunderstanding what this does or should this root path be passed along to the router?

ragna/deploy/_ui/app.py
```diff
@@ -23,17 +22,13 @@
 
 
 class App(param.Parameterized):
-    def __init__(self, *, hostname, port, api_url, origins, open_browser):
```
blakerosenthal, line 26: Does `open_browser` have a new home? I noticed that `ragna deploy` doesn't take me to my browser anymore.

blakerosenthal (approved):
It's really neat to see the API and the UI under one roof. The simplified config is much nicer to read too.

ragna/deploy/_ui/app.py
pmeier: Yeah that is kinda unfortunate, but no way around it for now. `uvicorn` has no option to open the browser on startup. I have some ideas on how to get around that and will open an issue. For now, let's not block the overall development over this.

ragna/deploy/_config.py
pmeier: TL;DR: You are seeing 404, because you set the `root_path` option, but you are not behind a proxy that enforces this.

The `root_path` config option is needed when you run Ragna behind a proxy, e.g. [`jupyter-server-proxy`](https://pypi.org/project/jupyter-server-proxy/). What this option does is to automatically handle the routing for you inside our code.

Imagine Ragna is deployed behind a proxy and users can access it at `my-company.com/ragna`. In this configuration, the `root_path` would be `/ragna`. When we now define an endpoint as `@app.post("/foo")`, FastAPI automatically routes any request to `/ragna/foo` to the endpoint. Meaning, the `root_path` is an abstraction so our application is agnostic to any potential proxy in front of it.

 See the [config reference](
https://ragna.chat/en/stable/references/config/#root_path). If you think the documentation is not sufficient, feel free to send a PR to improve it.

ragna/deploy/_ui/app.py
pmeier: See #430.

PR #419, opened 2024-05-17
Fix #415 : chat bubble

### LINKED ISSUE #415: chat bubble cannot be loaded
### Bug description

Chat bubble icon in left side bar is missing

image

This was broken in #312. Previously it looked like

image

The problem is the relative link in https://github.com/Quansight/ragna/blob/0cb205f24ca9fb322ac32a2e1aaec6b34928cd2e/ragna/deploy/_ui/css/left_sidebar/button.css#L27

Since we are currently in `/css/left_sidebar`, it looks for `/css/left_sidebar/imgs/chat_bubble.svg` and fails.

### How to reproduce the bug?

Open the web UI and look at the left sidebar

`main` after #312

### Anything else?

This can be fixed by using `url("../../imgs/chat_bubble.svg")`, but I don't know if that is the best way to do it.

pierrotsmnrd:
adding back the chat bubble that disappeared when I refactored the CSS

### DIFF
--- a/ragna/deploy/_ui/css/left_sidebar/button.css
+++ b/ragna/deploy/_ui/css/left_sidebar/button.css
@@ -24,7 +24,7 @@
 }
 
 :host(.chat_button) div button:before {
-  content: url("imgs/chat_bubble.svg");
+  content: url("../../imgs/chat_bubble.svg");
   margin-right: 10px;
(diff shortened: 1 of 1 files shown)


pmeier (approved):
Thanks Pierre!

PR #424, opened 2024-05-24
Add OpenAI API compatible assistant

nenb:
PoC - do not merge.

**Details**
1. Download https://huggingface.co/Mozilla/Mistral-7B-Instruct-v0.2-llamafile/blob/main/mistral-7b-instruct-v0.2.Q5_K_M.llamafile (or host your own model locally)
2. `export BASE_URL=http://localhost:8080` (or wherever your model is available locally)
3. Add `assistants = [
    "ragna.assistants.OpenAIApiCompatible",
]` to `ragna.toml`

### DIFF
--- a/ragna/assistants/_openai_api_compatible.py
+++ b/ragna/assistants/_openai_api_compatible.py
@@ -1,0 +1,74 @@
+import json
+import os
+from typing import AsyncIterator, cast
+
+import httpx
+
+import ragna
+from ragna.core import PackageRequirement, Requirement, Source
+
+from ._api import ApiAssistant
+
+
+class OpenAIApiCompatible(ApiAssistant):
+    def __init__(self) -> None:
+        self._client = httpx.AsyncClient(
+            headers={"User-Agent": f"{ragna.__version__}/{self}"},
+            timeout=60,
+        )
+        self._base_url = os.environ.get("BASE_URL")
+
+    @classmethod
+    def requirements(cls) -> list[Requirement]:
+        return [PackageRequirement("httpx_sse")]
+
+    @classmethod
+    def display_name(cls) -> str:
+        return "OpenAIApiCompatible"
+
+    def _make_system_content(self, sources: list[Source]) -> str:
+        instruction = (
+            "You are an helpful assistants that answers user questions given the context below. "
+            "If you don't know the answer, just say so. Don't try to make up an answer. "
+            "Only use the sources below to generate the answer."
+        )
+        return instruction + "\n\n".join(source.content for source in sources)
+
+    async def _call_api(
+        self, prompt: str, sources: list[Source], *, max_new_tokens: int
+    ) -> AsyncIterator[str]:
+        import httpx_sse
+
+        async with httpx_sse.aconnect_sse(
+            self._client,
+            "POST",
+            f"{self._base_url}/v1/chat/completions",
+            headers={
+                "Content-Type": "application/json",
+            },
+            json={
+                "messages": [
+                    {
+                        "role": "system",
+                        "content": self._make_system_content(sources),
+                    },
+                    {
+                        "role": "user",
+                        "content": prompt,
+                    },
+                ],
+                "temperature": 0.0,
+                "max_tokens": max_new_tokens,
+                "stream": True,
+            },
+        ) as event_source:
+            await self._assert_api_call_is_success(event_source.response)
+            async for sse in event_source.aiter_sse():
+                data = json.loads(sse.data)
+                choice = data["choices"][0]
+                if choice["finish_reason"] is not None:
+                    break
+                try:
+                    yield cast(str, choice["delta"]["content"])
+                except KeyError:
+                    pass
--- a/ragna/assistants/__init__.py
+++ b/ragna/assistants/__init__.py
@@ -10,6 +10,7 @@ __all__ = [
     "Gpt4",
     "Jurassic2Ultra",
     "RagnaDemoAssistant",
+    "OpenAIApiCompatible",
 ]
 
 from ._ai21labs import Jurassic2Ultra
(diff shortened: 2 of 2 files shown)


ragna/assistants/_openai_api_compatible.py
```diff
@@ -0,0 +1,74 @@
...
+            "Only use the sources below to generate the answer."
+        )
+        return instruction + "\n\n".join(source.content for source in sources)
+
+    async def _call_api(
+        self, prompt: str, sources: list[Source], *, max_new_tokens: int
+    ) -> AsyncIterator[str]:
+        import httpx_sse
+
+        async with httpx_sse.aconnect_sse(
```
pmeier, line 42: Not all APIs use SSE for streaming, e.g. Ollama (#376). Meaning, although the request is the same, the response is coming back in a different protocol and has to be handled accordingly.

ragna/assistants/_openai_api_compatible.py
```diff
@@ -0,0 +1,74 @@
...
+                        "content": self._make_system_content(sources),
+                    },
+                    {
+                        "role": "user",
+                        "content": prompt,
+                    },
+                ],
+                "temperature": 0.0,
+                "max_tokens": max_new_tokens,
+                "stream": True,
```
pmeier, line 62: All of OpenAI, Ollama, and vLLM need to get passed a model here as well.

ragna/assistants/_openai_api_compatible.py
```diff
@@ -0,0 +1,74 @@
...
+
+    async def _call_api(
+        self, prompt: str, sources: list[Source], *, max_new_tokens: int
+    ) -> AsyncIterator[str]:
+        import httpx_sse
+
+        async with httpx_sse.aconnect_sse(
+            self._client,
+            "POST",
+            f"{self._base_url}/v1/chat/completions",
```
pmeier, line 45: Azure OpenAI needs a [different scheme](https://learn.microsoft.com/en-us/azure/ai-services/openai/chatgpt-quickstart?tabs=command-line%2Cpython-new&pivots=rest-api#rest-api):

```
$AZURE_OPENAI_ENDPOINT/openai/deployments/gpt-35-turbo/chat/completions?api-version=2024-02-01
```

pmeier (changes requested):
Thanks Nick for the PR. As much as I would like to have this generic assistant, it won't work as implemented right now. I've added a few comments below indicating places that would need to be made more flexible to serve "all" OpenAI compatible APIs.

My suggestion would be to create an `OpenAIAPICompatible` base class with a public method `make_request` that can be overridden by subclasses. Plus we also need to handle the different streaming methods.

ragna/assistants/_openai_api_compatible.py
nenb: Given that we already have this logic (work by you!) for the Google assistant, can we just re-use it here?

If so, perhaps just a separate block of code, and another environment variable to control whether this block is executed?

ragna/assistants/_openai_api_compatible.py
pmeier: It is compliant. You just need to use a different URL as posted above. The `gpt-35-turbo` in there is the name of the deployment, e.g. the model name. Meaning, similar to the llamafile approach in this PR, you don't pass a model in the request (see https://github.com/Quansight/ragna/pull/424#discussion_r1615584286). Apart from that, this is exactly the same API as the public OpenAI one.

nenb: > It is compliant.

My understanding was that it is not compliant. Can you point me to some resources on this please? 

As a quick check, I went to the OpenAPI spec for Azure, and I saw [this](https://github.com/Azure/azure-rest-api-specs/blob/c3cc9abe085093ba880ee3eeb792edb4fa789553/specification/cognitiveservices/data-plane/AzureOpenAI/inference/stable/2024-02-01/inference.json#L1225) example for the `/chat/completions` endpoint, which appears to not be compliant. A quick skim of the Azure API documentation also seems to suggest [several](https://learn.microsoft.com/en-us/azure/ai-services/openai/reference) other examples.

pmeier: > My understanding was that it is not compliant. Can you point me to some resources on this please?

I can't. We have used it in an environment that I cannot share publicly.

> As a quick check, I went to the OpenAPI spec for Azure, and I saw [this](https://github.com/Azure/azure-rest-api-specs/blob/c3cc9abe085093ba880ee3eeb792edb4fa789553/specification/cognitiveservices/data-plane/AzureOpenAI/inference/stable/2024-02-01/inference.json#L1225) example for the `/chat/completions` endpoint, which appears to not be compliant.

IIUC, these are just extra parameters that Azure OpenAI supports but OpenAI does not. Meaning Azure OpenAI seems to be a superset of the public API. I can attest to the fact that you do not need to set any of these parameters to use it.

> A quick skim of the Azure API documentation also seems to suggest [several](https://learn.microsoft.com/en-us/azure/ai-services/openai/reference) other examples.

I did not see anything that would make it non-compliant. Could you share the specific section / example that you are referring to?

nenb: > Meaning Azure OpenAI seems to be a superset of the public API.
Maybe this reduces to me having a different (incorrect?) definition of what a compliant API is then. I would have assumed that a superset automatically made it non-compliant.

> I did not see anything that would make it non-compliant.
They had stuff related to an `operations` [endpoint](https://learn.microsoft.com/en-us/azure/ai-services/openai/reference#get-a-generated-image-result-dall-e-2-preview) that I didn't recognise on a quick glance. Having a version number as a query parameter also seemed like it could create compatibility problems, unless things were deprecated/added at the same time for both API endpoints, which they aren't. But again, maybe I have a different or incorrect understanding of what compatibility means here.

nenb:
Superseded by #425

PR #429, opened 2024-06-10
Load message history to core chat from schema chat

blakerosenthal:
Basic prereq for passing chat history down the pipeline. History is already present in the `schema.Chat` object loaded from the database, so we just need to make sure it also gets loaded into the `ragna.core.Chat` object inside [schema_to_core_chat](https://github.com/Quansight/ragna/blob/647a61e3103cb2e54b7c5bc2da5df1e2906b059b/ragna/deploy/_api/core.py#L213).

https://github.com/Quansight/ragna/blob/647a61e3103cb2e54b7c5bc2da5df1e2906b059b/ragna/deploy/_api/core.py#L290-L297

### DIFF
--- a/ragna/deploy/_api/core.py
+++ b/ragna/deploy/_api/core.py
@@ -234,7 +234,7 @@ def app(*, config: Config, ignore_unavailable_components: bool) -> FastAPI:
             **chat.metadata.params,
         )
         core_chat._messages = [
-            schemas.Message.to_core(message) for message in chat.messages
+            message.to_core() for message in chat.messages
         ]
         core_chat._prepared = chat.prepared
 
--- a/ragna/deploy/_api/core.py
+++ b/ragna/deploy/_api/core.py
@@ -233,7 +233,9 @@ def app(*, config: Config, ignore_unavailable_components: bool) -> FastAPI:
             chat_name=chat.metadata.name,
             **chat.metadata.params,
         )
-        core_chat._messages = [message.to_core() for message in chat.messages]
+        core_chat._messages = [
+            message.to_core() for message in chat.messages
+        ]
         core_chat._prepared = chat.prepared
 
         return core_chat
--- a/ragna/deploy/_api/core.py
+++ b/ragna/deploy/_api/core.py
@@ -233,11 +233,9 @@ def app(*, config: Config, ignore_unavailable_components: bool) -> FastAPI:
             chat_name=chat.metadata.name,
             **chat.metadata.params,
         )
-        # FIXME: We need to reconstruct the previous messages here. Right now this is
-        #  not needed, because the chat itself never accesses past messages. However,
-        #  if we implement a chat history feature, i.e. passing past messages to
-        #  the assistant, this becomes crucial.
-        core_chat._messages = []
+        core_chat._messages = [
+            schemas.Message.to_core(message) for message in chat.messages
+        ]
         core_chat._prepared = chat.prepared
 
         return core_chat
--- a/ragna/deploy/_api/core.py
+++ b/ragna/deploy/_api/core.py
@@ -233,9 +233,7 @@ def app(*, config: Config, ignore_unavailable_components: bool) -> FastAPI:
             chat_name=chat.metadata.name,
             **chat.metadata.params,
         )
-        core_chat._messages = [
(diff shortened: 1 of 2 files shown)


ragna/deploy/_api/schemas.py
```diff
@@ -66,6 +83,14 @@ def from_core(cls, message: ragna.core.Message) -> Message:
             sources=[Source.from_core(source) for source in message.sources],
         )
 
+    @classmethod
+    def to_core(cls, message: Message) -> ragna.core.Message:
+        return ragna.core.Message(
+            content=message.content,
+            role=message.role,
+            sources=[Source.to_core(source) for source in message.sources],
+        )
```
pmeier, line 92: I don't understand why these are class methods. They should be default instance methods

```suggestion
    def to_core(self) -> ragna.core.Message:
        return ragna.core.Message(
            content=self.content,
            role=self.role,
            sources=[source.to_core() for source in self.sources],
        )
```

ragna/deploy/_api/core.py
```diff
@@ -233,11 +233,9 @@ def schema_to_core_chat(
...
             **chat.metadata.params,
         )
-        # FIXME: We need to reconstruct the previous messages here. Right now this is
-        #  not needed, because the chat itself never accesses past messages. However,
-        #  if we implement a chat history feature, i.e. passing past messages to
-        #  the assistant, this becomes crucial.
-        core_chat._messages = []
+        core_chat._messages = [
+            schemas.Message.to_core(message) for message in chat.messages
+        ]
```
pmeier, line 238:
```suggestion
        core_chat._messages = [
            message.to_core() for message in chat.messages
        ]
```

pmeier (changes requested):

ragna/deploy/_api/schemas.py
blakerosenthal: That makes way more sense. Idk why I did it that way.

ragna/deploy/_api/schemas.py
```diff
@@ -66,6 +66,13 @@ def from_core(cls, message: ragna.core.Message) -> Message:
             sources=[Source.from_core(source) for source in message.sources],
         )
 
+    def to_core(self) -> ragna.core.Message:
+        return ragna.core.Message(
+            content=self.content,
+            role=self.role,
+            sources=[],
+        )
```
blakerosenthal, line 74: @pmeier Tell me if this makes sense:
For the purpose of chat history, we don't need Sources for each individual message since they already exist on the chat object. So we don't need to rebuild them here (?)

pmeier: > since they already exist on the chat object

They don't. `Source`'s are returned by the `SourceStorage` and are never saved on the `Chat` but only ever on the `Message`

https://github.com/Quansight/ragna/blob/8ea62a3de0a7131b439b522c8a90587b5b2080ad/ragna/core/_rag.py#L221-L227

If you have a look at the ORM model for the chat, we also have no connection to the sources table, but only the messages

https://github.com/Quansight/ragna/blob/8ea62a3de0a7131b439b522c8a90587b5b2080ad/ragna/deploy/_api/orm.py#L75-L92

The messages table however is connected to the sources table

https://github.com/Quansight/ragna/blob/8ea62a3de0a7131b439b522c8a90587b5b2080ad/ragna/deploy/_api/orm.py#L95-L100

pmeier (approved):
LGTM after 8a984a6a5245e6b3daf75ba4c13474a244f812cf is reverted as explained in https://github.com/Quansight/ragna/pull/429#discussion_r1635930993.

blakerosenthal:
@pmeier Mypy points out that you can't construct a `ragna.core.Document` from a `schema.Document` because it's missing the `metadata` field. I see a couple ways around that:

1. Add a `metadata` field to `schema.Document`, although it seems like that might interact weirdly with the multi-document code that's strewn around that keeps the documents and metadata separate. It would create some redundancies to clean up, but I think it could be nice to just have metadata on the schema object.
2. Cascade all the object conversions directly in `schema_to_core_chat` kind of like the functions found in `database.py`.

I'm leaning toward option 1 unless there's a simpler way.

PR #431, opened 2024-06-12
re-enable browser opening functionality 

### LINKED ISSUE #430: Re-enable the option to open the browser when UI is started
### Feature description

This feature already exists `main`, but was removed with #418 while doing the deploy rework (https://github.com/Quansight/ragna/pull/418#discussion_r1635333975). In contrast to `panel`, `uvicorn` does not have an option to open a browser on startup.

We might be able to use the [`lifespan`](https://fastapi.tiangolo.com/advanced/events/) functionality to do this. However, we should survey the panel code first to see how they do it.

### Value and/or benefit

Opening the browser when running `ragna deploy` is a good UX when trying Ragna.

### Anything else?

_No response_

pmeier:
Closes #430.

### DIFF
--- a/ragna/_cli/core.py
+++ b/ragna/_cli/core.py
@@ -96,6 +96,10 @@ def deploy(
             )
         ),
     ] = False,
+    open_browser: Annotated[
+        Optional[bool],
+        typer.Option(help="Open a browser when Ragna is deployed."),
+    ] = None,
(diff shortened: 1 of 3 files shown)


blakerosenthal (approved):
Nice usage of fastapi's lifespan events 🥳

PR #432, opened 2024-06-13
WIP draft of generate() outside of chat.answer()

dillonroach:
Takes format from chat.answer and duplicates, without sources, as chat.generate which then gets implemented in Assistant as well. Got a bit turned around with the async aspect in the stream = False case, room for cleanup. The exl2 assistant is currently instantiating a number of vars it likely should be taking from Chat instead (e.g. stream:bool)

### DIFF
--- a/environment-dev-exl.yml
+++ b/environment-dev-exl.yml
@@ -1,0 +1,43 @@
+name: ragna-dev-exl
+channels:
+  - pytorch
+  - nvidia
+  - conda-forge
+  - defaults
+dependencies:
+  - python =3.11
+  - pip
+  - git-lfs
+  - jupyterlab>=3
+  - pandas
+  - numpy
+  - panel
+  - tokenizers
+  - pytorch=2.3.1
+  - pytorch-cuda=12.1
+  - cuda-nvcc
+  - rich
+  - pip:
+      - python-dotenv
+      - pytest >=6
+      - pytest-mock
+      - pytest-asyncio
+      - pytest-playwright
+      - mypy ==1.10.0
+      - pre-commit
+      - types-aiofiles
+      - sqlalchemy-stubs
+      - setuptools-scm
+      - pip-tools
+      # documentation
+      - mkdocs
+      - mkdocs-material
+      - mkdocstrings[python]
+      - mkdocs-gen-files
+      - material-plausible-plugin
+      - mkdocs-gallery >=0.10
+      - mdx_truly_sane_lists
+      # exl2
+      - ninja
+      - packaging
+      - exllamav2@https://github.com/turboderp/exllamav2/releases/download/v0.1.5/exllamav2-0.1.5+cu121.torch2.3.1-cp311-cp311-linux_x86_64.whl
--- a/ragna/assistants/__init__.py
+++ b/ragna/assistants/__init__.py
@@ -18,8 +18,10 @@ __all__ = [
     "Jurassic2Ultra",
     "LlamafileAssistant",
     "RagnaDemoAssistant",
+    "Exl2Assistant",
 ]
 
+from ._exl2 import Exl2Assistant
 from ._ai21labs import Jurassic2Ultra
 from ._anthropic import ClaudeHaiku, ClaudeOpus, ClaudeSonnet
 from ._cohere import Command, CommandLight
--- a/ragna/assistants/_exl2.py
+++ b/ragna/assistants/_exl2.py
@@ -1,0 +1,123 @@
+import re
+import textwrap
+from typing import Iterator, Union, cast
+
+from ragna.core import Assistant, Source
+from pathlib import Path
+
+from exllamav2 import(
+    ExLlamaV2,
+    ExLlamaV2Config,
+    ExLlamaV2Cache,
+    ExLlamaV2Tokenizer,
+    ExLlamaV2Cache_Q4,
+)
+
+from exllamav2.generator import (
+    ExLlamaV2BaseGenerator,
+    ExLlamaV2Sampler,
+    ExLlamaV2DynamicGenerator,
+    ExLlamaV2DynamicJob,
+    ExLlamaV2DynamicGeneratorAsync,
+    ExLlamaV2DynamicJobAsync,
+)
+
+import time
+
+class Exl2Assistant(Assistant):
+    """Exl2Assistant - example to instantiate and run inference in process
+    """
+
+    @classmethod
+    def display_name(cls) -> str:
+        return "Ragna/Exl2Assistant"
+
+    # TODO; known needs: - pytorch, pytorch-cuda, [cuda-nvcc, rich, ninja, packaging, flash-attn] for paged attention batching, exllamav2
+    # @classmethod
+    # def requirements(cls, protocol: HttpStreamingProtocol) -> list[Requirement]:
+    #     streaming_requirements: dict[HttpStreamingProtocol, list[Requirement]] = {
+    #         HttpStreamingProtocol.SSE: [PackageRequirement("httpx_sse")],
+    #     }
+    # return streaming_requirements.get(protocol, [])
+
+    def __init__(
+        self,
+    ) -> None:
+        self._stream = False
+        self._paged = False
+        self._max_seq_length = 8192
+        self._max_new_tokens = 512
+        self._model_directory = ""
+        self._load()
+        
+    def _load(self):
+        self._config = ExLlamaV2Config(self._model_directory)
+        self._config.prepare()
+        self._config.max_seq_len = self._max_seq_length
+        self._model = ExLlamaV2(self._config)
+        self._cache = ExLlamaV2Cache_Q4(self._model, lazy = True, max_seq_len=self._config.max_seq_len)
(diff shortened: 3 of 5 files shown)


dillonroach:
This PR needs to coordinate with https://github.com/Quansight/ragna/pull/438 - wherever the final pattern lands there, I need to integrate in the generate() and answer()(_stream()) ingest points.  This is also modulo however we expect to pass prompt str or messages, etc, through the pre-process stage. Will make this a point of conversation in our meet Monday.

dillonroach:
@pmeier - I'd already built out most of the assistant examples, so just did them all this pass- but we don't have to discuss them all for first round; lets just pick _anthropic and _openai and see how those are looking specifically.

In the interest of helping this merge quickly, I'll also pull out exl2 for another PR that's just about integrating that as an assistant on its own.

ragna/assistants/_openai.py
```diff
@@ -23,9 +23,45 @@ def _make_system_content(self, sources: list[Source]) -> str:
...
+                        },
+                    ]
+            return messages
+        else:
+            system_message = [{"role":"system", "content":system_prompt}]
+            messages = [{"role":i["role"],"content":i["content"]} for i in prompt if i["role"] != "system"]
+            return system_message.extend(messages)
+    
+    async def generate(
+        self, prompt: Union[str,list[Message]], system_prompt: str, *, max_new_tokens: int
```
pmeier, line 51: Both `system_prompt` and `max_new_tokens` should have a reasonable default values. We don't want to force users to always put something there. Plus, `system_prompt` should probably also be keyword only. Calling `.generate()` with two strings filled with instructions can become confusing real quick.

ragna/assistants/_anthropic.py
```diff
@@ -36,12 +36,50 @@ def _instructize_system_prompt(self, sources: list[Source]) -> str:
             + "</documents>"
         )
 
-    async def answer(
-        self, messages: list[Message], *, max_new_tokens: int = 256
+    def _render_prompt(self, prompt: Union[str, list[Message]]) -> Union[str, list]:
```
pmeier, line 39: How can this return a `str`?

ragna/assistants/_anthropic.py
```diff
@@ -36,12 +36,50 @@ def _instructize_system_prompt(self, sources: list[Source]) -> str:
...
+        Returns:
+            ordered list of dicts with 'content' and 'role' keys
+        """
+        if isinstance(prompt, str):
+            return [
+                {
+                    "content": prompt,
+                    "role": "user",
+                }
+            ]
```
pmeier, line 52: Instead of this pattern of handling strings and messages completely different, I would like to just wrap the prompt into a `Message` and proceed on the same branch from there on. Like

```python
if isinstance(prompt, str):
     messages = [Message(content=prompt, role=MessageRole.USER)]
else:
     messages = prompt

...
```

dillonroach: Hrm - the notion here was for _render_prompt() to convert from `list[Message]`s into whatever form the particular assistant would actually process at the API call level - e.g. here the anthropic 'messages' won't be 1:1 aligned with ragna list[Message]

Called https://github.com/Quansight/ragna/blob/82637dda6450184ec411ce55d0442dc87ffb6496/ragna/assistants/_anthropic.py#L95 right before it gets sent in to the API end-point.

In your proposal, would we convert any prompt into list[Message] and then from that into final API call format? If so, then I'm in agreement and will make the change.

ragna/assistants/_anthropic.py
dillonroach: yep, good call - I had originally meant to try to make all assistants follow the same in/out pattern, but I see I didn't even keep it in exl2 where I just set `-> str` - will drop the Union and make it just `list`

ragna/assistants/_anthropic.py
pmeier: Yes that is what I meant. I want to avoid having two branches with different return logic. The `prompt: str` input is just for convenience. We just put it into the "right" format and process everything after it the same way.

The output is of course dependent on the assistant and does not need to follow a common format. That is the reason why we are not putting this on the `Assistant` class as there is no standardization possible apart from our convention of calling the method `_render_prompt` everywhere.

ragna/assistants/_ai21labs.py
```diff
@@ -41,17 +74,23 @@ async def answer(
...
-                yield cast(str, data["outputs"][0]["text"])
+                yield data
+
+    async def answer(
+        self, messages: list[Message], *, max_new_tokens: int = 256
+    ) -> AsyncIterator[str]:
+        message = messages[-1]
+        async for data in self.generate(
+            [message],
+            system_prompt=self._make_system_content(message.sources),
```
pmeier, line 90: Given that we only use this function here and the system prompt is very specific to the RAG use case, should we just inline the function? Otherwise, we need to rename it to something like `_make_rag_system_content` to avoid confusion.

This applies to all assistant in this PR.

ragna/assistants/_anthropic.py
```diff
@@ -10,10 +10,6 @@ class AnthropicAssistant(HttpApiAssistant):
     _STREAMING_PROTOCOL = HttpStreamingProtocol.SSE
     _MODEL: str
 
-    @classmethod
-    def _extra_requirements(cls) -> list[Requirement]:
-        return [PackageRequirement("httpx_sse")]
```
pmeier, line 15: @dillonroach Driveby. The requirements for the streaming are handled automatically.

ragna/assistants/_cohere.py
```diff
@@ -40,23 +76,35 @@ async def answer(
                 "authorization": f"Bearer {self._api_key}",
             },
             json={
-                "preamble_override": self._make_preamble(),
-                "message": prompt,
+                "preamble_override": system_prompt,
+                "message": self._render_prompt(prompt),
```
pmeier, line 80: While the message indeed can only be a single string here, the endpoint has a [`chat_history`](https://docs.cohere.com/reference/chat#request.body.chat_history) parameter. And that one takes the previous message similar to all other assistants.

I would let `_render_prompt` return a tuple of the chat history and the current user message, e.g.

```python
chat_history, message = self._render_prompt(prompt)
```

ragna/assistants/_google.py
```diff
@@ -25,19 +25,38 @@ def _instructize_prompt(self, prompt: str, sources: list[Source]) -> str:
...
+    def _render_prompt(self, prompt: Union[str, list[Message]]) -> list[dict]:
+        if isinstance(prompt, str):
+            messages = [Message(content=prompt, role=MessageRole.USER)]
+        else:
+            messages = prompt
+        return [
+            {"parts": [{"text": message.content}]}
+            for message in messages
+            if message.role is not MessageRole.SYSTEM
+        ]
```
pmeier, line 37: This also needs the role: https://ai.google.dev/api/caching#Content

tests/assistants/test_api.py
```diff
@@ -82,7 +82,7 @@ def __init__(self, base_url):
         super().__init__()
         self._endpoint = f"{base_url}/{self._STREAMING_PROTOCOL.name.lower()}"
 
-    async def answer(self, messages):
+    async def generate(self, messages):
```
pmeier, line 85: @dillonroach I added the new generate / answer architecture to this test class as well to make sure we are testing as close to the actual implementation as possible.

pmeier (commented):
@dillonroach I pushed 8e120c6cc4db9b1ec911ea7a101519e8aacf1163. The two main changes I made are

1. Fix the return types of `.generate()` for all assistants. It now returns an async iterator of the full data the API returns. Since this function is aimed as a general building block for users, we shouldn't assume they only want the text.

   However, we can make this assumption for `.answer()`. Meaning, this function calls `.generate()` and extracts the text.
- The `_render_prompt` and corresponding functions treated incoming messages as dictionaries. `ragna.core.Message` is a proper object and thus key access will not work. Furthermore, `message.role` is the `ragna.core.MessageRole` enum and cannot be used directly. Instead, you need to use `message.role.value` if you want to have the string.

With my commit, CI should be green. Note that I base my review on the latest commit, i.e. including my cleanup. Thus, please make sure to review what I have pushed, because the comments otherwise might make no sense.

ragna/assistants/_ai21labs.py
dillonroach: My personal preference is to rename the function but keep it as a function call - I do see that if someone is calling a custom rag-like call outside of answer, they're perhaps likely to want to customize that piece as well, so who knows how it gets directly reused, but it does make things like subclassing a touch cleaner. I'll push a commit with the name change in  a moment

ragna/assistants/_cohere.py
dillonroach: Conflicted on this one - seems like this would fit specifically with the pre-process pass, or this puts this one specific assistant ahead of the others in terms of capabilities - it certainly doesn't hurt anything, so happy to do so here, but also see how we might want to see what a pre-process stage looks like for all assistants and implement in one go.

Issue #392, opened 2024-04-16
Standalone web UI


pmeier:
Our web UI is currently sits on top of the REST API and cannot be deployed as a standalone application. This architecture was chosen to avoid giving us a benefit over third-party UIs that have to sit on top of the REST API. Meaning, if the REST API we provide is sufficient for our web UI, it is also sufficient for third-party applications.

However, this requires HTTP communication between the web UI and REST API. Coupled with the fact that we are currently mixing requests from JS (document upload) and Python (everything else) we are also prone to running into CORS issues. See #268 for an example.

Meaning, deploying the Ragna web UI is harder than it needs to be. And the only reason for that is a constraint that we set for ourselves. I feel this was a mistake and we should fix it.

My proposal for this is to create an "backend engine" in Python that uses Ragnas Python API and handles the database connection etc. This engine would then be used by the REST API as well as the web UI to do the actual computation. Here is a picture I drew while discussing this with @blakerosenthal.

![refactor](https://github.com/Quansight/ragna/assets/6849766/ab04afa2-84cb-4a7e-b385-5c67c868e6e1)

In this architecture, we can still deploy the REST API and web UI at the same time, but we now also have the ability to deploy the web UI as standalone. Plus, our original constraint is also somewhat achievable: if the engine is sufficient to drive our web UI and we expose all functionality we need also on the REST API, the REST API should also be sufficient for third-party web UIs.

The only blocker that I currently see is the whole auth workflow. I have plans (and a [branch](https://github.com/Quansight/ragna/tree/auth)) to implement this properly and fixing #178 in the process. But that probably needs to happen before we have a go at this refactor here. Unfortunately, we are currently blocked by `panel`, i.e. holoviz/panel#6658 . I plan on working on this soon.

pmeier:
Fixed in #443.

PR #434, opened 2024-06-25
introduce engine for API

pmeier:
First part of #392. This just introduces the engine and let's the API use it. After this is merged, I send a follow-up that refactors the UI to also use the engine, which in turn will complete #392.


ragna/core/_components.py
```diff
@@ -157,6 +159,8 @@ def __init__(
         *,
         role: MessageRole = MessageRole.SYSTEM,
         sources: Optional[list[Source]] = None,
+        id: Optional[uuid.UUID] = None,
+        timestamp: Optional[datetime.datetime] = None,
```
pmeier, line 163: It was kinda weird to not have these fields here given that we have them in the database as well as in the schema. Without them, converting back and forth became harder, because we'd need to keep track of this information manually.

ragna/core/_rag.py
```diff
@@ -41,20 +45,69 @@ class Rag(Generic[C]):
         ```
     """
 
-    def __init__(self) -> None:
-        self._components: dict[Type[C], C] = {}
+    def __init__(
```
pmeier, line 48: Two new functionalities for the `ragna.Rag` object:

1. When creating a new chat with `Rag().chat()`, one now can also pass the display name of any loaded component instead of passing an instance or the class.
2. The object can now be created with a configuration object, which loads all configured source storages and assistants.

We had this before, but only for the API. While refactoring, it made the code a lot cleaner when moving it here.

ragna/deploy/_api.py
```diff
@@ -0,0 +1,163 @@
...
+def make_router(config: Config, engine: Engine) -> APIRouter:
+    router = APIRouter(tags=["API"])
+
+    def get_user() -> str:
+        return default_user()
+
+    UserDependency = Annotated[str, Depends(get_user)]
+
+    # TODO: the document endpoints do not go through the engine, because they'll change
+    #  quite drastically when the UI no longer depends on the API
```
pmeier, line 35: What the inline comment says. This will be addressed in the follow-up PR

tests/deploy/api/conftest.py
```diff
@@ -0,0 +1,41 @@
+import contextlib
+import json
+
+import httpx
+import pytest
+
+
+@pytest.fixture(scope="package", autouse=True)
+def enhance_raise_for_status(package_mocker):
```
pmeier, line 9: I was annoyed by the lack of information while debugging. `response.raise_for_status()` is a nice thing if you just want to check _if_ there is an error. With this, we now automagically also see the content if something fails alongside the status code.

ragna/deploy/_database.py
```diff
@@ -0,0 +1,274 @@
...
+            assistant=chat.metadata.assistant,
+            params=chat.metadata.params,
+            messages=[
+                self.message(message, chat_id=chat.id) for message in chat.messages
+            ],
+            prepared=chat.prepared,
+        )
+
+
+class OrmToSchemaConverter:
```
blakerosenthal, line 238: These classes are soo nice to have

ragna/deploy/_engine.py
```diff
@@ -0,0 +1,205 @@
...
+            self._database.update_chat(
+                session, chat=self._to_schema.chat(core_chat), user=user
+            )
+
+    def delete_chat(self, *, user: str, id: uuid.UUID) -> None:
+        with self._database.get_session() as session:
+            self._database.delete_chat(session, user=user, id=id)
+
+
+class SchemaToCoreConveter:
```
blakerosenthal, line 119:
```suggestion
class SchemaToCoreConverter:
```

ragna/deploy/_engine.py
```diff
@@ -0,0 +1,205 @@
...
+            chat_name=chat.metadata.name,
+            **chat.metadata.params,
+        )
+        core_chat._messages = [self.message(message) for message in chat.messages]
+        core_chat._prepared = chat.prepared
+
+        return core_chat
+
+
+class CoreToSchemaConveter:
```
blakerosenthal, line 163:
```suggestion
class CoreToSchemaConverter:
```

blakerosenthal (approved):
Amazing

pmeier:
Test failures are unrelated and will be fixed by #435.

PR #435, opened 2024-06-26
add support for chromadb>=0.5.1

pmeier:
`chromadb>=0.5.1` changed the output of `collection.query`. They added a new key-value pair to the result, which should be BC. However, since we before this PR iterated over the full dictionary and assumed that every key-value pair follows the same pattern, we are bitten by this.

This PR circumvents this assumption by only selecting the keys we actually need. Meaning, unless Chroma introduces a BC breaking change to them, we should be future proof.

### DIFF
--- a/ragna/source_storages/_chroma.py
+++ b/ragna/source_storages/_chroma.py
@@ -83,6 +83,7 @@ class Chroma(VectorDatabaseSourceStorage):
             str(chat_id), embedding_function=self._embedding_function
         )
 
+        include = ["distances", "metadatas", "documents"]
         result = collection.query(
             query_texts=prompt,
             n_results=min(
@@ -100,22 +101,19 @@ class Chroma(VectorDatabaseSourceStorage):
                 max(int(num_tokens * 2 / chunk_size), 100),
                 collection.count(),
             ),
-            include=["distances", "metadatas", "documents"],
+            include=include,
         )
 
         num_results = len(result["ids"][0])
-        result = {
-            key: [None] * num_results if value is None else value[0]  # type: ignore[index]
-            for key, value in result.items()
-        }
+        result = {key: result[key][0] for key in ["ids", *include]}
         # dict of lists -> list of dicts
         results = [
-            {key[:-1]: value[idx] for key, value in result.items()}
+            {key: value[idx] for key, value in result.items()}
             for idx in range(num_results)
         ]
 
         # That should be the default, but let's make extra sure here
-        results = sorted(results, key=lambda r: r["distance"])
+        results = sorted(results, key=lambda r: r["distances"])
 
         # TODO: we should have some functionality here to remove results with a high
         #  distance to keep only "valid" sources. However, there are two issues:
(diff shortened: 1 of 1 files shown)


ragna/source_storages/_chroma.py
```diff
@@ -100,22 +101,19 @@ def retrieve(
...
         # dict of lists -> list of dicts
         results = [
-            {key[:-1]: value[idx] for key, value in result.items()}
+            {key: value[idx] for key, value in result.items()}
             for idx in range(num_results)
         ]
 
         # That should be the default, but let's make extra sure here
-        results = sorted(results, key=lambda r: r["distance"])
+        results = sorted(results, key=lambda r: r["distances"])
```
nenb, line 116: Question: was this a bug?

ragna/source_storages/_chroma.py
```diff
@@ -100,22 +101,19 @@ def retrieve(
...
-            include=["distances", "metadatas", "documents"],
+            include=include,
         )
 
         num_results = len(result["ids"][0])
-        result = {
-            key: [None] * num_results if value is None else value[0]  # type: ignore[index]
-            for key, value in result.items()
-        }
+        result = {key: result[key][0] for key in ["ids", *include]}
```
nenb, line 108: Question: Handling the `None` case is no longer relevant?

ragna/source_storages/_chroma.py
```diff
@@ -127,11 +125,11 @@ def retrieve(
         return self._take_sources_up_to_max_tokens(
             (
                 Source(
-                    id=result["id"],
-                    document=document_map[result["metadata"]["document_id"]],
-                    location=result["metadata"]["page_numbers"],
-                    content=result["document"],
-                    num_tokens=result["metadata"]["num_tokens"],
+                    id=result["ids"],
```
nenb, line 128: Question: Is the addition of an 's' to all keys a `chroma` thing or a `ragna` thing? And is it correct to add the `s` to the id field as well?

ragna/source_storages/_chroma.py
pmeier: Sorry, I should have elaborated on this. I previously thought that it would be a good idea to turn the plural forms into singulars when going from dict of lists to list of dicts, i.e.

```python
{"distances": [...], "ids": [...]}
```

to

```
[{"distance": ..., "id": ...}, ...]
```

From a language standpoint this makes sense given that each result in the new list now has a singular distance, ID, etc. However, this makes it harder to map the Chroma documentation to the results that we want to return and I think we should value this higher than correct language. Thus, I've removed the transforming of the plural into the singular form by cutting of the final charactly (L113, `key[:-1]` -> L111, `key`).

ragna/source_storages/_chroma.py
pmeier: Correct. Anything that could be returned by Chroma, but is not present in the `include` list, results in `None` values. Since we no longer iterate over the full dictionary, we no longer need to care about them.

ragna/source_storages/_chroma.py
pmeier: See https://github.com/Quansight/ragna/pull/435#discussion_r1654976899.

PR #436, opened 2024-06-28
refactor protocol model extraction to only check extra parameters

pmeier:
This PR changes two things:

1. The previous implementation for the extraction of the protocol models relied on the fact that the concrete subclasses use the exact same names for the protocol parameters, i.e. the parameters defined by the protocol base class. This is rather restricting, undocumented, and has bitten us before (https://github.com/Quansight/ragna/pull/369#issuecomment-2019756910). With this PR we identify the extra parameters by skipping the first `n` parameters in the concrete signature, where `n` is the number of protocol parameters.
2. Previously we just passed through the Pydantic validation error in case of a failure. While this contains all the information you need if you know the internals of Ragna, it is hard to parse for users. Especially, because they don't have any touching point with Pydantic and we are thus "leaking internals". This PR adds logic to parse the raw errors and format them in a user friendly way.

For example:

```python
from ragna import Rag, source_storages
from ragna.core import Assistant


class ValidationAssistant(Assistant):
    def answer(
        self,
        prompt,
        sources,
        bool_param: bool,
        int_param: int,
        float_param: float,
        string_param: str,
    ):
        pass


async def main():
    async with Rag().chat(
        documents=["README.md"],
        source_storage=source_storages.RagnaDemoSourceStorage,
        assistant=ValidationAssistant,
        unknown=True,
        bool_param=False,
        float_param="string_value",
        string_param=42,
    ):
        pass


import asyncio

asyncio.run(main())
```

#### `main`

```
pydantic_core._pydantic_core.ValidationError: 4 validation errors for ec32e2f5-ad2d-4d08-8987-609f8e47a6d0
int_param
  Field required [type=missing, input_value={'user': 'philip', 'chat_...ue', 'string_param': 42}, input_type=dict]
    For further information visit https://errors.pydantic.dev/2.7/v/missing
float_param
  Input should be a valid number [type=float_type, input_value='string_value', input_type=str]
    For further information visit https://errors.pydantic.dev/2.7/v/float_type
string_param
  Input should be a valid string [type=string_type, input_value=42, input_type=int]
    For further information visit https://errors.pydantic.dev/2.7/v/string_type
unknown
  Extra inputs are not permitted [type=extra_forbidden, input_value=True, input_type=bool]
    For further information visit https://errors.pydantic.dev/2.7/v/extra_forbidden
```

#### PR

```
ragna.core.RagnaException: Validating the Chat parameters resulted in 4 errors:

The following parameters are unknown:

- unknown=True

The following parameters are missing:

- int_param: int

The following parameters have the wrong type:

- string_param: str = 42
- float_param: float = 'string_value'
```

### DIFF
--- a/ragna/core/_components.py
+++ b/ragna/core/_components.py
@@ -4,7 +4,15 @@ import abc
 import enum
 import functools
 import inspect
-from typing import AsyncIterable, AsyncIterator, Iterator, Optional, Type, Union
+from typing import (
+    AsyncIterable,
+    AsyncIterator,
+    Iterator,
+    Optional,
+    Type,
+    Union,
+    get_type_hints,
+)
 
 import pydantic
 import pydantic.utils
@@ -42,6 +50,16 @@ class Component(RequirementsMixin):
     def _protocol_models(
         cls,
     ) -> dict[tuple[Type[Component], str], Type[pydantic.BaseModel]]:
+        # This method dynamically builds a pydantic.BaseModel for the extra parameters
+        # of each method that is listed in the __ragna_protocol_methods__ class
+        # variable. These models are used by ragna.core.Chat._unpack_chat_params to
+        # validate and distribute the **params passed by the user.
+
+        # Walk up the MRO until we find the __ragna_protocol_methods__ variable. This is
+        # the indicator that we found the protocol class. We use this as a reference for
+        # which params of a protocol method are part of the protocol (think positional
+        # parameters) and which are requested by the concrete class (think keyword
+        # parameters).
         protocol_cls, protocol_methods = next(
             (cls_, cls_.__ragna_protocol_methods__)  # type: ignore[attr-defined]
             for cls_ in cls.__mro__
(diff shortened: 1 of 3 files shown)


ragna/core/_components.py
```diff
@@ -42,30 +50,47 @@ def __repr__(self) -> str:
...
-            concrete_params = inspect.signature(method).parameters
-            protocol_params = inspect.signature(
-                getattr(protocol_cls, method_name)
-            ).parameters
-            extra_param_names = concrete_params.keys() - protocol_params.keys()
-
-            models[(cls, method_name)] = pydantic.create_model(  # type: ignore[call-overload]
+            params = iter(inspect.signature(method).parameters.values())
+            annotations = get_type_hints(method)
+            # Skip over the protocol parameters in order for the model below to only
```
blakerosenthal, line 76: This is the main thing that was biting me when trying to add new required parameters to methods on `Assistant` subclasses. This seems to resolve everything for me, although full disclosure I haven't been able to think through all the nuances of how this might affect custom user components.

ragna/core/_rag.py
```diff
@@ -251,27 +255,120 @@ def _parse_documents(self, documents: Iterable[Any]) -> list[Document]:
...
+            chat_model = ChatModel.model_validate(params, strict=True)
+
+        chat_params = chat_model.model_dump(exclude_none=True)
         return {
             fn: model(**chat_params).model_dump()
             for fn, model in component_models.items()
         }
 
+    @contextlib.contextmanager
+    def _format_validation_error(
```
blakerosenthal, line 290: This makes it way easier to parse all my various user errors, thank you 🙏

blakerosenthal (approved):
This is a little over my head but it fixes a lot of the issues I was having last week

Issue #421, opened 2024-05-21
Chat history


pmeier:
### Feature description

Currently each prompt passed to `Chat.answer` will be independent of the chat history. Meaning, although it is presented as one chat, each prompt-answer-pair is independent from each other. This can cause some confusion as users might reference a previous prompt or answer will get a nonsensical answer given that the assistant is not aware of them.

To overcome this, I propose we add a chat history feature that can be toggled on and off. I'm not sure on what level this should apply:

1. On a global level, i.e. being passed to `ragna.Rag()` or set in the configuration
2. On a chat level, i.e. being passed to `Rag().chat()`
3. On a message level, i.e. being passed to `Chat().answer()`

I prefer 2. here, but I'm open to other opinions.

I'm not sure about the interface either. Naively, I would say that instead of passing the prompt to the assistant

https://github.com/Quansight/ragna/blob/84cf4f627ebf52b061681a7a8b106daef3e79a1d/ragna/core/_components.py#L216

We could also pass a list of messages here. The current case would have only a single item in there, whereas with chat history we would have multiple entries. If we go for this implementation, we also need to decide whether we pass `str`s or `ragna.core.Messages` and whether we pass a flat list or a list of prompt-answer-pairs. Let's survey the ecosystem and not reinvent the wheel here.

### Value and/or benefit

Less user confusion and actual conversations rather than independent Q/A.

### Anything else?

To pass historical messages to the assistant, we already have access to the `Chat._messages` attribute

https://github.com/Quansight/ragna/blob/84cf4f627ebf52b061681a7a8b106daef3e79a1d/ragna/core/_rag.py#L168

However, on the API side we never bothered to fill this up when loading a chat from the DB, since the chat so far never accesses the historical messages:

https://github.com/Quansight/ragna/blob/84cf4f627ebf52b061681a7a8b106daef3e79a1d/ragna/deploy/_api/core.py#L236-L240

blakerosenthal:
I lean towards option 2 as well: history at the chat level, maybe to default to True for all chats via the persistent config or during startup at the CLI level. Chat histories could be loaded in bulk from the DB and toggled on a per-chat basis via the UI.

blakerosenthal:
[...]

PR #438, opened 2024-07-04
support chat history

blakerosenthal:
This is a very rough first pass at https://github.com/Quansight/ragna/issues/421. I decided to investigate the naive approach of replacing
```python
class Assistant
    ...
    def answer(self, prompt: str, sources: list[Source]):
        ...
```
with
```python
class Assistant
    ...
    def answer(self, messages: list[Message]):
        ...
```
for just the HTTP-like assistants as a POC. Since a `Message` instance can hold a list of `Source`s, the `_messages` list in `Rag.Chat` serves as a cache of raw messages that live in the database (and what the user sees), while each assistant is responsible for formatting the message list according to the needs of the individual LLM.

### DIFF
--- a/ragna/assistants/_demo.py
+++ b/ragna/assistants/_demo.py
@@ -36,7 +36,7 @@ class RagnaDemoAssistant(Assistant):
         ).strip()
 
     def _default_answer(self, messages: list[Message]) -> str:
-        prompt = messages[0].content.strip()
+        prompt = messages[-1].content.strip()
         sources_display = []
         for message in messages:
             sources = message.sources
--- a/ragna/assistants/_demo.py
+++ b/ragna/assistants/_demo.py
@@ -39,9 +39,8 @@ class RagnaDemoAssistant(Assistant):
         ).strip()
 
     def _default_answer(self, messages: list[Message]) -> str:
-        prompt = messages[-1].content.strip()
+        prompt, sources = (message := messages[-1]).content, message.sources
         sources_display = []
-        sources = messages[-1].sources
         for source in sources:
             source_display = f"- {source.document.name}"
             if source.location:
--- a/ragna/assistants/_demo.py
+++ b/ragna/assistants/_demo.py
@@ -1,8 +1,7 @@
-import re
 import textwrap
 from typing import Iterator
 
-from ragna.core import Assistant, Source
+from ragna.core import Assistant, Message
 
 
 class RagnaDemoAssistant(Assistant):
@@ -22,11 +21,8 @@ class RagnaDemoAssistant(Assistant):
     def display_name(cls) -> str:
         return "Ragna/DemoAssistant"
 
-    def answer(self, prompt: str, sources: list[Source]) -> Iterator[str]:
-        if re.search("markdown", prompt, re.IGNORECASE):
-            yield self._markdown_answer()
-        else:
-            yield self._default_answer(prompt, sources)
+    def answer(self, messages: list[Message]) -> Iterator[str]:
+        yield self._default_answer(messages)
 
     def _markdown_answer(self) -> str:
         return textwrap.dedent(
@@ -39,16 +35,19 @@ class RagnaDemoAssistant(Assistant):
             """
         ).strip()
 
-    def _default_answer(self, prompt: str, sources: list[Source]) -> str:
+    def _default_answer(self, messages: list[Message]) -> str:
+        prompt = messages[0].content.strip()
         sources_display = []
-        for source in sources:
-            source_display = f"- {source.document.name}"
-            if source.location:
-                source_display += f", {source.location}"
-            source_display += f": {textwrap.shorten(source.content, width=100)}"
-            sources_display.append(source_display)
-        if len(sources) > 3:
-            sources_display.append("[...]")
+        for message in messages:
+            sources = message.sources
+            for source in sources:
+                source_display = f"- {source.document.name}"
+                if source.location:
+                    source_display += f", {source.location}"
+                source_display += f": {textwrap.shorten(source.content, width=100)}"
+                sources_display.append(source_display)
+            if len(sources) > 3:
+                sources_display.append("[...]")
 
         return (
             textwrap.dedent(
--- a/ragna/assistants/_demo.py
+++ b/ragna/assistants/_demo.py
@@ -24,7 +24,8 @@ class RagnaDemoAssistant(Assistant):
     def answer(self, messages: list[Message]) -> Iterator[str]:
         if "markdown" in messages[-1].content.lower():
             yield self._markdown_answer()
-        yield self._default_answer(messages)
+        else:
+            yield self._default_answer(messages)
 
     def _markdown_answer(self) -> str:
         return textwrap.dedent(
--- a/ragna/assistants/_demo.py
+++ b/ragna/assistants/_demo.py
@@ -22,6 +22,8 @@ class RagnaDemoAssistant(Assistant):
         return "Ragna/DemoAssistant"
 
     def answer(self, messages: list[Message]) -> Iterator[str]:
+        if "markdown" in messages[-1].content.lower():
+            yield self._markdown_answer()
         yield self._default_answer(messages)
 
     def _markdown_answer(self) -> str:
--- a/ragna/assistants/_demo.py
+++ b/ragna/assistants/_demo.py
@@ -1,7 +1,7 @@
 import textwrap
 from typing import Iterator
 
-from ragna.core import Assistant, Message
+from ragna.core import Assistant, Message, MessageRole
 
 
 class RagnaDemoAssistant(Assistant):
@@ -38,24 +38,26 @@ class RagnaDemoAssistant(Assistant):
     def _default_answer(self, messages: list[Message]) -> str:
         prompt = messages[-1].content.strip()
         sources_display = []
-        for message in messages:
-            sources = message.sources
-            for source in sources:
-                source_display = f"- {source.document.name}"
-                if source.location:
-                    source_display += f", {source.location}"
-                source_display += f": {textwrap.shorten(source.content, width=100)}"
-                sources_display.append(source_display)
-            if len(sources) > 3:
-                sources_display.append("[...]")
+        sources = messages[-1].sources
+        for source in sources:
+            source_display = f"- {source.document.name}"
+            if source.location:
+                source_display += f", {source.location}"
+            source_display += f": {textwrap.shorten(source.content, width=100)}"
+            sources_display.append(source_display)
+        if len(sources) > 3:
+            sources_display.append("[...]")
 
+        n_messages = len([m for m in messages if m.role == MessageRole.USER])
         return (
             textwrap.dedent(
                 """
-                I'm a demo assistant and can be used to try Ragnas workflow.
+                I'm a demo assistant and can be used to try Ragna's workflow.
                 I will only mirror back my inputs. 
+
+                So far I have received {n_messages} messages.
                 
-                Your prompt was:
+                Your last prompt was:
 
                 > {prompt}
 
@@ -65,5 +67,10 @@ class RagnaDemoAssistant(Assistant):
                 """
             )
             .strip()
-            .format(name=str(self), prompt=prompt, sources="\n".join(sources_display))
+            .format(
+                name=str(self),
+                n_messages=n_messages,
+                prompt=prompt,
+                sources="\n".join(sources_display),
+            )
         )
--- a/ragna/assistants/_openai.py
+++ b/ragna/assistants/_openai.py
@@ -2,7 +2,7 @@ import abc
 from functools import cached_property
 from typing import Any, AsyncIterator, Optional, cast
 
-from ragna.core import Source
+from ragna.core import Message, MessageRole
 
 from ._http_api import HttpApiAssistant, HttpStreamingProtocol
 
@@ -14,17 +14,43 @@ class OpenaiLikeHttpApiAssistant(HttpApiAssistant):
     @abc.abstractmethod
     def _url(self) -> str: ...
 
-    def _make_system_content(self, sources: list[Source]) -> str:
+    # TODO: move to user config
+    def _make_system_content(self) -> str:
         # See https://github.com/openai/openai-cookbook/blob/main/examples/How_to_format_inputs_to_ChatGPT_models.ipynb
         instruction = (
-            "You are an helpful assistants that answers user questions given the context below. "
+            "You are a helpful assistant that answers user questions given the context below. "
             "If you don't know the answer, just say so. Don't try to make up an answer. "
-            "Only use the sources below to generate the answer."
+            "Only use the included messages below to generate the answer."
         )
-        return instruction + "\n\n".join(source.content for source in sources)
+
+        return Message(
+            content=instruction,
+            role=MessageRole.SYSTEM,
+        )
+
+    def _format_message_sources(self, messages: list[Message]) -> str:
+        sources_prompt = "Include the following sources in your answer:"
+        formatted_messages = []
+        for message in messages:
+            if message.role == MessageRole.USER:
+                formatted_messages.append(
+                    {
+                        "content": sources_prompt
+                        + "\n\n".join(source.content for source in message.sources),
+                        "role": MessageRole.SYSTEM,
+                    }
+                )
+
+            formatted_messages.append(
+                {"content": message.content, "role": message.role}
+            )
+        return formatted_messages
 
     def _stream(
-        self, prompt: str, sources: list[Source], *, max_new_tokens: int
+        self,
+        messages: list[dict],
+        *,
+        max_new_tokens: int,
     ) -> AsyncIterator[dict[str, Any]]:
         # See https://platform.openai.com/docs/api-reference/chat/create
         # and https://platform.openai.com/docs/api-reference/chat/streaming
@@ -35,16 +61,7 @@ class OpenaiLikeHttpApiAssistant(HttpApiAssistant):
             headers["Authorization"] = f"Bearer {self._api_key}"
 
         json_ = {
-            "messages": [
-                {
-                    "role": "system",
-                    "content": self._make_system_content(sources),
-                },
-                {
-                    "role": "user",
-                    "content": prompt,
-                },
-            ],
+            "messages": messages,
             "temperature": 0.0,
             "max_tokens": max_new_tokens,
             "stream": True,
@@ -55,9 +72,15 @@ class OpenaiLikeHttpApiAssistant(HttpApiAssistant):
         return self._call_api("POST", self._url, headers=headers, json=json_)
 
     async def answer(
-        self, prompt: str, sources: list[Source], *, max_new_tokens: int = 256
+        self,
+        messages: list[Message] = [],
+        *,
+        max_new_tokens: int = 256,
     ) -> AsyncIterator[str]:
-        async for data in self._stream(prompt, sources, max_new_tokens=max_new_tokens):
+        formatted_messages = self._format_message_sources(messages)
+        async for data in self._stream(
+            formatted_messages, max_new_tokens=max_new_tokens
+        ):
             choice = data["choices"][0]
             if choice["finish_reason"] is not None:
                 break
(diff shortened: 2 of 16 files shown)


ragna/assistants/_demo.py
```diff
@@ -22,11 +21,8 @@ class RagnaDemoAssistant(Assistant):
     def display_name(cls) -> str:
         return "Ragna/DemoAssistant"
 
-    def answer(self, prompt: str, sources: list[Source]) -> Iterator[str]:
-        if re.search("markdown", prompt, re.IGNORECASE):
-            yield self._markdown_answer()
-        else:
-            yield self._default_answer(prompt, sources)
+    def answer(self, messages: list[Message]) -> Iterator[str]:
```
blakerosenthal, line 24: TEMP: will put this back

ragna/assistants/_demo.py
```diff
@@ -39,16 +35,19 @@ def _markdown_answer(self) -> str:
             """
         ).strip()
 
-    def _default_answer(self, prompt: str, sources: list[Source]) -> str:
+    def _default_answer(self, messages: list[Message]) -> str:
+        prompt = messages[-1].content.strip()
```
blakerosenthal, line 39: Either need to trust the order of `messages` or make sure `core.Message` inherits timestamp from `schema.Message`

ragna/assistants/_openai.py
```diff
@@ -14,17 +14,43 @@ class OpenaiLikeHttpApiAssistant(HttpApiAssistant):
...
+            "Only use the included messages below to generate the answer."
         )
-        return instruction + "\n\n".join(source.content for source in sources)
+
+        return Message(
+            content=instruction,
+            role=MessageRole.SYSTEM,
+        )
+
+    def _format_message_sources(self, messages: list[Message]) -> str:
```
blakerosenthal, line 31: Example of what this could look like for OpenAI-like assistants. There could also be an attribute on the Assistant that tells this function how to truncate the message list based on the size of the target LLM's context window.

ragna/core/_rag.py
```diff
@@ -195,6 +195,9 @@ async def prepare(self) -> Message:
         await self._run(self.source_storage.store, self.documents)
         self._prepared = True
 
+        system_prompt = self.assistant._make_system_content()
```
blakerosenthal, line 198: It feels easiest to include the basic system prompt here, but maybe it should be regenerated and prepended to the list of messages for every call to the LLM and not stored here. This would make the most sense if the Assistant might truncate the message list and chop off the system prompt.

ragna/core/_rag.py
```diff
@@ -220,17 +223,20 @@ async def answer(self, prompt: str, *, stream: bool = False) -> Message:
...
+        question = Message(content=prompt, role=MessageRole.USER, sources=sources)
+        self._messages.append(question)
+
         answer = Message(
-            content=self._run_gen(self.assistant.answer, prompt, sources),
+            content=self._run_gen(self.assistant.answer, self._messages),
             role=MessageRole.ASSISTANT,
             sources=sources,
         )
-        if not stream:
```
blakerosenthal, line 232: I was struggling to get streaming to work with the list of messages. Either I would get an error like this when the answer's Message content is still an AsyncIterable
```
    | Traceback (most recent call last):
    |   File "/home/blake/miniconda3/envs/ragna-dev/lib/python3.9/site-packages/starlette/responses.py", line 261, in wrap
    |     await func()
    |   File "/home/blake/miniconda3/envs/ragna-dev/lib/python3.9/site-packages/starlette/responses.py", line 250, in stream_response
    |     async for chunk in self.body_iterator:
    |   File "/home/blake/code/ragna/ragna/deploy/_api/core.py", line 328, in to_jsonl
    |     async for model in models:
    |   File "/home/blake/code/ragna/ragna/deploy/_api/core.py", line 302, in message_chunks
    |     content_chunk = await anext(core_answer_stream)
    |   File "/home/blake/code/ragna/ragna/_compat.py", line 85, in anext_impl
    |     return await ait.__anext__()
    |   File "/home/blake/code/ragna/ragna/core/_components.py", line 202, in __aiter__
    |     async for chunk in self._content_stream:
    |   File "/home/blake/code/ragna/ragna/core/_rag.py", line 403, in _run_gen
    |     async for item in async_gen:
    |   File "/home/blake/miniconda3/envs/ragna-dev/lib/python3.9/site-packages/starlette/concurrency.py", line 65, in iterate_in_threadpool
    |     yield await anyio.to_thread.run_sync(_next, as_iterator)
    |   File "/home/blake/miniconda3/envs/ragna-dev/lib/python3.9/site-packages/anyio/to_thread.py", line 56, in run_sync
    |     return await get_async_backend().run_sync_in_worker_thread(
    |   File "/home/blake/miniconda3/envs/ragna-dev/lib/python3.9/site-packages/anyio/_backends/_asyncio.py", line 2177, in run_sync_in_worker_thread
    |     return await future
    |   File "/home/blake/miniconda3/envs/ragna-dev/lib/python3.9/site-packages/anyio/_backends/_asyncio.py", line 859, in run
    |     result = context.run(func, *args)
    |   File "/home/blake/miniconda3/envs/ragna-dev/lib/python3.9/site-packages/starlette/concurrency.py", line 54, in _next
    |     return next(iterator)
    |   File "/home/blake/code/ragna/ragna/assistants/_demo.py", line 25, in answer
    |     yield self._default_answer(messages)
    |   File "/home/blake/code/ragna/ragna/assistants/_demo.py", line 39, in _default_answer
    |     prompt = messages[-1].content.strip()
    |   File "/home/blake/code/ragna/ragna/core/_components.py", line 219, in content
    |     raise RuntimeError(
    | RuntimeError: Message content cannot be accessed without having iterated over it, e.g. `async for chunk in message`, or reading the content, e.g. `await message.read()`, first.
```
or I anger the async gods further by trying to convert the message `.content` into a string by iterating over it in advance:
```
   | Traceback (most recent call last):
    |   File "/home/blake/miniconda3/envs/ragna-dev/lib/python3.9/site-packages/starlette/responses.py", line 261, in wrap
    |     await func()
    |   File "/home/blake/miniconda3/envs/ragna-dev/lib/python3.9/site-packages/starlette/responses.py", line 250, in stream_response
    |     async for chunk in self.body_iterator:
    |   File "/home/blake/code/ragna/ragna/deploy/_api/core.py", line 328, in to_jsonl
    |     async for model in models:
    |   File "/home/blake/code/ragna/ragna/deploy/_api/core.py", line 302, in message_chunks
    |     content_chunk = await anext(core_answer_stream)
    |   File "/home/blake/code/ragna/ragna/_compat.py", line 85, in anext_impl
    |     return await ait.__anext__()
    |   File "/home/blake/code/ragna/ragna/core/_components.py", line 202, in __aiter__
    |     async for chunk in self._content_stream:
    |   File "/home/blake/code/ragna/ragna/core/_rag.py", line 403, in _run_gen
    |     async for item in async_gen:
    |   File "/home/blake/code/ragna/ragna/assistants/_openai.py", line 81, in answer
    |     formatted_messages = await self._format_message_sources(messages)
    |   File "/home/blake/code/ragna/ragna/assistants/_openai.py", line 35, in _format_message_sources
    |     await message.read()
    |   File "/home/blake/code/ragna/ragna/core/_components.py", line 212, in read
    |     async for _ in self:
    |   File "/home/blake/code/ragna/ragna/core/_components.py", line 202, in __aiter__
    |     async for chunk in self._content_stream:
    | RuntimeError: anext(): asynchronous generator is already running
```

ragna/assistants/_demo.py
pmeier: We can trust the order of the messages. When using only the Python API, we are just appending messages to a list. As for restoring from DB, we explicitly sort them

https://github.com/Quansight/ragna/blob/9776ec5b06b573ee0b672e07e23c13cd7e0e5fb3/ragna/deploy/_api/orm.py#L89-L91

ragna/assistants/_demo.py
```diff
@@ -39,16 +35,19 @@ def _markdown_answer(self) -> str:
...
         sources_display = []
-        for source in sources:
-            source_display = f"- {source.document.name}"
-            if source.location:
-                source_display += f", {source.location}"
-            source_display += f": {textwrap.shorten(source.content, width=100)}"
-            sources_display.append(source_display)
-        if len(sources) > 3:
-            sources_display.append("[...]")
+        for message in messages:
```
pmeier, line 41: Let's not overwhelm the user with information when using this demo assistant. Not sure if we even want to include any of the old messages. Maybe just add a small statistic like "before this question you asked N more".

ragna/core/_rag.py
```diff
@@ -220,17 +223,20 @@ async def answer(self, prompt: str, *, stream: bool = False) -> Message:
                 detail=RagnaException.EVENT,
             )
 
-        self._messages.append(Message(content=prompt, role=MessageRole.USER))
-
         sources = await self._run(self.source_storage.retrieve, self.documents, prompt)
 
+        question = Message(content=prompt, role=MessageRole.USER, sources=sources)
```
pmeier, line 228: Until this, we only ever attached the sources to the answer and not the question. I'm open to a "philosophical debate" on where they should go, but I don't want them on the question and the answer. Why not just keep passing the sources to `Assistant.answer`?

ragna/core/_rag.py
pmeier: Changing the signature of `Assistant.answer` should have no effect. Could you please post a snippet to reproduce this?

ragna/core/_components.py
```diff
@@ -237,13 +239,25 @@ class Assistant(Component, abc.ABC):
...
 
+    def _make_system_content(self):
+        return Message(
+            content=(
+                "You are a helpful assistant that answers user questions given the context below. "
+                "If you don't know the answer, just say so. Don't try to make up an answer. "
+                "Only use the included messages below to generate the answer."
+            ),
+            role=MessageRole.SYSTEM,
+        )
```
pmeier, line 250: Careful about the terminology here. What you have here is a "system prompt", i.e. instructions for an LLM. A "system message" in Ragna is something like a welcome message or errors, i.e. something that the system (Ragna) sends to the user instead of the assistant.

Meaning, wrapping the system prompt into a system message makes no sense.

ragna/core/_rag.py
pmeier: I don't understand why the chat would need the system prompt of the assistant. Instead of creating it here and pass it to the Assistant as part of the messages, why not let the Assistant create it in the first place?

ragna/core/_components.py
pmeier: I also don't understand why this is here. Given that there is no standard for a system prompt, why are we putting this on the base class?

ragna/assistants/_demo.py
blakerosenthal: Great, that makes it easy

ragna/core/_rag.py
blakerosenthal: I think I agree. I also mostly talked myself out of putting it here as well. I think I wanted `_messages` to be a complete historical list of messages that could be sent to the LLM, but system prompts can stay hard coded on the Assistant (or better yet, in a config somewhere?)

ragna/core/_rag.py
blakerosenthal: The issue is with trying to use `message.content` in `Assistant._format_message_sources`. Some of the Messages in the message list are still `AsyncIterable`s. If I try to call `.read()` on a message within the assistant I get
```
RuntimeError: anext(): asynchronous generator is already running
```
Here's an example of that:
```python
def Assistant:
    ...
    async def _format_message_sources(self, messages: list[Message]) -> str:
        sources_prompt = "Include the following sources in your answer:"
        formatted_messages = []
        for message in messages:
            content = await message.read()
            if message.role == MessageRole.USER:
                formatted_messages.append(
                    {
                        "content": sources_prompt
                        + "\n\n".join(source.content for source in message.sources),
                        "role": MessageRole.SYSTEM,
                    }
                )

            formatted_messages.append({"content": content, "role": message.role})
        return formatted_messages

    async def answer(
        self,
        messages: list[Message] = [],
        *,
        max_new_tokens: int = 256,
    ) -> AsyncIterator[str]:
        formatted_messages = await self._format_message_sources(messages)
        async for data in self._stream(
            formatted_messages, max_new_tokens=max_new_tokens
        ):
            choice = data["choices"][0]
            if choice["finish_reason"] is not None:
                break

            yield cast(str, choice["delta"]["content"])
```
Then running `chat.answer` will produce the error when streaming is true. I just pushed a commit that demos this behavior.

ragna/assistants/_demo.py
blakerosenthal: I like that a lot. I can make that change.

ragna/core/_rag.py
blakerosenthal: This was something I meant to ask about in my original list of questions last night but I ran out of steam. I'm really curious about this from a design perspective. Let's imagine a design where `chat._messages` is a list of `Message` objects, as currently implemented. Now also say the assistant is responsible for formatting each raw message for the LLM API call. I can imagine at least a couple scenarios for how we would format historical messages:
1. The assistant reformats the message list every time it receives a new `.answer` call. So if we want to split up a prompt/sources pair into two separate json objects (i.e. `{'content': '<user question>', 'role': 'user'}` + `{'content': 'please use only the following sources: ...', 'role': 'system'}` we either need the sources to be attached to the question or we need to reach forward in the messages list to grab the sources from the answer. Something feels a little strange about the latter, but I agree that sources shouldn't be on both the question and the answer, just one or the other.
2. Alternatively, rather than reformat the LLM message JSON every time, we only format new prompt/source pairs and maintain a list of message dictionaries or rendered json somewhere of previously formatted messages. The we'd just need to append the new messages and send them along. This then becomes another data structure to maintain.

ragna/core/_components.py
blakerosenthal: Fully agree. This shouldn't be here. Originally I was calling this from the `Chat` object, in which case we'd need at least a stub that throws a `NotImplemented` error on the Assistant class, but after your comment above and just thinking about it a bit more it should just be implemented on a per-assistant basis, if needed. And beyond that, I was confusing the purpose of a system message in the first place. There's no point wrapping this in a `Message` at all.

ragna/core/_rag.py
dillonroach: Just be sure there's no attempt at a 'universal' system prompt anywhere - on the assistant we'll have a general one for .answer(), but e.g. we'll want different system prompts for things like all the pre-process steps or anything that's outside the standard chain. Simply to say, it's reasonable to have a standard system prompt on .answer() but we don't want it fixed on the general llm calls in .generate() (wip).

ragna/assistants/_openai.py
```diff
@@ -14,17 +14,43 @@ class OpenaiLikeHttpApiAssistant(HttpApiAssistant):
...
+        sources_prompt = "Include the following sources in your answer:"
```
dillonroach, line 32: Depending on what order we try to merge PRs - this piece will eventually be overwritten by the notion of the preprocess stage. It's probably fine for this version imo.

ragna/core/_rag.py
dillonroach: Recall, we eventually want ['chat history' messages]+prompt->preprocess->retrieval so the messages coming in the history before the current prompt are(were) all strings to the LLM, but we may want to be able to act upon them with the distinction between content and source - e.g. if messages from the user are tracked at a 'content' + sources level, then it might be easier to process the sources down in later stages - e.g. you may not want to carry content+N-chunks forward from every user prompt.

Keeping the sources attached to prompts would be particularly useful in preprocess tracking where a given user-role message may be (content+sources) and you want to streamline the context there (parse out/summarize/select sources).
Just keep in mind that 'previous' messages make sense to have the sources attached, while the 'current' prompt we'll still be sending through processing and retrieval.

If I'm thinking about a pre-process stage class, receiving 
`messages = [{system,...}, {user,.. content, sources}, {assistant,.. answer}...]` makes more sense than the source being with the answer - the 'content' in all previous user messages would contain the sources in that string as well (minus current prompt, which would not yet have sources).

ragna/core/_rag.py
pmeier: Without having it reproduced, I think the error comes from the fact that you want to use the old assistant message to generate a new one, but have never accessed it. This is not a standard use case as we can safely assume that a user would consume an answer before asking a follow-up question. And after a message once has been `.read()`, you can safely access `.content`.

blakerosenthal: That makes sense to me and is how I'd expect this to behave. I experience this problem after asking a single question, so it's a mystery why `messages` contains an unread answer ever. I'll dig a little more.

ragna/core/_rag.py
blakerosenthal: > the 'content' in all previous user messages would contain the sources in that string as well (minus current prompt, which would not yet have sources).

So you're thinking of the stage of the pipeline where we have the most recent user prompt but have not yet retrieved any sources from source storage, is that right? Would you want `.content` to contain string-formatted sources from previous questions, or would it be more flexible to still have them as `Source` objects?

blakerosenthal:
@dillonroach I haven't updated any assistants other than the ones that inherit from `OpenaiLikeHttpApiAssistant` and the demo. Do you anticipate [m]any merge issues with your work if I went ahead and did that / do you have any other thoughts on how to pass multiple messages to the other assistants? I was thinking it would be a good idea to have a message list length limit but you are probably the one to know what that should be.

dillonroach:
> @dillonroach I haven't updated any assistants other than the ones that inherit from `OpenaiLikeHttpApiAssistant` and the demo. Do you anticipate [m]any merge issues with your work if I went ahead and did that / do you have any other thoughts on how to pass multiple messages to the other assistants?

I would intentionally stay hands-off from your PR on the assistants, I'll be overhauling that anyway (a reference implement in one or two places doesn't hurt anything, but I'll likely just use it as ref and tear it up, so don't be too worried about details).

> I was thinking it would be a good idea to have a message list length limit but you are probably the one to know what that should be.

That's on the assistants to handle and there's nothing motivating it beyond the context window of the LLM - each assistant should have an idea of a max sequence length they can handle, and if the length coming in is too long, they'll basically need to drop the first n messages to account for the issue. E.g. (assuming message 1 is system prompt) (pseudo) messages[1] + messages[1+n:] - ideally, here, you'd do a token-count for the messages and anticipate the max length your generation will take, such that all messages+upcoming gen can fit within the llm context window.

blakerosenthal:
> I would intentionally stay hands-off from your PR on the assistants

@dillonroach Do you want to just merge this branch into yours and make a mega-PR, or would you prefer I coordinate with Philip to get mine merged into `main` first?

tests/deploy/api/test_e2e.py
```diff
@@ -6,8 +6,10 @@
 from ragna.deploy import Config
 from ragna.deploy._api import app
 from tests.deploy.utils import TestAssistant, authenticate_with_api
+from tests.utils import skip_on_windows
 
 
+@skip_on_windows
```
pmeier, line 12: What is the issue here? We shouldn't just skip our E2E test unless we are sure that this is testing issue on Windows.

pmeier (approved):
Thanks Blake! I pushed a minor cleanup commit (d7fbccbe492031a912873842b843ce93bc4a9666) to avoid some back and forth.

tests/deploy/api/test_e2e.py
pmeier: The [failure](https://github.com/Quansight/ragna/actions/runs/9891178695/job/27321052554?pr=438#step:4:169) looks real and serious:

```
Differing items:
  {'role': 'user'} != {'role': 'assistant'}
  {'id': 'a5a9ab44-2571-4dd1-889c-1ea8c61e6072'} != {'id': '0de4bf10-7e1c-4e54-8b8d-2aa760868335'}
  {'content': '?'} != {'content': "I'm a demo assistant and can be used to try Ragna's workflow.\nI will only mirror back my inputs. \n\nSo far I have received 1 messages.\n\nYour last prompt was:\n\n> ?\n\nThese are the sources I was given:\n\n- test.txt: ! "}
```

blakerosenthal: Agreed, there's something funky going on with transaction processing on Windows but it's intermittent (sometimes the GH tests pass and sometimes they don't). I'll try to get to the bottom of this.

blakerosenthal: Opened https://github.com/Quansight/ragna/issues/447

PR #443, opened 2024-07-08
use backend engine in UI

### LINKED ISSUE #392: Standalone web UI

### LINKED ISSUE #406: Add UI support for batch document upload
### Feature description

Currently batch document upload (#404) is only available via the REST API. We should also support this with the UI. 

### Value and/or benefit

It will allow faster upload of documents with the UI for deployment where there is non-negligible latency associated with calls to the database.

### Anything else?

_No response_

pmeier:
Follow-up to #441 and closes #392, closes #406. This leaves our UI code in a hot mess. But I thought for reviewing it will be easier to just drop in the engine where we previously used the API instead of starting yet another larger refactoring. However, this has to happen eventually and after this PR I'd prefer sooner than later.

After this PR we only have the auth rework left, that I need to finish before we can publish a first release of `bokeh-fastapi`. Meaning, unless you have some hard blocking comments, I'd prefer to leave larger stuff for follow-up PRs. If you find something, flag it and I'll put up an issue.


blakerosenthal (approved):
This is so much nicer. The ApiWrapper class feels slightly misnamed now though.

PR #444, opened 2024-07-08
update assistants in docker config

pmeier:


blakerosenthal (approved):
LGTM. What's the reason for not including the demo assistant?

pmeier:
That's a fair question. It was removed as part of #333. TBH, I don't remember my reasoning and I didn't write it down either.

I guess we should add it back so `docker run ragna` will give you a functioning experience albeit not useful for actual Q/A. We don't have to do the same for the source storages, because both Chroma and LanceDB are available.

PR #445, opened 2024-07-08
#391 Chat Ordering based upon the order of interaction

### LINKED ISSUE #391: Chat Ordering Newest to Oldest
### Feature description

Currently chats are listed oldest first (highest) in `ragna ui`.  It seems more helpful more most use cases to reverse that order and have newer chats appear higher in the chat list in the side bar of ragna ui.

### Value and/or benefit

Better UX

### Anything else?

_No response_

arjxn-py:
This PR modifies the `LeftSidebar` component to display chats in reverse order, showing the newest chats first instead of the oldest.
Should fix #391

### DIFF
--- a/ragna/deploy/_ui/left_sidebar.py
+++ b/ragna/deploy/_ui/left_sidebar.py
@@ -60,8 +60,7 @@ class LeftSidebar(pn.viewable.Viewer):
     @pn.depends("refresh_counter", "chats", "current_chat_id", on_init=True)
     def __panel__(self):
         self.chats.sort(
-            key=lambda chat: chat['messages'][-1]['timestamp'],
-            reverse=True
+            key=lambda chat: chat['messages'][-1]['timestamp'], reverse=True
         )
 
         self.chat_buttons = []
--- a/ragna/deploy/_ui/left_sidebar.py
+++ b/ragna/deploy/_ui/left_sidebar.py
@@ -1,3 +1,5 @@
+from datetime import datetime
+
 import panel as pn
 import param
 
@@ -59,13 +61,12 @@ class LeftSidebar(pn.viewable.Viewer):
 
     @pn.depends("refresh_counter", "chats", "current_chat_id", on_init=True)
     def __panel__(self):
-        from datetime import datetime
         epoch = datetime(1970, 1, 1)
         self.chats.sort(
             key=lambda chat: (
                 epoch if not chat["messages"] else chat["messages"][-1]["timestamp"]
             ),
-            reverse=True
+            reverse=True,
         )
 
         self.chat_buttons = []
--- a/ragna/deploy/_ui/left_sidebar.py
+++ b/ragna/deploy/_ui/left_sidebar.py
@@ -60,7 +60,7 @@ class LeftSidebar(pn.viewable.Viewer):
     @pn.depends("refresh_counter", "chats", "current_chat_id", on_init=True)
     def __panel__(self):
         self.chats.sort(
-            key=lambda chat: chat['messages'][-1]['timestamp'], reverse=True
+            key=lambda chat: chat["messages"][-1]["timestamp"], reverse=True
         )
 
         self.chat_buttons = []
--- a/ragna/deploy/_ui/left_sidebar.py
+++ b/ragna/deploy/_ui/left_sidebar.py
@@ -59,8 +59,13 @@ class LeftSidebar(pn.viewable.Viewer):
 
     @pn.depends("refresh_counter", "chats", "current_chat_id", on_init=True)
     def __panel__(self):
+        from datetime import datetime
+        epoch = datetime(1970, 1, 1)
         self.chats.sort(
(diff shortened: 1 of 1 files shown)


pmeier:
@blakerosenthal Could you check if this change is indeed enough?

blakerosenthal:
This doesn't quite work, unfortunately. `self.chats` comes indirectly from [database.py:get_chats](https://github.com/arjxn-py/ragna/blob/00d9c028ee141c55d9c9a1413155c9fece3c6557/ragna/deploy/_api/database.py#L175) which doesn't guarantee order. There's no timestamp on `orm.Chat` either otherwise we could sort in `get_chats`. 

Option 1: add a (created_at) timestamp to Chats and then add an `order_by` to the table.
Option 2: dig into the `messages` and pull out either the first or last message timestamp. I kind of like this option because then the UI could show chats by how recently they were used, not when they were created. This could be accomplished with a `last_updated` field on the chat table as well.

image

pmeier:
I'd prefer option 1. by a big margin.

arjxn-py:
To be honest I also like the latter (option 2) more as I feel it contains more utility for the user.

arjxn-py:
Hi @blakerosenthal, happy to have your quick look again on it. Thanks 🙂

ragna/deploy/_ui/left_sidebar.py
```diff
@@ -59,6 +59,10 @@ def refresh(self):
 
     @pn.depends("refresh_counter", "chats", "current_chat_id", on_init=True)
     def __panel__(self):
+        self.chats.sort(
+            key=lambda chat: chat["messages"][-1]["timestamp"], reverse=True
+        )
```
blakerosenthal, line 64: Just need to be careful with unprepared chats that don't have any messages yet. Something like this?
```suggestion
        from datetime import datetime
        epoch = datetime(1970, 1, 1)
        self.chats.sort(
            key=lambda chat: (
                epoch if not chat["messages"] else chat["messages"][-1]["timestamp"]
            ),
            reverse=True
        )
```

blakerosenthal: unless unprepared chats should go at the top?
```suggestion
        from datetime import datetime
        self.chats.sort(
            key=lambda chat: (
                datetime.now() if not chat["messages"] else chat["messages"][-1]["timestamp"]
            ),
            reverse=True
        )
```

arjxn-py: Thanks for the quick response & the suggestion really does makes sense so happy to commit it.
However i have one small doubt before that in my case, Every time I create a new chat it already has one message i.e. *How can I help you with the documents?* so the situation of chat with no messages didn't actually arise. 
Is there a way that I can create a chat with no message i.e. unprepared chat (I hope I'm referring it right)

pmeier: Having a chat with no messages, i.e. an unprepared one, can only arise if the chat is created through the API. For that, this is a two-step process, i.e. the user has to send two HTTP requests, e.g.

https://github.com/Quansight/ragna/blob/55f7fc5d3d13f321eebd16d3b0901ab35ac36bc7/scripts/add_chats.py#L51-L66

Of course this is the same in the UI, but there we can guarantee that they are performed back to back, without ever leaving the user in an intermediate state

https://github.com/Quansight/ragna/blob/55f7fc5d3d13f321eebd16d3b0901ab35ac36bc7/ragna/deploy/_ui/api_wrapper.py#L83-L98

Meaning, you'll never be able to create the issue @blakerosenthal mentioned with just the UI.

However, we also don't have a mechanism to deal with such chats in the UI yet. Meaning, if we just create a chat with the API and start the UI afterwards, you will be seeing an empty chat that you cannot prepare, but that will error when you send any prompt, because it is not prepared.

So I guess it makes sense to just move these chats at the end of the list. Ultimately, we should add functionality that triggers the preparation if at any point a user selects an unprepared chat. @arjxn-py Do you want to open an issue for that?

arjxn-py: > Do you want to open an issue for that?

Yes, I just opened #448, please correct me if needed.

> So I guess it makes sense to just move these chats at the end of the list.

Happy to merge @blakerosenthal's first suggestion to handle unprepared chats this way, until they are clicked on `ui` & prepared automatically.

pmeier:
I'm ok with limiting the scope of this change to only adapt the order when the page (re-)loaded. But ultimately, we want the ordering change to be dynamic, e.g. if I ask a new question in an old chat, that should automatically move to the top of the list rather than this only happening the next time I load the page. @arjxn-py or @blakerosenthal could one of you open a new issue for this?

pmeier (approved):
LGTM, thanks Arjun!

arjxn-py:
> could one of you open a new issue for this?

Thanks @pmeier, opened #457 for this, feel free to correct me there if needed.

PR #446, opened 2024-07-09
#211 Enable users to upload folders with `file_uploader` component in…

### LINKED ISSUE #211: Add support to select folders in web UI
If I have all the documents that I have stored in one folder, it would be quite convenient to just select it rather than opening it and selecting all files in it. This should not work recursively, meaning only documents directly in the folder I selected should be used rather than more in potential subfolders.

arjxn-py:
This PR enhances the `FileUploader` component by adding functionality to allow users to select entire folders for upload. This uses the `webkitdirectory` attribute in the file input element in order to enable users to select folders too & parse all files inside at once.
Should fix #211

### DIFF
--- a/ragna/deploy/_ui/components/file_uploader.py
+++ b/ragna/deploy/_ui/components/file_uploader.py
@@ -153,6 +153,7 @@ class FileUploader(ReactiveHTML, Widget):  # type: ignore[misc]
                             multiple
                             onchange="${script('file_input_on_change')}" 
                             accept="${allowed_documents_str}"
+                            webkitdirectory
                     /> 
                 </div>
                 <div id="fileListContainer" class="fileListContainer">


arjxn-py:
Folders are being uploaded correctly but `webkitdirectory` is avoiding files from being accepted.

arjxn-py:
> SO I guess there is no way to do both? And if so, we need to prioritize file upload over directory upload. If you can confirm this, we can update our issue to low prio, because it would require a custom component with a lot of logic. And I'm not even sure we want to maintain that even if we can get it to work.

Yes I confirm that there's no way to do both with a single component right now. But one workaround could be that we can have two component side by side i.e. `file_uploader` & `folder_uploader`. We can have a logic that if a user have uploaded either or both of them then he can create the chat.

pmeier:
@arjxn-py In the light of the discovery that we cannot upload individual files and folders at the same time, I have closed #211 and in turn will also close this PR. Still, thanks for investigation!

PR #459, opened 2024-07-22
#458 Automate Docker Requirements Workflow for PR Creation and Review Requests

### LINKED ISSUE #458: Automating Workflow for Docker Requirement Fixes
Currently, the `docker-requirements / update (pull_request)` workflow is configured to be triggered manually. When activated, it opens a separate PR to fix Docker requirements. While this ensures updates are controlled, it requires manual intervention, which can delay necessary fixes.

Maybe we can configure the workflow to be automatically triggered on every push to the main branch, creating a PR if Docker requirements need fixing. cc: @pmeier

arjxn-py:
Should Fix #458 

This PR updates the Docker requirements workflow to:

- Automatically create PRs to update requirements-docker.lock when changes are detected.
- Dynamically request reviews from the author and the person who merged the last PR into main.

I've also tested this updated workflow a number of times on my fork to- 
- Here's the [test workflow run](https://github.com/arjxn-py/ragna/actions/runs/10049115488/job/27774541367#step:6:1)
- Here's [the automatically created PR](https://github.com/arjxn-py/ragna/pull/6) when I bumped down `panel==1.4.2` on my fork for testing.

### DIFF
--- a/.github/workflows/update-docker-requirements.yml
+++ b/.github/workflows/update-docker-requirements.yml
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - main
+  workflow_dispatch:
 
 permissions:
   pull-requests: write
--- a/.github/workflows/update-docker-requirements.yml
+++ b/.github/workflows/update-docker-requirements.yml
@@ -4,6 +4,10 @@ on:
   push:
     branches:
       - main
+    paths:
+      - ".github/workflows/update-docker-requirements.yml"
+      - "pyproject.toml"
+      - "requirements-docker.lock"
   workflow_dispatch:
 
 permissions:
@@ -44,18 +48,16 @@ jobs:
           echo "needs-update=${NEEDS_UPDATE}" | tee --append $GITHUB_OUTPUT
 
       - name: Get last merged PR info
-        id: pr_info
+        id: pr-info
         env:
           GITHUB_TOKEN: ${{ secrets.GH_TOKEN }}
         run: |
           LAST_MERGED_PR=$(gh pr list --repo $GITHUB_REPOSITORY --state closed --json number --jq '.[0].number')
           echo "Last merged PR number: $LAST_MERGED_PR"
-          PR_INFO=$(gh pr view $LAST_MERGED_PR --repo $GITHUB_REPOSITORY --json author,mergedBy --jq '{author: .author.login, mergedBy: .mergedBy.login}')
+          PR_INFO=$(gh pr view $LAST_MERGED_PR --repo $GITHUB_REPOSITORY --json author,mergedBy --jq '{author: .author.login, merger: .mergedBy.login}')
           echo "PR Info: $PR_INFO"
-          echo "pr_author=$(echo $PR_INFO | jq -r .author)" >> $GITHUB_ENV
-          echo "pr_mergedBy=$(echo $PR_INFO | jq -r .mergedBy)" >> $GITHUB_ENV
-          echo "PR Author: $(echo $PR_INFO | jq -r .author)"
-          echo "PR MergedBy: $(echo $PR_INFO | jq -r .mergedBy)"
+          echo "author=$(echo ${PR_INFO} | jq -r .author)" | tee --append $GITHUB_OUTPUT
+          echo "merger=$(echo ${PR_INFO} | jq -r .merger)" | tee --append $GITHUB_OUTPUT
 
       - name: Open a PR to update the requirements
         if: ${{ steps.update.outputs.needs-update }}
--- a/.github/workflows/update-docker-requirements.yml
+++ b/.github/workflows/update-docker-requirements.yml
@@ -32,12 +32,6 @@ jobs:
       - name: Install GitHub CLI
         run: sudo apt-get install gh
 
-      - name: Set up GitHub CLI authentication
-        env:
-          GH_TOKEN: ${{ secrets.GH_TOKEN }}
-        run: |
-          gh auth login --with-token <<< $GH_TOKEN
-
       - name: Update docker requirements
         id: update
         run: |
@@ -52,7 +46,7 @@ jobs:
       - name: Get last merged PR info
         id: pr_info
         env:
-          GH_TOKEN: ${{ secrets.GH_TOKEN }}
+          GITHUB_TOKEN: ${{ secrets.GH_TOKEN }}
         run: |
           LAST_MERGED_PR=$(gh pr list --repo $GITHUB_REPOSITORY --state closed --json number --jq '.[0].number')
           PR_INFO=$(gh pr view $LAST_MERGED_PR --repo $GITHUB_REPOSITORY --json author,mergedBy --jq '{author: .author.login, mergedBy: .mergedBy.login}')
--- a/.github/workflows/update-docker-requirements.yml
+++ b/.github/workflows/update-docker-requirements.yml
@@ -52,8 +52,8 @@ jobs:
           PR_NUMBER=$(gh api \
             -H 'Accept: application/vnd.github+json' \
             /repos/${{ github.repository }}/commits/${{ github.sha }}/pulls \
-            --jq '.[0].number' || echo "none")
-          if [ "${PR_NUMBER}" != "none" ]; then
+            --jq '.[0].number')
+          if [ $? -eq 0 ]; then
             PR_INFO=$(gh pr view ${PR_NUMBER} --json author,mergedBy --jq '{author: .author.login, merger: .mergedBy.login}')
 
             if [ $? -eq 0 ]; then
--- a/.github/workflows/update-docker-requirements.yml
+++ b/.github/workflows/update-docker-requirements.yml
@@ -1,12 +1,9 @@
 name: docker-requirements
 
 on:
-  pull_request:
-    paths:
-      - ".github/workflows/update-docker-requirements.yml"
-      - "pyproject.toml"
-      - "requirements-docker.lock"
-  workflow_dispatch:
+  push:
+    branches:
+      - main
 
 permissions:
   pull-requests: write
--- a/.github/workflows/update-docker-requirements.yml
+++ b/.github/workflows/update-docker-requirements.yml
@@ -49,9 +49,13 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GH_TOKEN }}
         run: |
           LAST_MERGED_PR=$(gh pr list --repo $GITHUB_REPOSITORY --state closed --json number --jq '.[0].number')
+          echo "Last merged PR number: $LAST_MERGED_PR"
           PR_INFO=$(gh pr view $LAST_MERGED_PR --repo $GITHUB_REPOSITORY --json author,mergedBy --jq '{author: .author.login, mergedBy: .mergedBy.login}')
+          echo "PR Info: $PR_INFO"
           echo "pr_author=$(echo $PR_INFO | jq -r .author)" >> $GITHUB_ENV
           echo "pr_mergedBy=$(echo $PR_INFO | jq -r .mergedBy)" >> $GITHUB_ENV
+          echo "PR Author: $(echo $PR_INFO | jq -r .author)"
+          echo "PR MergedBy: $(echo $PR_INFO | jq -r .mergedBy)"
 
       - name: Open a PR to update the requirements
         if: ${{ steps.update.outputs.needs-update }}
--- a/.github/workflows/update-docker-requirements.yml
+++ b/.github/workflows/update-docker-requirements.yml
@@ -13,6 +13,7 @@ on:
 permissions:
   pull-requests: write
   contents: write
+  issues: write
 
 jobs:
   update:
@@ -20,7 +21,7 @@ jobs:
     defaults:
       run:
         shell: bash -el {0}
-
+    
     steps:
       - name: Checkout repository
         uses: actions/checkout@v4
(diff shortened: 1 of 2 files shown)


arjxn-py:
We should also set a repository secret with name `GH_TOKEN` that would most likely be a fine grained access token. cc: @pmeier

.github/workflows/update-docker-requirements.yml
```diff
@@ -1,11 +1,9 @@
...
 
 on:
-  pull_request:
-    paths:
-      - ".github/workflows/update-docker-requirements.yml"
-      - "pyproject.toml"
-      - "requirements-docker.lock"
+  push:
+    branches:
+      - main
```
arjxn-py, line 6:
```suggestion
      - main
    paths:
      - ".github/workflows/update-docker-requirements.yml"
      - "pyproject.toml"
      - "requirements-docker.lock"
```

If you'd only like to run this if these files are changes, feel free to commit this suggestion.

.github/workflows/update-docker-requirements.yml
```diff
@@ -42,10 +43,22 @@ jobs:
           NEEDS_UPDATE=$(git diff --quiet && echo 'false' || echo 'true')
           echo "needs-update=${NEEDS_UPDATE}" | tee --append $GITHUB_OUTPUT
 
+      - name: Get last merged PR info
+        id: pr_info
+        env:
+          GITHUB_TOKEN: ${{ secrets.GH_TOKEN }}
+        run: |
+          LAST_MERGED_PR=$(gh pr list --repo $GITHUB_REPOSITORY --state closed --json number --jq '.[0].number')
+          echo "Last merged PR number: $LAST_MERGED_PR"
```
pmeier, line 52: This logic is flawed unfortunately. Just because we have the last merged PR, it doesn't have to be merged into `main`. Imagine merging a PR into a feature branch.

Instead, we need to query the API for the [PR associated with the commit](https://docs.github.com/en/rest/commits/commits?apiVersion=2022-11-28#list-pull-requests-associated-with-a-commit) we are currently running the workflow on. It should be something like

```bash
RESPONSE=$(gh api \
  -H 'Accept: application/vnd.github+json' \
  -H 'X-GitHub-Api-Version: 2022-11-28' \
  /repos/{owner}/{repo}/commits/${{ github.sha }}/pulls)
if [ $? -ne 0 ]; then
  echo 'has-pr=false' | tee --append $GITHUB_OUTPUT
  exit 0
fi

echo 'has-pr=true' | tee --append $GITHUB_OUTPUT
PR_NUMBER=$(jq '.[0].number' <<< "${RESPONSE}")
```

We have to adjust the body of the PR that we open based on `${{ ${{ steps.pr-info.outputs.has-pr }} }}`.

.github/workflows/update-docker-requirements.yml
```diff
@@ -42,10 +43,22 @@ jobs:
           NEEDS_UPDATE=$(git diff --quiet && echo 'false' || echo 'true')
           echo "needs-update=${NEEDS_UPDATE}" | tee --append $GITHUB_OUTPUT
 
+      - name: Get last merged PR info
+        id: pr_info
+        env:
+          GITHUB_TOKEN: ${{ secrets.GH_TOKEN }}
```
pmeier, line 49: Do we really need that? Everything we do here should be public information.

.github/workflows/update-docker-requirements.yml
```diff
@@ -42,10 +43,22 @@ jobs:
           NEEDS_UPDATE=$(git diff --quiet && echo 'false' || echo 'true')
           echo "needs-update=${NEEDS_UPDATE}" | tee --append $GITHUB_OUTPUT
 
+      - name: Get last merged PR info
+        id: pr_info
```
pmeier, line 47: Let's name this
 
```suggestion
        id: pr-info
```

as we do everywhere else in this repo as well.

.github/workflows/update-docker-requirements.yml
```diff
@@ -42,10 +43,22 @@ jobs:
...
+        env:
+          GITHUB_TOKEN: ${{ secrets.GH_TOKEN }}
+        run: |
+          LAST_MERGED_PR=$(gh pr list --repo $GITHUB_REPOSITORY --state closed --json number --jq '.[0].number')
+          echo "Last merged PR number: $LAST_MERGED_PR"
+          PR_INFO=$(gh pr view $LAST_MERGED_PR --repo $GITHUB_REPOSITORY --json author,mergedBy --jq '{author: .author.login, mergedBy: .mergedBy.login}')
+          echo "PR Info: $PR_INFO"
```
pmeier, line 54: That is just duplicated information with the stuff below.

.github/workflows/update-docker-requirements.yml
```diff
@@ -42,10 +43,22 @@ jobs:
...
+          echo "pr_author=$(echo $PR_INFO | jq -r .author)" >> $GITHUB_ENV
+          echo "pr_mergedBy=$(echo $PR_INFO | jq -r .mergedBy)" >> $GITHUB_ENV
+          echo "PR Author: $(echo $PR_INFO | jq -r .author)"
+          echo "PR MergedBy: $(echo $PR_INFO | jq -r .mergedBy)"
```
pmeier, line 58: 1. Let's not assign this to environment variable if its just arbitrary information rather than something another tool consumes automatically.
2. Instead `echo`ing the value separately, let's use `tee`

```suggestion
          echo "author=$(echo ${PR_INFO} | jq -r .author)" | tee --append $GITHUB_OUTPUT
          echo "merger=$(echo ${PR_INFO} | jq -r .merger)" | tee --append $GITHUB_OUTPUT
```

With this, you can now access the information with `${{ steps.pr-info.outputs.author }}`.

.github/workflows/update-docker-requirements.yml
```diff
@@ -42,10 +43,22 @@ jobs:
...
           echo "needs-update=${NEEDS_UPDATE}" | tee --append $GITHUB_OUTPUT
 
+      - name: Get last merged PR info
+        id: pr_info
+        env:
+          GITHUB_TOKEN: ${{ secrets.GH_TOKEN }}
+        run: |
+          LAST_MERGED_PR=$(gh pr list --repo $GITHUB_REPOSITORY --state closed --json number --jq '.[0].number')
+          echo "Last merged PR number: $LAST_MERGED_PR"
+          PR_INFO=$(gh pr view $LAST_MERGED_PR --repo $GITHUB_REPOSITORY --json author,mergedBy --jq '{author: .author.login, mergedBy: .mergedBy.login}')
```
pmeier, line 53: Nit: simplify

```suggestion
          PR_INFO=$(gh pr view $LAST_MERGED_PR --repo $GITHUB_REPOSITORY --json author,mergedBy --jq '{author: .author.login, merger: .mergedBy.login}')
```

.github/workflows/update-docker-requirements.yml
```diff
@@ -56,10 +69,8 @@ jobs:
           # prettier-ignore
           body:
             https://github.com/${{ github.repository }}/actions/runs/${{ github.run_id }}
-          reviewers: ${{ github.actor }}
+          reviewers: ${{ env.pr_author }},${{ env.pr_mergedBy }}
```
pmeier, line 72: We can leave the merger here, but I don't know whether we can put an arbitrary user as reviewer. For example, I cannot ask for your review on a PR. We should just ping the author in the body.

.github/workflows/update-docker-requirements.yml
```diff
@@ -56,10 +69,8 @@ jobs:
...
 
       - name: Show diff
-        # prettier-ignore
-        if:
-          ${{ steps.update.outputs.needs-update && github.event_name != 'workflow_dispatch' }}
+        if: ${{ steps.update.outputs.needs-update }}
```
pmeier, line 75: This whole step can be removed as we either get PR or we don't. No one will look at the action run anyway.

pmeier (commented):
Thanks Arjun. I left a bunch of comments. We also need to think of a mechanism to surface a failure of this workflow somehow, because if we never get PRs that could also mean that the workflow is broken. So maybe use an action that opens an issue on failures?

.github/workflows/update-docker-requirements.yml
arjxn-py: > Thoughts?

Your suggestion makes sense, Thanks. 
I'm on it to refine it the same way.

.github/workflows/update-docker-requirements.yml
```diff
@@ -42,24 +47,40 @@ jobs:
...
+        id: pr-check
+        env:
+          GITHUB_TOKEN: ${{ secrets.GH_TOKEN }}
+        run: |
+          PR_NUMBER=$(gh api \
+            -H 'Accept: application/vnd.github+json' \
+            /repos/${{ github.repository }}/commits/${{ github.sha }}/pulls \
+            --jq '.[0].number' || echo "none")
+          if [ "${PR_NUMBER}" != "none" ]; then
+            HAS_PR='true'
```
pmeier, line 60: I guess we don't even need this information and can just check the merger or author for "none". Or am I missing something? If need, let's remove it.

.github/workflows/update-docker-requirements.yml
```diff
@@ -42,24 +47,40 @@ jobs:
...
 
+      - name: Check if commit is associated with a PR
+        id: pr-check
+        env:
+          GITHUB_TOKEN: ${{ secrets.GH_TOKEN }}
+        run: |
+          PR_NUMBER=$(gh api \
+            -H 'Accept: application/vnd.github+json' \
+            /repos/${{ github.repository }}/commits/${{ github.sha }}/pulls \
+            --jq '.[0].number' || echo "none")
```
pmeier, line 58: Why the change to defaulting to "none" here instead of relying on the exit code?

.github/workflows/update-docker-requirements.yml
```diff
@@ -42,24 +47,40 @@ jobs:
...
-          reviewers: ${{ github.actor }}
-
-      - name: Show diff
-        # prettier-ignore
-        if:
-          ${{ steps.update.outputs.needs-update && github.event_name != 'workflow_dispatch' }}
-        run: git diff --exit-code
+          body: |
+            Automatic update of requirements-docker.lock.
+            [Workflow run details](https://github.com/${{ github.repository }}/actions/runs/${{ github.run_id }})
```
pmeier, line 85: Currently the author information is not used. We should just add the equivalent of `cc @${{ author }}` to the body if we have an author.

.github/workflows/update-docker-requirements.yml
arjxn-py: I guess yes, as when i tried removing that, it throws this error - [https://github.com/arjxn-py/ragna/actions/runs/10141546877/job/28039000468#step:6:26](https://github.com/arjxn-py/ragna/actions/runs/10141546877/job/28039000468#step:6:26)

.github/workflows/update-docker-requirements.yml
arjxn-py: Removed

.github/workflows/update-docker-requirements.yml
arjxn-py: That's done, cc: @author if author!=None

.github/workflows/update-docker-requirements.yml
arjxn-py: Removed it, yet to test it. 
When testing this branch on my fork it fails because it cannot find #459 PR on the fork.

arjxn-py: here is a test run - [https://github.com/arjxn-py/ragna/actions/runs/10183118614/job/28167297481](https://github.com/arjxn-py/ragna/actions/runs/10183118614/job/28167297481)

arjxn-py:
Here'a a better test run - [https://github.com/arjxn-py/ragna/actions/runs/10183204935/job/28167581020](https://github.com/arjxn-py/ragna/actions/runs/10183204935/job/28167581020)

& the sample PR that's been opened as a result - [https://github.com/arjxn-py/ragna/pull/10](https://github.com/arjxn-py/ragna/pull/10)

arjxn-py:
> We also need to think of a mechanism to surface a failure of this workflow somehow, because if we never get PRs that could also mean that the workflow is broken. So maybe use an action that opens an issue on failures?

Question: if there are No PR's, there's no update on dependencies. Does `requirements-docker.lock` still needs updates in that case?

pmeier (commented):
Looking pretty good Arjun! A couple more comments inline.

pmeier:
> if there are No PR's, there's no update on dependencies. Does `requirements-docker.lock` still needs updates in that case?

No. If we (sadly) ever get to the point where there are no commits on this repo anymore, there is also no need to update docker requirements. If that somehow is needed for some reason, you can also do `git checkout main && git commit -m 'trigger CI' --allow-empty && git push`.

.github/workflows/update-docker-requirements.yml
```diff
@@ -42,24 +45,73 @@ jobs:
...
+          GITHUB_TOKEN: ${{ secrets.GH_TOKEN }}
+        run: |
+          PR_NUMBER=$(gh api \
+            -H 'Accept: application/vnd.github+json' \
+            /repos/${{ github.repository }}/commits/${{ github.sha }}/pulls \
+            --jq '.[0].number')
+          if [ $? -eq 0 ]; then
+            PR_INFO=$(gh pr view ${PR_NUMBER} --json author,mergedBy --jq '{author: .author.login, merger: .mergedBy.login}')
+
+            if [ $? -eq 0 ]; then
```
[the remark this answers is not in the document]
arjxn-py: Thanks for pointing this out. I guess i overdid it 😅 
Just simplified the logic

.github/workflows/update-docker-requirements.yml
```diff
@@ -42,24 +45,73 @@ jobs:
...
 
-      - name: Show diff
-        # prettier-ignore
-        if:
-          ${{ steps.update.outputs.needs-update && github.event_name != 'workflow_dispatch' }}
-        run: git diff --exit-code
+  notify-failure:
+    runs-on: ubuntu-latest
+    needs: update
+    if: failure()
```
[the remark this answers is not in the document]
arjxn-py: Done

.github/workflows/update-docker-requirements.yml
```diff
@@ -42,24 +45,73 @@ jobs:
...
+    steps:
+      - name: Create failure issue
+        uses: actions/github-script@v6
```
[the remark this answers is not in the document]
arjxn-py: Thanks for suggesting, I switched to this only.
I didn't actually looked into for a previous existing action & just started implementing. Something for me to keep in mind from next time.

arjxn-py:
> No. If we (sadly) ever get to the point where there are no commits on this repo anymore, there is also no need to update docker requirements. If that somehow is needed for some reason, you can also do `git checkout main && git commit -m 'trigger CI' --allow-empty && git push`.

I don't think we need `git checkout main && git commit -m 'trigger CI' --allow-empty && git push` or it'd help as we have certain files that trigger this workflow. But it should be a nice thing to configure something like `dependabot`, I believe this workflow would compliment it nicely. And also keep the project updated.

pmeier (approved):
LGTM, thanks Arjun!

PR #460, opened 2024-07-25
Add metadata filter support for chromadb

nenb:
WIP


ragna/core/_rag.py
```diff
@@ -236,9 +236,9 @@ async def answer(self, prompt: str, *, stream: bool = False) -> Message:
         return answer
 
     def _parse_input(
-        self, input: Iterable[Any]
-    ) -> tuple[Optional[list[Document]], MetadataFilter, bool]:
-        if isinstance(input, MetadataFilter):
+        self, input: Optional[Iterable[Any]]
```
pmeier, line 239: This improves it, but still is not quite right. Let's make it

```suggestion
        self, input: Union[MetadataFilter, None, Iterable[Union[Document, str, Path]]]
```

I'm aware that this is a mouthful, but we are heavily overloading `input` here. We probably need to change this annotation in multiple places.

ragna/source_storages/_chroma.py
```diff
@@ -36,19 +36,26 @@ def __init__(self) -> None:
             )
         )
 
-    def _get_collection(self) -> chromadb.Collection:
+    def _get_collection(self, corpus_name: Optional[str]) -> chromadb.Collection:
+        from chromadb.api.types import EmbeddingFunction as ChromaEmbeddingFunction
+
+        if corpus_name is None:
+            corpus_name = self._embedding_id
```
pmeier, line 43: Am I missing something or is `self._embedding_id` not set anywhere?

ragna/source_storages/_chroma.py
```diff
@@ -36,19 +36,26 @@ def __init__(self) -> None:
...
+
         return self._client.get_or_create_collection(
-            self._embedding_id, embedding_function=self._embedding_function
+            corpus_name,
+            embedding_function=cast(ChromaEmbeddingFunction, self._embedding_function),
```
pmeier, line 47: Is the `cast` required here? We are already doing that when setting the attribute

https://github.com/Quansight/ragna/blob/7c5728af3673868222ded6f06bb02f6f6d9cf03a/ragna/source_storages/_vector_database.py#L71-L74

ragna/source_storages/_chroma.py
```diff
@@ -36,19 +36,26 @@ def __init__(self) -> None:
...
         )
 
     def store(
         self,
         documents: list[Document],
         *,
+        corpus_name: Optional[str] = None,
```
pmeier, line 54: Ok for now, but we should discuss the name at some point. We can either standardize on one term, e.g. `corpus_name`, for all source storages, or use the specific one, e.g. `collection_name` for Chroma and something else for LanceDB.

ragna/source_storages/_demo.py
```diff
@@ -63,9 +63,11 @@ def store(self, documents: list[Document]) -> None:
     }
 
     def _apply_filter(
-        self, metadata_filter: MetadataFilter
+        self, metadata_filter: Optional[MetadataFilter]
     ) -> list[tuple[int, dict[str, Any]]]:
-        if metadata_filter.operator is MetadataOperator.RAW:
+        if metadata_filter is None:
+            return [(idx, row) for idx, row in enumerate(self._storage)]
```
pmeier, line 69: Nit

```suggestion
            return list(enumerate(self._storage))
```

pmeier (commented):
This is looking pretty good Nick. I left a few comments.

ragna/source_storages/_chroma.py
pmeier: It was only set on `main`. I updated the base branch.

ragna/source_storages/_chroma.py
nenb: I don't have any strong opinions here. Happy for you to push whatever change you think is best.

pmeier (approved):
Stamping for velocity. We can deal with the test failures in a follow-up, e.g. #461.

PR #461, opened 2024-07-25
Use MetadataFilter in LanceDB

blakerosenthal:
Add metadata filter to lancedb storage


ragna/source_storages/_lancedb.py
```diff
@@ -81,6 +94,7 @@ def store(
                             "id": str(uuid.uuid4()),
                             "document_id": str(document.id),
                             "document_name": str(document.name),
+                            **document.metadata,
```
blakerosenthal, line 97: Currently lancedb doesn't allow partial table updates, so if we create a schema with all the possible metadata fields, each insertion needs to contain every field.

@pmeier Do you have a preference between these options?
1. Create the table schema at runtime with `id`, `document_id`, `document_name`, `page_numbers`, `text`, and `metadata`, where `metadata` is some kind of struct or serialized dict. Then `_translate_metadata_filter` will need to account for some of the fields being nested in the `metadata` column. Pros: can create the schema at runtime, can call `.store()` multiple times without rebuilding the schema. Cons: some potentially messy nesting.
2. Create top-level fields from the union of keys from the document metadatas and build the schema within `.store()` (this is what I have here). Cons: need to build out empty values for documents that come in with missing metadata fields because LanceDB won't accept missing fields.
3. Just create a simple schema as in Option 1 but exclude the `metadata` field. Create the table from this basic schema, then when a document comes in with additional metadata, add new columns for the metadata. Pros: matches most closely the behavior of Chroma. Cons: possibly slow unless [`LanceTable.add_columns`](https://github.com/lancedb/lancedb/blob/513926960db94f534f3fee69fafbc02d0e576488/python/python/lancedb/table.py#L1740) is okay to run a lot.

cc @nenb

pmeier: IMO

1. Unless there is some magic that Lance readily offers, this will be a nightmare to implement. Imagine metadata being a serialized JSON object. How are we going to effectively filter on the individual fields?
2. / 3. I'm unsure what the difference is. We should probably treat the first document that we see as "prototype" and build the table with its metadata. In case all documents have the same metadata, which is a reasonable assumption for >= 80% of uses cases, this requires no extra work. In case the assumption doesn't hold, we'll have to
    1. fill missing metadata fields with `None` or whatever is appropriate
    2. `add_columns` in case we encounter new metadata.

Does that make sense?

blakerosenthal: Definitely makes sense. I'll go ahead and make this change. My remaining question is do we ever expect users to call `.store()` multiple times and, if so, should the method
1. append
2. overwrite
3. offer a `overwrite: bool = False` param?

tests/source_storages/test_source_storages.py
```diff
@@ -1,41 +1,99 @@
...
+                ]
+            ),
+            2,
+            id="or-nested",
+        ),
+        pytest.param(MetadataFilter.eq("key", "value"), 2, id="eq"),
+        pytest.param(MetadataFilter.ne("key", "value"), 5, id="ne"),
+        pytest.param(MetadataFilter.in_("key", ["foo", "bar"]), 2, id="in"),
+        pytest.param(MetadataFilter.not_in("key", ["foo", "bar"]), 5, id="not_in"),
+        pytest.param(None, 7, id="none"),
```
blakerosenthal, line 73: @pmeier @nenb I need to investigate why some of these tests are failing. Chroma and LanceDB return different sources for the same MetadataFilters over the same documents.

ragna/source_storages/_lancedb.py
```diff
@@ -43,31 +43,54 @@ def __init__(self) -> None:
...
+
+        self._schema = pa.schema(fields)
 
     _VECTOR_COLUMN_NAME = "embedded_text"
 
+    def _get_table(self, corpus_name: Optional[str] = None):
+
+        table_name = corpus_name or self._embedding_id
+        return self._db.create_table(
+            name=table_name,
```
pmeier, line 68: Nit: make this concise

```suggestion
        return self._db.create_table(
            name=corpus_name or self._embedding_id,
```

ragna/source_storages/_lancedb.py
```diff
@@ -43,31 +43,54 @@ def __init__(self) -> None:
...
+            pa.field("page_numbers", pa.string()),
+            pa.field("text", pa.string()),
+            pa.field(
+                self._VECTOR_COLUMN_NAME,
+                pa.list_(pa.float32(), self._embedding_dimensions),
+            ),
+            pa.field("num_tokens", pa.int32()),
+        ]
+
+        self._schema = pa.schema(fields)
```
pmeier, line 60: Nit: make it concise

```suggestion
        self._schema = pa.schema(
            [
                pa.field("id", pa.string()),
                pa.field("document_id", pa.string()),
                pa.field("document_name", pa.string()),
                pa.field("page_numbers", pa.string()),
                pa.field("text", pa.string()),
                pa.field(
                    self._VECTOR_COLUMN_NAME,
                    pa.list_(pa.float32(), self._embedding_dimensions),
                ),
                pa.field("num_tokens", pa.int32()),
            ]
        )
```

ragna/source_storages/_lancedb.py
pmeier: As discussed, we assume 1. for the builtin source storages.

ragna/source_storages/_lancedb.py
```diff
@@ -43,31 +43,54 @@ def __init__(self) -> None:
...
+            set(table.schema.names)
+            | set([k for doc in documents for k in doc.metadata.keys()])
+        )
+        missing_fields = [
+            field for field in all_fields if field not in table.schema.names
+        ]
+
+        if missing_fields:
+            table.add_columns({field: "''" for field in missing_fields})
+            self._schema = table.schema
```
pmeier, line 93: Any reason they have to be lists? Keeping them as sets makes this easier and on top avoids the extra conversion.

```suggestion
        document_fields = {field for document in documents for field in document.metadata.keys()}
        schema_fields = set(table.schema.names)
        
        missing_fields = document_fields - schema_fields
        if missing_fields:
            table.add_columns({field: "''" for field in missing_fields})
            self._schema = table.schema

        extra_fields = schema_fields - document_fields
        extra_metadata = {field: "''" for field in extra_fields}
```

ragna/source_storages/_lancedb.py
```diff
@@ -78,9 +101,13 @@ def store(
                 table.add(
                     [
                         {
+                            field: "''" for field in all_fields
+                        }  # ensure all columns exist
+                        | {
```
pmeier, line 106: Cont'd from suggestion above

```suggestion
```

ragna/source_storages/_lancedb.py
```diff
@@ -78,9 +101,13 @@ def store(
...
                             "id": str(uuid.uuid4()),
                             "document_id": str(document.id),
                             "document_name": str(document.name),
+                            **document.metadata,
```
pmeier, line 110: Cont'd from suggestion above

```suggestion
                            # Unpacking the default metadata first
                            # so it can be overridden by the concrete
                            # values if available
                            **extra_metadata
                            **document.metadata,
```

test_metadata.ipynb
pmeier, line 1: Remove before merge.

pmeier (commented):
@blakerosenthal Tests and lint / format CI is failing. I'll review the test when everything is green otherwise.

blakerosenthal:
@pmeier Thanks for all the comments. Ran into a couple quirks with LanceDB's `add_columns` method, but otherwise all your suggestions worked out. Do you know what's happening with mypy in [this commit](https://github.com/Quansight/ragna/pull/461/commits/5821a4aebf66b4cd44d36d9a1904f759cb2d7558)? It's [complaining](https://github.com/Quansight/ragna/actions/runs/10290674791/job/28481174606) about missing types in LanceDB but I thought that the overrides in pyproject.toml would take care of that.

pmeier:
> Do you know what's happening with mypy

Yes. If you have code like this

```python
from typing import TYPE_CHECKING

if TYPE_CHECKING:
    from ragna import Rag

def foo() -> Rag:
    from ragna import Rag

    return Rag()
```

it looks like this should work. Running `mypy` on this succeeds. But running this script fails with

```
  File "/home/philip/git/ragna/main.py", line 7, in <module>
    def foo() -> Rag:
NameError: name 'Rag' is not defined
```

The reason is that Python evaluates the annotations at definition. For example

```python
from ragna import Rag

def foo() -> Rag:
    return Rag()

print(foo.__annotations__)
```

```
{'return': <class 'ragna.core.Rag'>}
```

And since at definition in the top script, the `Rag` object is not available, you face the error.

That being said, there is an easy solution for it: turn the annotation into a string.

```python
from typing import TYPE_CHECKING

if TYPE_CHECKING:
    from ragna import Rag

def foo() -> "Rag":
    from ragna import Rag

    return Rag()

print(foo.__annotations__)
```

```
{'return': 'Rag'}
```

Now `mypy` and Python are happy. Internally `mypy` resolves the string back to its actual type with [`typing.get_type_hints()`](https://docs.python.org/3/library/typing.html#typing.get_type_hints) and so does Ragna as well:

https://github.com/Quansight/ragna/blob/2066dcdeb949060873a8552449b848b245e0e985/ragna/core/_components.py#L73-L75

Finally, because it is a little tedious to always stringify the annotations, there is a shortcut. If you `from __future__ import annotations` at the top of your module, the stringification is applied automatically in the background:

```python
from __future__ import annotations

from typing import TYPE_CHECKING

if TYPE_CHECKING:
    from ragna import Rag

def foo() -> Rag:
    from ragna import Rag

    return Rag()

print(foo.__annotations__)
```

And this is what I did in 4a3b32e22fde58274206529585b41f4c0a8e9b22.

PR #425, opened 2024-05-27
refactor assistant streaming and create OpenAI compliant base class

pmeier:
This came from an offline discussion with @nenb and supersedes #424. It also paves the way for #375.

The main two changes are

1. Factor out the streaming protocol logic, i.e. SSE and JSONL streaming, to avoid code duplication and easy switching if this is the only change between assistants. The latter part lead directly to 2.
2. Implement a generic OpenAI compliant base class allowing the selection of the streaming protocol, an arbitrary URL as well as optional model selection.

More details inline.

PR #462, opened 2024-07-26
fix streaming handling for builtin assistants

pmeier:
My patch in #425 is broken when using the Python API with `asyncio.run`, i.e. the regular use case. As implemented in #425, `break`ing out of the loop in `.answer()` means that the async iterator is not fully exhausted, i.e. `AsyncStopIteration` is never raised. This leads to the context manager that handles the HTTP request staying open longer than `Chat.answer`. When it is cleaned up eventually, internally it calls `await request.aclose()`. However, if at that point the async event loop is already killed, e.g. because the coroutine called by `asyncio.run` is finished, you'll get an error:

```
Task exception was never retrieved
future: <Task finished name='Task-5' coro=<<async_generator_athrow without __name__>()> exception=RuntimeError('aclose(): asynchronous generator is already running')>
RuntimeError: aclose(): asynchronous generator is already running
Exception ignored in: <coroutine object Response.aclose at 0x73f95efdfcc0>
  File "/home/philip/.conda/envs/ragna-deploy-dev/lib/python3.9/site-packages/httpx/_models.py", line 1008, in aclose
    await self.stream.aclose()
  File "/home/philip/.conda/envs/ragna-deploy-dev/lib/python3.9/site-packages/httpx/_client.py", line 155, in aclose
    await self._stream.aclose()
  File "/home/philip/.conda/envs/ragna-deploy-dev/lib/python3.9/site-packages/httpx/_transports/default.py", line 259, in aclose
    await self._httpcore_stream.aclose()
  File "/home/philip/.conda/envs/ragna-deploy-dev/lib/python3.9/site-packages/httpcore/_async/connection_pool.py", line 374, in aclose
    await self._stream.aclose()
  File "/home/philip/.conda/envs/ragna-deploy-dev/lib/python3.9/site-packages/httpcore/_synchronization.py", line 225, in __exit__
    self._anyio_shield.__exit__(exc_type, exc_value, traceback)
  File "/home/philip/.conda/envs/ragna-deploy-dev/lib/python3.9/site-packages/anyio/_backends/_asyncio.py", line 406, in __exit__
    if current_task() is not self._host_task:
  File "/home/philip/.conda/envs/ragna-deploy-dev/lib/python3.9/asyncio/tasks.py", line 38, in current_task
    loop = events.get_running_loop()
RuntimeError: no running event loop
```

To fix this, this PR introduces the same pattern for our streaming handling that `httpx` and `httpx_sse` use as well: a context manager creates the stream and the stream can than be iterated on. 

https://github.com/Quansight/ragna/blob/7c5728af3673868222ded6f06bb02f6f6d9cf03a/ragna/assistants/_http_api.py#L83-L89

https://github.com/Quansight/ragna/blob/7c5728af3673868222ded6f06bb02f6f6d9cf03a/ragna/assistants/_http_api.py#L99-L103

Basically, we switch from

```python
async for data in self._call_api(...):
    ...
```

to

```python
async with self._call_api(...) as stream:
    async for data in stream:
        ...
```

With this pattern, the cleanup is correctly handled when we exit `.answer()` and thus get rid of the error.

### DIFF
--- a/tests/assistants/test_api.py
+++ b/tests/assistants/test_api.py
@@ -1,12 +1,19 @@
+import asyncio
+import itertools
+import json
 import os
+import time
+from pathlib import Path
 
+import httpx
 import pytest
 
 from ragna import assistants
 from ragna._compat import anext
-from ragna.assistants._http_api import HttpApiAssistant
+from ragna._utils import timeout_after
+from ragna.assistants._http_api import HttpApiAssistant, HttpStreamingProtocol
 from ragna.core import Message, RagnaException
-from tests.utils import skip_on_windows
+from tests.utils import background_subprocess, get_available_port, skip_on_windows
 
 HTTP_API_ASSISTANTS = [
     assistant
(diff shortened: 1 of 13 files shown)


pmeier:
> Could you add a unit test to prevent this regression in the future?

I thought about it, but I honestly have trouble coming up with one that is not unreasonably complex. We need a server that we can hit for HTTP streaming. And the best option we have here is to start our own.

I'll see if we can have this in a reasonable way.

tests/deploy/ui/test_ui.py
```diff
@@ -11,12 +10,7 @@
 from ragna._utils import timeout_after
 from ragna.deploy import Config
 from tests.deploy.utils import TestAssistant
-
-
-def get_available_port():
```
pmeier, line 16: Moved it to the generic test utils as this is no longer just needed for deploy.

tests/utils.py
```diff
@@ -1,7 +1,27 @@
...
 
 import pytest
 
 skip_on_windows = pytest.mark.skipif(
     platform.system() == "Windows", reason="Test is broken skipped on Windows"
 )
+
+
+@contextlib.contextmanager
+def background_subprocess(*args, stdout=sys.stdout, stderr=sys.stdout, **kwargs):
```
pmeier, line 15: We had this before, but it was removed in #322. Probably can also be used by the UI tests, but we can do that in a follow-up.

tests/assistants/test_api.py
```diff
@@ -30,3 +37,105 @@ async def test_api_call_error_smoke(mocker, assistant):
...
+        expected_chunk = next(expected_chunks)
+        assert actual_chunk == expected_chunk
+
+    with pytest.raises(StopIteration):
+        next(expected_chunks)
+
+
+@pytest.mark.parametrize("streaming_protocol", list(HttpStreamingProtocol))
+def test_http_streaming_termination(streaming_server, streaming_protocol):
+    # Non-regression test for https://github.com/Quansight/ragna/pull/462
```
pmeier, line 122: @nenb This test fails when I revert my patch in this PR. See the CI runs for 0b9211ee366f844570c4edb3ce40e28280951ce5.

PR #464, opened 2024-08-02
add session based auth workflow

### LINKED ISSUE #158: Demo authentication should enforce a username
There is no check for that. Worse, if running without a central password, not putting anything for username and password actually authenticates. Reported in https://github.com/Quansight/ragna/discussions/156#discussioncomment-7460848.

### LINKED ISSUE #178: Refactor authentication in the UI

### LINKED ISSUE #179: Add NoAuthentication and make it the default
Per title when #178 is done. We have had several reports that the default demo authentication is no intuitive:

- #167
- https://github.com/Quansight/ragna/discussions/156#discussioncomment-7460848
- https://github.com/Quansight/ragna/issues/166#issuecomment-1801813313

We tried to mitigate it somewhat with 

- #157
- #168

but ultimately adding extra documentation doesn't make it any more intuitive.

Since no one needs authentication for local experimentation, we should have a `NoAuthentication` class that should be the default.

cc @nenb

pmeier:
Closes #158, closes #178, and closes #179. This is a massive PR that completely overhauls our auth flow. The old auth flow had multiple problems detailed in https://github.com/Quansight/ragna/issues/178#issuecomment-2003254762.

With this PR we now have a session based auth workflow that is independent of the UI through the following endpoints:

- `GET /login`: user defined login page
- `POST /login`: login endpoint for form-based auth, e.g. username and password
- `GET /oauth-callback`: login endpoint for OAuth based auth
- `GET /logout`: logout endpoint to clear the active session

The login endpoints can be configured through the `ragna.config.Auth` object.

After a successful login we set a session cookie and create an entry for the corresponding user in a key-value-store. With this PR we have support for in-memory and Redis key-value-stores.

Although we don't have a way to display it yet, the auth flow already supports authenticating with an API token, which is generated for each user.

Todo in follow-up PRs:

- documentation
- tests
- automatic handling for refresh tokens when using OAuth
- generic OAuth object

I'm aware that reviewing this is quite the ask. Happy to schedule some time to walk you through any part of the change.


pmeier:
Self merging for velocity

PR #467, opened 2024-08-04
try docker requirements update

pmeier:
cc @arjxn-py Just a test for your patch in #459. If all goes well, we should see a PR being opened against `main` and I should be pinged on it.

### DIFF
--- a/.github/workflows/update-docker-requirements.yml
+++ b/.github/workflows/update-docker-requirements.yml
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - main
+      - test-ci
     paths:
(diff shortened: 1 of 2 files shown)


pmeier:
Ok, that did not work: https://github.com/Quansight/ragna/actions/runs/10239123151

Could you send PR against the `test-ci` branch until we have it right? We'll clean up afterwards and merge the fix into main.

arjxn-py:
> Could you send PR against the `test-ci` branch until we have it right? We'll clean up afterwards and merge the fix into main.

On it @pmeier, found a minor issue with the docker issue template.
Apologies, I should have been more precautious.

PR #474, opened 2024-08-04
fix docker requirements update workflow

pmeier:
Follow-up to https://github.com/Quansight/ragna/pull/467#issuecomment-2267662822.

cc @arjxn-py


arjxn-py (approved):
Looks Good. Thanks @pmeier 💯

PR #476, opened 2024-08-05
support pydantic (de)serialization on MetadataFilter

pmeier:
The is a prerequisite for using `MetadataFilter` as part of FastAPI. I've scrapped the actual JSON serialization, because that would create JSON inside JSON with pydantic and pydantic knows how to serialize dicts anyway.

### DIFF
--- a/ragna/core/_metadata_filter.py
+++ b/ragna/core/_metadata_filter.py
@@ -1,9 +1,11 @@
 from __future__ import annotations
 
 import enum
-import json
 import textwrap
-from typing import Any, Literal, Sequence, cast
+from typing import Any, Literal, Sequence, Union, cast
+
+import pydantic
+import pydantic_core
 
 
 class MetadataOperator(enum.Enum):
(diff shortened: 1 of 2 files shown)


nenb (approved):
I was working through this, and I was a little bit dissatisfied that `dict_schema` was only able to do a very basic check on input schemas eg `Model(mf={"IN": {}})` passes the input schema check, but then throws an obscure validation error.

It then occurred to me, do we really need the `children` key?

For example, instead of
```
{'OR': {'children': [{'RAW': {'filter': 'raw'}}, {'GE': {'key': 1}}]}}
```

could we have
```
{'OR': [{'RAW': {'filter': 'raw'}}, {'GE': {'key': 1}}]}
```

If so, perhaps this could simplify some of the `pydantic` logic here?

nenb:
Also, I haven't looked into it, but a lot of tests seem to fail. Not sure what is going on...

pmeier:
> It then occurred to me, do we really need the `children` key?

Nope

https://github.com/Quansight/ragna/blob/ad9c1307b31a8e67f09411acecc84e0b4a58ba57/ragna/core/_metadata_filter.py#L23-L26

Per comment, the same for the raw filter. So you can break down your example even more to

```
{'OR': [{'RAW': 'raw'}, {'GE': {'key': 1}}]}
```

That is as concise as it gets. However we also need to consider the use case of translating it. As discussed offline, I dislike the bits you highlighted as well. I invite everyone to come up with a better scheme and I'm happy to adopt it.

> Also, I haven't looked into it, but a lot of tests seem to fail. Not sure what is going on...

Just minor stuff that should be fixed by #461.

pmeier:
Thinking about this more, I realized that there is no requirement for the primitive version to have a 1-to-1 relationship to the Python object. Meaning, we can go for the simplifications proposed in https://github.com/Quansight/ragna/pull/476#pullrequestreview-2219217627 and https://github.com/Quansight/ragna/pull/476#issuecomment-2269759736 and still keep one Python object that always has the `operator`, `key`, and `value` fields.

I implemented that in 70ac0b51fa2865868e79fb4e33667fcb67e32b8a. Here is the example from above:

```python
import json

from ragna.core import MetadataFilter

metadata_filter = MetadataFilter.or_(
    [
        MetadataFilter.raw("raw"),
        MetadataFilter.ge("key", 1),
    ]
)

print(metadata_filter)
print(json.dumps(metadata_filter.to_primitive()))
```

```
OR(
  RAW('raw'),
  GE('key', 1),
)
{"OR": [{"RAW": "raw"}, {"GE": {"key": 1}}]}
```

PR #479, opened 2024-08-06
support medata filter in the REST API

pmeier:


ragna/core/_rag.py
```diff
@@ -191,7 +191,7 @@ async def prepare(self) -> Message:
                 "Chat is already prepared",
                 chat=self,
                 http_status_code=400,
-                detail=RagnaException.EVENT,
+                http_detail=RagnaException.EVENT,
```
pmeier, line 194: Driveby because I hit this during development and I was wondering why I don't get diagnostics back.

ragna/deploy/_api/orm.py
```diff
@@ -19,14 +19,16 @@ class Json(types.TypeDecorator):
 
     cache_ok = True
 
-    def process_bind_param(self, value: Any, dialect: Dialect) -> str:
+    def process_bind_param(self, value: Any, dialect: Dialect) -> Optional[str]:
+        if value is None:
+            return value
```
pmeier, line 24: We now also have a nullable JSON field with the `MetadataFilter` and thus also need to handle this here.

ragna/deploy/_api/orm.py
```diff
@@ -72,12 +74,27 @@ class Document(Base):
...
+class MetadataFilter(Base):
+    __tablename__ = "metadata_filters"
+
+    id = Column(types.Uuid, primary_key=True)  # type: ignore[attr-defined]
+    user_id = Column(ForeignKey("users.id"))
+
+    data = Column(Json, nullable=False)
+
+    chat_id = Column(ForeignKey("chats.id"), nullable=True)
+    chat = relationship("Chat", back_populates="metadata_filter", uselist=False)
```
pmeier, line 86: 1-to-1 relationship with a chat

nenb (approved):
Approving to unblock @pierrotsmnrd.

Will do a post-merge review now.

nenb:
_Taken from Slack:_

It seems fine. The only caveat that I have is that I may have missed something when it comes to serialising the metadata filter to the database, but I think the PR is small enough that it's fine!

PR #484, opened 2024-08-08
UI for metadata filters

pierrotsmnrd:
This PR adds a UI to open a new chat, either by uploading a document, or by using the existing corpus and adding metadata filters.

https://github.com/user-attachments/assets/0e53b28a-857c-491d-8a8f-1d04bee92c8c


ragna/deploy/_ui/central_view.py
```diff
@@ -182,24 +182,75 @@ def __init__(self, api_wrapper, **params):
         )
         self.on_click_chat_info = None
 
+    def convert_metadata_filters_to_human(self, metadata_filters):
```
pmeier, line 185: This is for the chat info side bar. Since we don't have proper design for this, can we maybe just use the default `__repr__` of the metadata filter?

https://github.com/Quansight/ragna/blob/aea6b6706265e342e41682a88ea80bface4961ec/ragna/core/_metadata_filter.py#L31-L46

It's not too bad.

ragna/deploy/_ui/central_view.py
```diff
@@ -182,24 +182,75 @@ def __init__(self, api_wrapper, **params):
...
+            details = f"<div class='details'>{pills}</div><br />\n\n"
+            grid_height = len(self.current_chat["metadata"]["input"]) // 3
 
-        grid_height = len(self.current_chat["metadata"]["documents"]) // 3
+        elif isinstance(self.current_chat["metadata"]["input"], dict):
+            # Metadata filters
+            title = "Metadata Filters"
+            metadata_filters_readable = self.convert_metadata_filters_to_human(
+                self.current_chat["metadata"]["input"]
+            )
```
pmeier, line 237: Continuing form above, this could just be

```suggestion
            metadata_filters_readable = str(MetadataFilter.from_primitive(self.current_chat["metadata"]["input"]))
```

ragna/deploy/_ui/central_view.py
```diff
@@ -225,11 +276,17 @@ def on_click_chat_info_wrapper(self, event):
...
                                 "grid-template": f"repeat({grid_height}, 1fr) / repeat(3, 1fr)",
                             },
-                        )
+                        ),
+                        ui.css(
+                            ":host(.chat_info_markdown) .details_block",
+                            {
+                                "display": "block",
+                            },
+                        ),
```
pmeier, line 289: Let's shed a few commas so the auto formatter unexplodes this

```suggestion
                        ui.css(
                            ":host(.chat_info_markdown) .details_block"
                            {
                                "display": "block"
                            }
                        ),
```

ragna/deploy/_ui/central_view.py
```diff
@@ -246,7 +303,7 @@ def on_click_source_info_wrapper(self, event, sources):
                 location = f": page(s) {location}"
             source_infos.append(
                 (
-                    f"<b>{rank}. {source['document']['name']}</b> {location}",
+                    f"<b>{rank}. {source['document_name']}</b> {location}",
```
pmeier, line 306: Driveby that is not related to this PR and should have been included in #489.

ragna/deploy/_ui/components/metadata_filters_builder.py
```diff
@@ -0,0 +1,307 @@
...
+
+import panel as pn
+import param
+
+from ragna.core._metadata_filter import MetadataFilter
+
+METADATA_FILTERS = {
+    "document_name": {"type": str},
+    "size": {"type": int},
+    "extension": {"type": str},
```
pmeier, line 11: We can keep this for now as it is supplied by our default `LocalDocument`, but they have to be variable before we can merge into `main`.

ragna/deploy/_ui/components/metadata_filters_builder.py
```diff
@@ -0,0 +1,307 @@
...
+    "section": {"type": str},
+    # "creation_date": {"type": datetime},
+}
+
+
+FILTERS_PER_TYPE = {
+    str: ["==", "!=", "in", "not in"],
+    int: ["==", "!=", ">", "<", ">=", "<=", "in", "not in"],
+    # datetime: ["==", "!=", ">", "<", ">=", "<="],
```
pmeier, line 20:
```suggestion
```

ragna/deploy/_ui/components/metadata_filters_builder.py
```diff
@@ -0,0 +1,307 @@
...
+import param
+
+from ragna.core._metadata_filter import MetadataFilter
+
+METADATA_FILTERS = {
+    "document_name": {"type": str},
+    "size": {"type": int},
+    "extension": {"type": str},
+    "section": {"type": str},
+    # "creation_date": {"type": datetime},
```
pmeier, line 13: Let's drop these, because they were specific for the demo we built.

```suggestion
```

ragna/deploy/_ui/components/metadata_filters_builder.py
```diff
@@ -0,0 +1,307 @@
...
+    # datetime: ["==", "!=", ">", "<", ">=", "<="],
+}
+
+PLACEHOLDERS_PER_METADATA_KEY = {
+    "document_name": "",
+    "size": "",
+    "extension": "",
+    "section": "",
+    # "creation_date": "YYYY-mm-dd HH:MM:SS",
+}
```
pmeier, line 29: Do we need this at all for now? Looks like all placeholders are empty at the current stage.

ragna/deploy/_ui/components/metadata_filters_builder.py
```diff
@@ -0,0 +1,307 @@
...
+            "",  # empty one for the first row
+            *METADATA_FILTERS.keys(),
+        ],
+        default="",
+    )
+    operator = param.Selector(
+        objects=[
+            "",  # empty one for the first row
+            # Filled later based on the key
+            # "raw"
```
pmeier, line 44: What does "raw" mean here?

ragna/deploy/_ui/components/metadata_filters_builder.py
```diff
@@ -0,0 +1,307 @@
...
+        else:
+            change_valid_status(True, self.value_text_input)
+
+        return result
+
+    def __str__(self):
+        return f"{self.key} {self.operator} {self.value}"
+
+    def __repr__(self):
+        return f"[key:{self.key} \t op:{self.operator} \t val:{self.value}]"
```
pmeier, line 175: How are these methods used?

ragna/deploy/_ui/components/metadata_filters_builder.py
```diff
@@ -0,0 +1,307 @@
...
+            return MetadataFilter.ge
+        if operator == "<=":
+            return MetadataFilter.le
+        if operator == "in":
+            return MetadataFilter.in_
+        if operator == "not in":
+            return MetadataFilter.not_in
+
+        # if operator == "raw":
+        #    return MetadataFilter.RAW
```
pmeier, line 196: Let's avoid a switch yard here

```suggestion
        return {
            "==": MetadataOperator.EQ,
            "!=": MetadataOperator.NE,
            ...,
        }[operator]
```

ragna/deploy/_ui/components/metadata_filters_builder.py
```diff
@@ -0,0 +1,307 @@
...
+                return int(value)
+            elif METADATA_FILTERS[self.key]["type"] == datetime:
+                return datetime.strptime(value, "%Y-%m-%d %H:%M:%S")
+            return value
+        except ValueError:
+            return value
+
+    def get_metadata_filter(self):
+        if self.key == "" or self.operator == "" or self.value == "":
+            return None
```
pmeier, line 210: Isn't this

```suggestion
        if self.is_empty():
            return None
```

ragna/deploy/_ui/components/metadata_filters_builder.py
```diff
@@ -0,0 +1,307 @@
...
+        return self.convert_operator(self.operator)(
+            key=self.key, value=self.convert_value(self.value)
+        )
```
pmeier, line 213: Instead of letting `convert_operator` return the convenience method, let's have it return the actual operator and instantiate the `MetadataFilter` manually:

```suggestion
        return MetadataFilter(
            operator=self.convert_operator(self.operator),
            key=self.key,
            value=self.convert_value(self.value),
        )
```

ragna/deploy/_ui/components/metadata_filters_builder.py
```diff
@@ -0,0 +1,307 @@
...
+        super().__init__(**params)
+
+        self.add_filter_button = pn.widgets.ButtonIcon(
+            icon="circle-plus", width=25, height=25
+        )
+        self.add_filter_button.on_click(self.did_click_add_filter_button)
+
+        self.delete_buttons = []
+
+        # dummy_row = FilterRow(key="document_name", operator="==", value="applications.md")
```
pmeier, line 229:
```suggestion
```

ragna/deploy/_ui/components/metadata_filters_builder.py
```diff
@@ -0,0 +1,307 @@
...
+            on_delete_callback=self.delete_metadata_filter,
+        )
+
+    def did_click_add_filter_button(self, event):
+        if len(self.metadata_filters) > 0 and not self.metadata_filters[-1].is_empty():
+            new_metadata_filter_row = self.empty_filter()
+            self.metadata_filters = self.metadata_filters + [new_metadata_filter_row]
+
+    def delete_metadata_filter(self, event):
+        print("should delete filter : ", event)
```
pmeier, line 246:
```suggestion
```

ragna/deploy/_ui/components/metadata_filters_builder.py
```diff
@@ -0,0 +1,307 @@
...
+        result = True
+        for filter in self.metadata_filters:
+            # If the last filter is empty, we do not want to validate it.
+            # This allows to have only one empty filter, to question on the whole corpus.
+            if filter == self.metadata_filters[-1] and filter.is_empty():
+                continue
+
+            if not filter.validate():
+                result = False
+                # Do not break, we want to call validate() on every filters
```
pmeier, line 293: Why?

ragna/deploy/_ui/modal_configuration.py
```diff
@@ -115,12 +124,34 @@ def __init__(self, api_wrapper, **params):
...
+                # The Metadata Filters Builder updates its UI when calling validate()
+                return
+
             self.start_chat_button.disabled = True
-            self.document_uploader.perform_upload(event, self.did_finish_upload)
+            asyncio.ensure_future(
+                self.did_finish_upload(
+                    self.metadata_filters_builder.get_metadata_filters()
+                )
+            )
```
pmeier, line 154: Why not make this method async and just `await` the result?

ragna/deploy/_ui/central_view.py
pierrotsmnrd: By using this : 
```metadata_filters_readable = MetadataFilter.from_primitive(self.current_chat["metadata"]["input"])```

I get the following result : 
Capture d’écran 2024-08-14 à 09 06 48

For comparison, the same filters in the current state : 
Capture d’écran 2024-08-14 à 09 07 22

What do you think is best ?

File: ragna/deploy/_ui/components/metadata_filters_builder.py (hunk 8 above)
Comment (pierrotsmnrd) [reply ragna/deploy/_ui/components/metadata_filters_builder.py:29]: Placeholders make a lot of sense for the datetimes, but we don't have any dt metadata anymore, so we don't really need them for the moment. Should I remove everything related to placeholders ?

File: ragna/deploy/_ui/components/metadata_filters_builder.py (hunk 10 above)
Comment (pierrotsmnrd) [reply ragna/deploy/_ui/components/metadata_filters_builder.py:175]: They were very handy while developing.

File: ragna/deploy/_ui/components/metadata_filters_builder.py (hunk 12 above)
Comment (pierrotsmnrd) [reply ragna/deploy/_ui/components/metadata_filters_builder.py:210]: No. is_empty is : 
``` self.key == "" and self.operator == "" and self.value == "" ``` 
ie if all the fields are empty.

Here, we want to return `None` if  at least one field is empty.

File: ragna/deploy/_ui/central_view.py (hunk 3 above)
Comment (pierrotsmnrd) [reply ragna/deploy/_ui/central_view.py]: Erf, I shouldn't have committed this directly from Github's interface, I didn't see there was a missing comma at the end of line 290. I'm fixing this now.

File: ragna/deploy/_ui/components/metadata_filters_builder.py (hunk 16 above)
Comment (pierrotsmnrd) [reply ragna/deploy/_ui/components/metadata_filters_builder.py:284]: Let's take this example : 
Capture d’écran 2024-08-14 à 09 44 14

We have 2 empty fields. We want to validate all of them, and not stop after the first, to highlight in red all the invalid ones : 
Capture d’écran 2024-08-14 à 09 44 41

File: ragna/deploy/_ui/modal_configuration.py (hunk 17 above)
Comment (pierrotsmnrd) [reply ragna/deploy/_ui/modal_configuration.py:154]: The function is a button callback. I have had issues in the past with making such callbacks async. But that seems to work now, so let's do it.

File: ragna/deploy/_ui/components/metadata_filters_builder.py (hunk 9 above)
Comment (pierrotsmnrd) [reply ragna/deploy/_ui/components/metadata_filters_builder.py:44]: Just a remainder of something I reworked afterwards. I remove it

File: ragna/deploy/_ui/central_view.py (hunk 1 above)
Comment (pmeier) [reply ragna/deploy/_ui/central_view.py:185]: In the first one, new lines are not displayed:

```python
from ragna.core import MetadataFilter

print(
    MetadataFilter.and_(
        [
            MetadataFilter.eq("document_name", "test"),
            MetadataFilter.gt("size", 100),
        ]
    )
)
```

```
AND(
  EQ('document_name', 'test'),
  GT('size', 100),
)
```

This is already an improvement. I think your solution suffers from the same problem: if you have more than two filters, it will be hard to display them in one line.

But that is somewhat besides the point: instead of writing a lot of logic that we'll have to maintain to create another human readable format, we could just go with what is already there. I know it is sometimes hard to let go of code that you have already written, but the benefit currently does not outweigh its maintenance cost.

Nothing is lost though. If we ultimately decide to go with your solution, we have the implementation in the commits.

ragna/deploy/_ui/components/metadata_filters_builder.py
pmeier: I think so, yes. Datetimes would be a nice feature, but until we have decided how we want to deal with them, let's shed dead code.

ragna/deploy/_ui/components/metadata_filters_builder.py
pmeier: Fair, but IMO should be removed before merge. Or at least reduced to either `__str__` or `__repr__`

ragna/deploy/_ui/components/metadata_filters_builder.py
pmeier: In that case the variable naming is confusing. So `filter` is not a row, but rather one element inside the row? Maybe name it `filter_component` than?

ragna/deploy/_ui/central_view.py
nenb: I have accepted Philip's suggestion in commit `47a199`.

ragna/deploy/_ui/central_view.py
nenb: I have accepted Philip's suggestion in commit 47a199.

pmeier:
Superseded by #499.

Issue #466, opened 2024-08-03
PDF viewer for sources


blakerosenthal:
### Feature description

We would like to be able to click on the `Source Info` button on the assistant's answer in the UI and show the list of relevant sources, with a `View` or `Open` button on each source. This button can either use Panel's built-in PDF viewer or open the file natively in the browser. Ideally the new pane would scroll to and highlight the exact source content in the file.

pmeier:
We need a way to get the document content to the UI. Thus, I propose three new API endpoints:

- `GET /documents` / `GET /documents/{id}`: Get all [documents](https://github.com/Quansight/ragna/blob/19d326ba4b6d8142b6e70f3f626e14267fb25ad1/ragna/deploy/_api/schemas.py#L18-L20) or one specific one associated with the current user. This is the same as what we have for [chats](https://github.com/Quansight/ragna/blob/19d326ba4b6d8142b6e70f3f626e14267fb25ad1/ragna/deploy/_api/core.py#L256-L264).
- `GET /documents/{id}/content`: Returns the content of the document.

smokestacklightnin:
The current plan of action is to add the get documents API endpoints along with their unit tests in #547 and add the UI elements in a subsequent PR

smokestacklightnin:
I have [completed](https://github.com/Quansight/ragna/pull/547) the [REST API endpoints](https://github.com/Quansight/ragna/issues/466#issuecomment-2267655430), and now it remains to [add the UI elements](https://github.com/Quansight/ragna/issues/551).

PR #486, opened 2024-08-12
Document viewer

blakerosenthal:
- Add API endpoints for document content
- Add source content viewer to UI

Closes https://github.com/Quansight/ragna/issues/466


ragna/deploy/_api/core.py
```diff
@@ -210,6 +210,30 @@ async def upload_document(
...
+        with get_session() as session:
+            _, metadata = database.get_document(session, user=user, id=id)
+            if "path" not in metadata:
+                raise HTTPException(
+                    status_code=400,
+                    detail="Document path not found",
+                )
+            with aiofiles.open(metadata["path"], "rb") as file:
+                while content := await file.read(1024):
+                    yield content
```
pmeier, line 235: Our documents have a `.read()` method that should be used here:

https://github.com/Quansight/ragna/blob/2066dcdeb949060873a8552449b848b245e0e985/ragna/core/_document.py#L75-L77

It is not async for now, but there is no need to require documents to be on the local file system. We just need the conversion to the `Document` object

https://github.com/Quansight/ragna/blob/2066dcdeb949060873a8552449b848b245e0e985/ragna/deploy/_api/schemas.py#L29-L37

Maybe this is also a good time to resolve this TODO first?

ragna/deploy/_ui/api_wrapper.py
```diff
@@ -62,6 +62,13 @@ async def get_chats(self):
             chat["messages"] = [self.improve_message(msg) for msg in chat["messages"]]
         return json_data
 
+    async def get_document_content(self, document_id):
+        async with self.client.stream(
+            "GET", f"/documents/{document_id}/content"
+        ) as stream:
+            async for chunk in stream.aiter_bytes():
+                print(chunk)
```
pmeier, line 70: I'm guessing this is still TODO?

blakerosenthal:
@pmeier I have two outstanding questions before being ready for a full review:
1. Are we only supporting PDFs right now, or do we want to include all supported types? It doesn't seem too difficult to just send any blob to the browser and let the browser decide whether to display it or offer it as a download.
2. I'm currently just calling the `Document.read()` method and sending the entire blob to the browser wrapped in a `fastapi.Response`. I'd be interested in your thoughts on how this should be chunked and streamed properly.

Here's a video of how it looks so far.

https://github.com/user-attachments/assets/9d299775-60c8-48fb-89d7-f90f7d4b2c4a

pmeier:
> 1. Are we only supporting PDFs right now, or do we want to include all supported types? It doesn't seem too difficult to just send any blob to the browser and let the browser decide whether to display it or offer it as a download.

Does the browser handle this automatically? Meaning, we just pass it a blob and the browser figures out whether it can show it (`.pdf`, `.md`, `.txt`) or offer it as download (`.pptx`, `.docx`)? If yes, that would be awesome and we should really have this feature for everything.

> 2. I'm currently just calling the `Document.read()` method and sending the entire blob to the browser wrapped in a `fastapi.Response`. I'd be interested in your thoughts on how this should be chunked and streamed properly.

FastAPI has a [`FileResponse`](https://fastapi.tiangolo.com/advanced/custom-response/#fileresponse) that handles this properly. Downside is that this requires a file on disk to work. Short-term I'd be ok to just read the data and write it to a temporary file that we remove in a [background task](https://fastapi.tiangolo.com/tutorial/background-tasks/) after the response has gone out. Long-term we need to build our own response type that does the same, but starts from `bytes` or an iterable thereof.

> Here's a video of how it looks so far.

Looks good. #466 also states that

> Ideally the new pane would scroll to and highlight the exact source content in the file.

Is that possible / planned?

blakerosenthal:
> Does the browser handle this automatically? Meaning, we just pass it a blob and the browser figures out whether it can show it (`.pdf`, `.md`, `.txt`) or offer it as download (`.pptx`, `.docx`)? If yes, that would be awesome and we should really have this feature for everything.

I thiiink so as long as the right mimetype is specified in the response header. I will confirm and let you know what I find out.

> FastAPI has a [`FileResponse`](https://fastapi.tiangolo.com/advanced/custom-response/#fileresponse) that handles this properly. Downside is that this requires a file on disk to work. Short-term I'd be ok to just read the data and write it to a temporary file that we remove in a [background task](https://fastapi.tiangolo.com/tutorial/background-tasks/) after the response has gone out. Long-term we need to build our own response type that does the same, but starts from `bytes` or an iterable thereof.

Okay, I'd be interested in seeing the level of effort on building our own. It seems like FileResponse is mostly just a wrapper around StreamingResponse that sets the right headers.

> > Ideally the new pane would scroll to and highlight the exact source content in the file.
> Is that possible / planned?
I need to think a bit more about how this could work. Might be best suited for a followup PR.

blakerosenthal:
@pmeier This is ready for another look! This is was the source viewer looks like now after putting the accordion widget back and adding a new button. And I added some MIME types to our supported documents so the browser knows what to do with them when it receives the blob. On my browser PDFs open in a new tab, text is displayed as HTML, and Word and Powerpoints are downloaded. For sources with page numbers there's also now an anchor that scrolls the view to the right page for in-browser sources.
Screenshot from 2024-08-29 08-37-16
Screenshot from 2024-08-29 08-37-30

ragna/core/_document.py
```diff
@@ -200,6 +200,14 @@ def supported_suffixes(cls) -> list[str]:
         """
         pass
 
+    @abc.abstractmethod
+    def mime_type(cls) -> str:
+        """
+        Returns:
+            MIME type of the document handler.
+        """
+        pass
```
pmeier, line 209: 1. Can we have a reasonable default here instead of making it abstract? E.g. downloading the file instead of trying to display it.
2. Why is this on the handler on not on the document itself?

ragna/deploy/_api/schemas.py
```diff
@@ -26,14 +26,14 @@ def from_core(cls, document: ragna.core.Document) -> Document:
             name=document.name,
         )
 
-    def to_core(self) -> ragna.core.Document:
+    def to_core(self, metadata={}) -> ragna.core.Document:
```
pmeier, line 29: 1. Do we even need a default value here?
2. If yes, don't set mutable objects as default values. 
    ```suggestion
        def to_core(self, metadata: Optional[dict[str, Any]] = None) -> ragna.core.Document:
        if metadata is None:
            metadata = {}
    ```

ragna/deploy/_ui/central_view.py
```diff
@@ -235,19 +234,37 @@ def on_click_chat_info_wrapper(self, event):
             ],
         )
 
-    def on_click_source_info_wrapper(self, event, sources):
+    async def on_click_source_info_wrapper(self, event, sources):
```
pmeier, line 237: Why did this change to `async`?

ragna/deploy/_ui/central_view.py
```diff
@@ -68,16 +68,17 @@ def __init__(
...
         self.assistant_toolbar = pn.Row(
             self.clipboard_button,
             pn.widgets.Button(
                 name="Source Info",
                 icon="info-circle",
                 css_classes=["source-info-button"],
-                on_click=lambda event: self.on_click_source_info_callback(
-                    event, self.sources
-                ),
+                on_click=click_source_info_callback_wrapper,
```
pmeier, line 81: Revert this in case the flagged change to async is not needed.

pmeier (commented):
This looks great, thanks Blake! One question though: in case the browser downloads the file, I get something like `"{uuid.UUID}.docx"`. Why isn't this the proper file name?

ragna/core/_document.py
blakerosenthal: Going to answer these out of order:
2) No reason at all, except that the document handler already parses file type from extension. Either the Document itself can also do that or can internally call DocumentHandler. Thoughts?
1) I don't think there's a default MIME type that will always trigger the browser to perform a download because the browser always needs some way to make sense of the incoming byte stream. If we put the method on Document though the default could be to try to parse the extension and raise a ValueError or RagnaException if it fails. I think I like that option.

ragna/deploy/_ui/central_view.py
blakerosenthal: I spent some time trying to make this work as a synchronous callback but was unsuccessful. Since the call to api was async it was throwing a lot of complaints but its back to being synchronous again now that its a JS callback so I'll see if it complains again!

blakerosenthal:
> This looks great, thanks Blake! One question though: in case the browser downloads the file, I get something like `"{uuid.UUID}.docx"`. Why isn't this the proper file name?

That's a typo, thanks for catching!

ragna/core/_document.py
pmeier: > No reason at all, except that the document handler already parses file type from extension. Either the Document itself can also do that or can internally call DocumentHandler. Thoughts?

Maybe we should make MIME-type just an optional parameter of `Document`. I like that more than putting it on the handler. Actually, the handler could also act on the MIME-type of the document rather than on the suffix. Of course that part does not need to happen in this PR. But IMO that makes more sense.

> I don't think there's a default MIME type that will always trigger the browser to perform a download

[RFC 2046 5.4.3](https://datatracker.ietf.org/doc/html/rfc2046#section-4.5.3)

> MIME implementations must at a minimum treat any unrecognized subtypes as being equivalent to "application/octet-stream".

That sounds like a reasonable default to me.

> If we put the method on Document though the default could be to try to parse the extension and raise a ValueError or RagnaException if it fails.

This ties into the first point of my comment. We could try to parse the MIME type from the suffix and if we can't determine it, we use `application/octet-stream`.

smokestacklightnin:
Closing this PR as it is superseded by #547. This branch is stale beyond repair.

**DO NOT DELETE THE BRANCH THAT THIS PR IS BASED ON**

Issue #487, opened 2024-08-12
Make corpus name part of the protocol rather than a chat protocol parameter


pmeier:
Initial discussion started in https://github.com/Quansight/ragna/pull/460#discussion_r1692061693. There we decided to just go with a `corpus_name: Optional[str]` as chat parameter. Same for LanceDB in #461.

However, we likely and up needing a handle for this outside of a chat as well. Think listing the available corpora, or listing the available metadata of a corpus. And if the the `corpus_name` is a fixed parameter for these endpoints, we should do the same for the protocol methods as well.

Thus, I propose we change the signatures from

- https://github.com/Quansight/ragna/blob/dd52e708ed77b779aadd34706883a4e90988644f/ragna/core/_components.py#L133
- https://github.com/Quansight/ragna/blob/dd52e708ed77b779aadd34706883a4e90988644f/ragna/core/_components.py#L142

to

- `def store(self, corpus_name: Optional[str], documents: list[Document])`
- `def retrieve(self, corpus_name: Optional[str], metadata_filter: MetadataFilter, prompt: str)`

Downside to this is that calling `SourceStorage.store` and `SourceStorage.retrieve` manually will be more annoying, since the `corpus_name` no longer has a default. But we are not optimizing for this workflow so I'm ok with this compromise.

nenb:
This seems fine to me. What would be the behaviour if no corpus name is passed?

pmeier:
> What would be the behaviour if no corpus name is passed?

Sorry, forgot to mention that. IMO the behavior should be "author defined". Meaning, the author should put in some behavior for the use case of the user not supplying a corpus name.

For our builtin source storages, we can simply continue to use the "global" corpus:

https://github.com/Quansight/ragna/blob/aea6b6706265e342e41682a88ea80bface4961ec/ragna/source_storages/_chroma.py#L39-L41

https://github.com/Quansight/ragna/blob/aea6b6706265e342e41682a88ea80bface4961ec/ragna/source_storages/_lancedb.py#L55-L57

If we at some point support #191, the behavior would change to "global corpus for the given embedding model".

nenb:
What should happen if a user wants to chat over multiple corpuses in a single chat?

Should a user be allowed to pass a list of corpus names when instantiating a `Chat` class?
https://github.com/Quansight/ragna/blob/2066dcdeb949060873a8552449b848b245e0e985/ragna/core/_rag.py#L148-L156

pmeier:
Fixed in #490.

PR #490, opened 2024-08-13
487 corpus name as protocol

nenb:
Addresses #487.


ragna/deploy/_api/core.py
```diff
@@ -289,6 +290,9 @@ async def prepare_chat(user: UserDependency, id: uuid.UUID) -> schemas.Message:
 
             welcome = schemas.Message.from_core(await core_chat.prepare())
 
+            if chat.prepared:
```
nenb, line 293: drive-by to tidy-up #482

ragna/source_storages/_demo.py
```diff
@@ -30,8 +30,9 @@ def display_name(cls) -> str:
 
     def __init__(self) -> None:
         self._storage: list[dict[str, Any]] = []
+        self._corpus_name = "Demo Corpus"
```
pmeier, line 33: Shouldn't `_storage` become `dict[str, list[dict[str, Any]]]`, i.e. mapping the corpus name to the data we store? And to simplify stuff, I would actually use `Optional[str]` as key to also handle the default `corpus_name=None` case.

tests/core/test_e2e.py
```diff
@@ -15,6 +15,7 @@ async def main(*, input, source_storage, assistant):
             input=input,
             source_storage=source_storage,
             assistant=assistant,
+            corpus_name="test-corpus",
```
pmeier, line 18: Unless we actually do something with `corpus_name`, should we just leave it as `None` here?

pmeier (approved):
LGTM if CI is green. Thanks Nick! If you find the time, maybe add a small test for the source storages that stores documents in corpus A and retrieves from corpus B. That should result in no sources being returned.

pmeier:
UI test failure is unrelated. Let's ignore it until we start merging to `main`.

Issue #488, opened 2024-08-12
corpus CLI


pmeier:
To be able to release the corpus API, we need a way for users to [CRUD](https://en.wikipedia.org/wiki/Create,_read,_update_and_delete) the corpora on a a given source storage. To make our lives a little easier, we are not targeting a UI for this yet, but should start with a CLI instead. Since the likely consumers of this feature are power users or admins, this is ok.

We can add a `ragna corpus` subcommand to the CLI. This in turn could have more subcommands:

- `ragna corpus list`: List all available corpora
- `ragna corpus ingest`: Ingest some documents into a given corpus (more on this later)
- `ragna corpus delete`: Delete a given corpus
- `ragna corpus metadata`: List all available metadata in a given corpus

---

Each command needs the source storage the action should be applied to. We have a few options here that we potentially can implement all:

- Add a `--source-storage` flag that accepts an import string similar to what we do in our config file, e.g. `--source-storage ragna.source_storages.Chroma`
- Allow passing a `--config` and only accept a `--source-storage` listed there. Also allow passing the source storage by its display name similar to the API, since we know the options.
- If we have a `--config` parameter and `--source-storage` is not passed, offer the user an interactive list of available source storages to select from

---

`ragna corpus ingest` is the trickiest of them. IMO a reasonable default behavior would be

- We recurse through a positionally supplied root directory
- We call `LocalDocument.from_path` on each path for which we have an available `DocumentHandler`

From there on it is just calling `SourceStorage.store` and injecting them into Ragna's database.

The tricky part comes when opening this up to other behavior than just the default:

- Can we assume that by calling `ragna corpus ingest` you want to ingest files from local disk?
- Should we enforce that every `Document` class has a `from_path` classmethod in order for us to create an arbitrary `Document` subclass if we have nothing more than a path?

I would love to hear from @nenb @blakerosenthal @dillonroach how this is done in the existing deployment.

nenb:
> I would love to hear from @nenb @blakerosenthal @dillonroach how this is done in the existing deployment.

- We simplified things a little bit by passing a file that contains a list of S3 prefixes (`s3://my-bucket/foo/bar/spam.pdf, ...`) as an additional argument to our CLI tool.

[...]

PR #491, opened 2024-08-14
488 corpus cli

nenb:
PoC that addresses #488.

Based on a [gist](https://gist.github.com/pmeier/cc1576839d3a25389080dc214640c397) produced by @pmeier, as well as comments from #488.

An example usage from the current branch is:
```
ragna corpus ingest $(find MY_DOCUMENT_DIRECTORY -type f) --metadata-fields metadata.json --corpus-name MY_CORPUS_NAME --user UNAME --report-failures 2>> failures.txt

cat metadata.json
# {"MY_DOCUMENT_DIRECTORY/MY_FILE1": {"test_field": true}, "MY_DOCUMENT_DIRECTORY/MY_FILE2": {"test_field": false}}
```

### DIFF
--- a/ragna/deploy/_cli/core.py
+++ b/ragna/deploy/_cli/core.py
@@ -1,3 +1,4 @@
+import json
 import subprocess
 import sys
 import time
@@ -8,10 +9,13 @@ import httpx
 import rich
 import typer
 import uvicorn
+from rich.progress import BarColumn, Progress, TextColumn, TimeRemainingColumn
 
 import ragna
 from ragna._utils import timeout_after
+from ragna.core._utils import default_user
 from ragna.deploy._api import app as api_app
+from ragna.deploy._api import database, orm
 from ragna.deploy._ui import app as ui_app
 
 from .config import ConfigOption, check_config, init_config
@@ -23,6 +27,8 @@ app = typer.Typer(
     add_completion=False,
     pretty_exceptions_enable=False,
 )
+corpus_app = typer.Typer()
+app.add_typer(corpus_app, name="corpus")
 
 
 def version_callback(value: bool) -> None:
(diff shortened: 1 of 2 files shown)


ragna/deploy/_cli/core.py
```diff
@@ -171,3 +177,215 @@ def wait_for_api() -> None:
...
+            default_factory=lambda: "ragna.source_storages.RagnaDemoSourceStorage",
+            help="Select source for storing vector embeddings.",
+        ),
+    ],
+    document_filesystem_handler: Annotated[
+        str,  # typer does not support ImportString
+        typer.Option(
+            "--document-filesystem-handler",
+            metavar="DOCUMENT_FILESYSTEM_HANDLER",
+            default_factory=lambda: "ragna.core.LocalDocument",
```
nenb, line 206: From #488: default to ingest files from local disk unless user supplies a different document filesystem handler.

ragna/deploy/_cli/core.py
```diff
@@ -171,3 +177,215 @@ def wait_for_api() -> None:
...
+    corpus_name: Annotated[
+        Optional[str],
+        typer.Option(
+            "--corpus-name",
+            metavar="CORPUS_NAME",
+            default_factory=lambda: None,
+            help="Name of the corpus to ingest the documents into.",
+        ),
+    ],
+    documents: list[Path],
```
nenb, line 228: From #488: Allow the user to pass an arbitrary amount of files.

ragna/deploy/_cli/core.py
```diff
@@ -171,3 +177,215 @@ def wait_for_api() -> None:
...
+    verbose: Annotated[bool, typer.Option("--verbose")] = False,
```
nenb, line 229: This is only used at the moment to capture files that were not ingested, but can be extended in various ways.

ragna/deploy/_cli/core.py
```diff
@@ -171,3 +177,215 @@ def wait_for_api() -> None:
...
+        document_filesystem_handler_class = DocumentFilesystemHandlerImport(
+            document_filesystem_handler_class=document_filesystem_handler
+        ).document_filesystem_handler_class
+    except ValidationError:
+        raise typer.BadParameter(
+            f"Unable to import document filesystem handler: {document_filesystem_handler}"
+        )
+
+    try:
+        getattr(document_filesystem_handler_class, "from_path")
```
nenb, line 248: From #488: Enforce that every `Document` class has a `from_path` classmethod.

ragna/deploy/_cli/core.py
```diff
@@ -171,3 +177,215 @@ def wait_for_api() -> None:
...
+        )
+
+    make_session = database.get_sessionmaker(database_url)
+
+    BATCH_SIZE = 10
+    bad_documents = []
+    with tqdm(total=len(documents) // BATCH_SIZE) as pbar:
+        for batch_number in range(0, len(documents), BATCH_SIZE):
+            document_collection = []
+            document_metadata_collection = []
```
nenb, line 261: Messy, and can probably be tidied a bit.

ragna/deploy/_cli/core.py
```diff
@@ -171,3 +177,215 @@ def wait_for_api() -> None:
...
+                            doc_instance.metadata,
+                        )
+                    )
+                except Exception:
+                    bad_documents.append(document)
+
+            if not document_collection:
+                continue
+
+            # TODO: add a check to see if the documents already exist in the database
```
nenb, line 279: Flagging a TODO

ragna/deploy/_cli/core.py
```diff
@@ -171,3 +177,215 @@ def wait_for_api() -> None:
...
+                )
+            except Exception:
+                bad_documents.extend(
+                    documents[batch_number : batch_number + BATCH_SIZE]
+                )
+
+            try:
+                source_storage_class().store(corpus_name, document_collection)
+            except Exception:
+                # TODO: should rollback the ragna database transaction here as well
```
nenb, line 297: Flagging a TODO

ragna/deploy/_cli/core.py
```diff
@@ -171,3 +177,215 @@ def wait_for_api() -> None:
...
+            pbar.update(1)
+
+        if verbose:
+            typer.echo(
+                f"Failed to ingest the following documents: {set(bad_documents)}"
+            )
+
+
+@corpus_app.command(help="List all available corpora.", name="list")
+def list_(
```
nenb, line 312: `list`, `delete` and `metadata` would probably require extending the interface/protocol methods in some way, so I didn't explore this any further.

ragna/deploy/_cli/core.py
```diff
@@ -171,3 +177,157 @@ def wait_for_api() -> None:
...
             process.communicate()
+
+
+@corpus_app.command(help="Ingest some documents into a given corpus.")
+def ingest(
+    documents: list[Path],
+    corpus_name: Optional[str] = typer.Option(
+        None, help="Name of the corpus to ingest the documents into."
+    ),
+    config: ConfigOption = "./ragna.toml",  # type: ignore[assignment]
```
nenb, line 188: Have now added `config` option as you suggested in #488

ragna/deploy/_cli/core.py
```diff
@@ -171,3 +177,157 @@ def wait_for_api() -> None:
...
+    user: Optional[str] = typer.Option(
+        None, help="User to link the documents to in the ragna database."
+    ),
+    verbose: bool = typer.Option(
+        False, help="Print the documents that could not be ingested."
+    ),
+    ignore_log: bool = typer.Option(
```
nenb, line 195: Short-term solution for checkpointing on files which have already been ingested. This option ignores the checkpoint logic.

Longer-term, I think this needs to be incorporated into the `ragna` database. Otherwise, it's very easy to add the same file to the vector database multiple times, which reduces performance (instead of returning 10 sources, you just get 10 copies of the same source), unless we have some logic to deal with this already.

ragna/deploy/_cli/core.py
```diff
@@ -171,3 +177,157 @@ def wait_for_api() -> None:
...
+        raise typer.BadParameter(
+            f"Could not connect to the database: {config.api.database_url}"
+        )
+
+    if user is None:
+        user = default_user()
+    with make_session() as session:
+        user_id = database._get_user_id(session, user)
+
+    # Log (JSONL) for recording which files previously added to vector database.
```
nenb, line 219: checkpoint logic, stored in a JSONL file on filesystem where CLI is run

ragna/deploy/_cli/core.py
nenb: I'v added a short-term checkpointing solution that stores all paths in a local file.

ragna/deploy/_cli/core.py
```diff
@@ -181,3 +188,175 @@ def wait_for_api() -> None:
...
+        )
+
+    try:
+        make_session = database.get_sessionmaker(config.api.database_url)
+    except Exception:
+        raise typer.BadParameter(
+            f"Could not connect to the database: {config.api.database_url}"
+        )
+
+    if metadata_fields:
```
nenb, line 230: This file should contain a mapping from individual filepaths to a dictionary of metadata fields for each filepath.

pmeier (approved):
I don't love it, but I don't want to spend more time on this initial feature. I added a few small changes:

- Factor out the corpus commands in a separate module
- Switch the input arguments to use `typing.Annotated` as is recommended and we are doing so in our CLI already
- Add a banner whenever some uses a subcommand of `ragna corpus` informing them that this part of the CLI is experimental and subject to change.

I'll open an issue with all of the comments I have about the functionality that we don't need to do here. Thanks Nick!

pmeier:
> I'll open an issue with all of the comments I have about the functionality that we don't need to do here.

Actually, I think most of it is already documented in #488.

Issue #493, opened 2024-08-20
Support for metadata filter suggestions in the web UI


pmeier:
In #484 we hardcode the available metadata. We cannot release with that. Instead we need a way to communicate this information from the backend to the web UI. For this we need

1. An endpoint on the API, i.e. `GET /corpuses/{name}/metadata`
2. A new abstract method on the `SourceStorage` class, e.g. `list_metadata` (name TBD). The return value should be `dict[str, tuple[type, list[Any]]]` with the keys being the available metadata keys and the values being a two-tuple of the type and the available values.
   - The `type` above might also be a `str`, e.g. `"int"`, `"float"`, etc., if that makes it easier.
   - The source storage might opt to return an empty list for the available values to indicate that no hints for the values are available.
   
   This function is potentially pretty expensive as one has to query the full database and potentially extract unique values from it. Thus, while we don't need to have it from the first version, we should have caching in mind.

   Lastly, I'm not sure yet if we want to make it a strictly abstract, i.e. decorating with `@abstractmethod`, because that would require everyone to implement this even if one doesn't want to work with corpuses. Instead we could also leave it undecorated and `raise NotImplementedError` instead and thus pushing the check to runtime. Thoughts?

In addition, 2. also has to be implemented on builtin source storages.

pmeier:
Afterthought to 1.: in #487 we decided to use `None` as sentinel for the default corpus as decided by the source storage. Not sure how this can work through the REST API though as `{name}` probably has to be a `str` and cannot be omitted. @nenb what do you think of removing the `None` option and instead use a string sentinel, e.g. `"default"`?

nenb:
>  @nenb what do you think of removing the None option and instead use a string sentinel, e.g. "default"?

This seems fine to me.

The rest of the issue also seems fine to me - let's leave `list_metadata` undecorated for the reason that you outlined.

PR #494, opened 2024-08-20
fix doc builds after recent changes

pmeier:
This is to get the CI green again on the `corpus-dev` branch. We still need to add actual documentation for the changes we have made. I'll flag some of them below.

### DIFF
--- a/docs/examples/gallery_streaming.py
+++ b/docs/examples/gallery_streaming.py
@@ -119,9 +119,10 @@ chat = (
         "/chats",
         json={
             "name": "Tutorial REST API",
-            "documents": [document],
+            "input": [document],
             "source_storage": source_storages.RagnaDemoSourceStorage.display_name(),
             "assistant": DemoStreamingAssistant.display_name(),
+            "corpus_name": None,
             "params": {},
         },
     )
--- a/ragna/core/_rag.py
+++ b/ragna/core/_rag.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import collections.abc
 import contextlib
 import datetime
 import inspect
@@ -12,8 +13,8 @@ from typing import (
     AsyncIterator,
     Awaitable,
     Callable,
+    Collection,
     Generic,
-    Iterable,
     Iterator,
     Optional,
     Type,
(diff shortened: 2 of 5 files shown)


ragna/core/_rag.py
```diff
@@ -241,11 +268,22 @@ async def answer(self, prompt: str, *, stream: bool = False) -> Message:
...
+            str,
+            Path,
+            Collection[Union[Document, str, Path]],
+        ],
     ) -> tuple[Optional[list[Document]], Optional[MetadataFilter], bool]:
         if isinstance(input, MetadataFilter) or input is None:
             return None, input, True
 
+        if not isinstance(input, collections.abc.Collection):
+            input = [input]
```
pmeier, line 285: Driveby to also allow single documents.

ragna/core/_rag.py
```diff
@@ -82,21 +83,34 @@ def _load_component(
...
+            Document,
+            str,
+            Path,
+            Collection[Union[Document, str, Path]],
+        ],
         *,
         source_storage: Union[Type[SourceStorage], SourceStorage],
         assistant: Union[Type[Assistant], Assistant],
-        corpus_name: Optional[str],
+        corpus_name: Optional[str] = None,
```
pmeier, line 97: @nenb I've added a default value here, because I don't want to force users to always set this one.

docs/examples/gallery_streaming.py
```diff
@@ -119,9 +119,10 @@ def answer(self, messages):
         "/chats",
         json={
             "name": "Tutorial REST API",
-            "documents": [document],
+            "input": [document],
             "source_storage": source_storages.RagnaDemoSourceStorage.display_name(),
             "assistant": DemoStreamingAssistant.display_name(),
+            "corpus_name": None,
```
pmeier, line 125: We probably want to have this as default value as well.

nenb: Is this still valid after your comments in #493?

nenb (approved):
Seems fine, but I'm not 💯 sure if you want to set the `corpus_name` as `None` or use a sentinel after #493.

pmeier:
Superseded by #498.

Issue #492, opened 2024-08-20
support selecting corpus in web UI


pmeier:
With #487 merged, we need a way to select the corpus we want in the web UI. I suggest we just assume we have < 10 corpuses and thus use a dropdown menu. To fill it we need two more things:

1. An endpoint on the API, i.e. `GET /corpuses`
2. A new abstract method on the `SourceStorage` class, e.g. `list_corpuses` (name TBD).

   I'm not sure yet if we want to make it a strictly abstract, i.e. decorating with `@abstractmethod`, because that would require everyone to implement this even if one doesn't want to work with corpuses. Instead we could also leave it undecorated and `raise NotImplementedError` instead and thus pushing the check to runtime. Thoughts?

In addition, 2. also has to be implemented on builtin source storages.

PR #495, opened 2024-08-20
List corpuses

blakerosenthal:
Addresses https://github.com/Quansight/ragna/issues/492

### DIFF
--- a/ragna/deploy/_api/core.py
+++ b/ragna/deploy/_api/core.py
@@ -151,7 +151,15 @@ def app(*, config: Config, ignore_unavailable_components: bool) -> FastAPI:
         source_storage: Optional[str] = None,
     ) -> dict[str, list[str]]:
         if source_storage is not None:
-            source_storages = [get_component(source_storage)]
+            component = components_map.get(source_storage)
+            if component is None or not isinstance(component, SourceStorage):
+                raise RagnaException(
+                    "Unknown source storage",
+                    display_name=source_storage,
+                    http_status_code=status.HTTP_422_UNPROCESSABLE_ENTITY,
+                    http_detail=RagnaException.MESSAGE,
+                )
+            source_storages = [component]
         else:
             source_storages = [
                 source_storage
--- a/ragna/deploy/_api/core.py
+++ b/ragna/deploy/_api/core.py
@@ -1,6 +1,6 @@
 import contextlib
 import uuid
-from typing import Annotated, Any, AsyncIterator, Iterator, Type, Union, cast
+from typing import Annotated, Any, AsyncIterator, Iterator, Optional, Type, Union, cast
 
 import aiofiles
 from fastapi import (
@@ -146,12 +146,18 @@ def app(*, config: Config, ignore_unavailable_components: bool) -> FastAPI:
         )
 
     @app.get("/corpuses")
-    async def get_corpuses(_: UserDependency) -> dict[str, list[str]]:
-        source_storages = [
-            source_storage
-            for source_storage in components_map.values()
-            if isinstance(source_storage, SourceStorage)
-        ]
+    async def get_corpuses(
+        _: UserDependency,
+        source_storage: Optional[str] = None,
+    ) -> dict[str, list[str]]:
+        if source_storage is not None:
+            source_storages = [get_component(source_storage)]
+        else:
+            source_storages = [
+                source_storage
+                for source_storage in components_map.values()
+                if isinstance(source_storage, SourceStorage)
+            ]
 
         return {
             source_storage.display_name(): source_storage.list_corpuses()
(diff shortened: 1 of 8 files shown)


blakerosenthal:
@pmeier Is this about right? I like the idea of raising `NotImplemented` on `source_storage.list_corpuses()` in case the user decides to ignore corpuses altogether.

Also right now the `/corpuses` endpoint returns the list of corpus names directly from the source_storages. I have two questions about this:
1. Should the api call distinguish which source storage the corpus name belongs to? E.g. `["LanceDB/<corpus_name>", "ChromaDB/<corpus_name>"]`? Maybe actually returning a JSON of `{"LanceDB": [<corpus_name>], "ChromaDB": [<corpus_name>]}` would be better?
2. For the built in source storages, the default corpus name is just the embedding ID. Maybe `list_corpuses()` should translate this to a prettier string like `"default"` for the sake of the UI.

Finally, should I add the actual UI changes as part of this issue, or will that be a separate issue?

pmeier:
> I like the idea of raising `NotImplemented` on `source_storage.list_corpuses()` in case the user decides to ignore corpuses altogether.

On second thought, we should probably raise a descriptive `RagnaException` instead. We need to tell users what they have to do if they encounter this error and then exit gracefully.

> 1. Should the api call distinguish which source storage the corpus name belongs to? E.g. `["LanceDB/<corpus_name>", "ChromaDB/<corpus_name>"]`? Maybe actually returning a JSON of `{"LanceDB": [<corpus_name>], "ChromaDB": [<corpus_name>]}` would be better?

You are right, that needs to happen. I would go for the JSON approach where the key is the `.display_name()`.


> 2\. For the built in source storages, the default corpus name is just the embedding ID. Maybe `list_corpuses()` should translate this to a prettier string like `"default"` for the sake of the UI.

Yes, that needs to happen on the source storage itself. I'm also inclined to remove the feature of using the embedding ID as name for now. There is no use and it makes it more complicated. So feel free to just remove the embedding ID and just use `"default"`.

pmeier:
> Finally, should I add the actual UI changes as part of this issue, or will that be a separate issue?

Up to you. When using two PRs we can merge this PR soon and don't have to wait for #484 to be merged.

pmeier (commented):
Looking good. I'll review properly when the open questions you had are addressed.

blakerosenthal:
> Up to you. When using two PRs we can merge this PR soon and don't have to wait for #484 to be merged.

Okay, let's keep them separate then

tests/deploy/api/test_e2e.py
```diff
@@ -86,6 +89,9 @@ def test_e2e(tmp_local_root, multiple_answer_chunks, stream_answer, corpus_name)
         assert len(chat["messages"]) == 1
         assert chat["messages"][-1] == message
 
+        corpuses = client.get("/corpuses").raise_for_status().json()
+        assert corpuses == {source_storage: [str(corpus_name)]}
```
pmeier, line 93:
```suggestion
        assert corpuses == {source_storage: [corpus_name]}
```

ragna/deploy/_api/core.py
```diff
@@ -145,6 +145,19 @@ async def get_components(_: UserDependency) -> schemas.Components:
             ],
         )
 
+    @app.get("/corpuses")
+    async def get_corpuses(_: UserDependency) -> dict[str, list[str]]:
```
pmeier, line 149: Following the discussion in https://github.com/Quansight/ragna/issues/493#issuecomment-2301776184 should we introduce a `source_storage` query parameter?

```suggestion
    async def get_corpuses(_: UserDependency, source_storage: Optional[str] = None) -> dict[str, list[str]]:
```

And with that we can do

```python
if source_storage is not None:
    source_storages = [components_map[source_storage]]
else:
    ...
```

Meaning, if the user does `GET /corpuses` they get everything, but they can do `GET /corpuses?source_storage=Chroma` to pre-filter it. Still, the output format should be the same in both cases.

pmeier (commented):
Thanks Blake, almost LGTM to me!

ragna/deploy/_api/core.py
blakerosenthal: I like that. I'll push that up.

tests/deploy/api/test_e2e.py
```diff
@@ -86,6 +90,21 @@ def test_e2e(tmp_local_root, multiple_answer_chunks, stream_answer, corpus_name)
         assert len(chat["messages"]) == 1
         assert chat["messages"][-1] == message
 
+        corpuses = client.get("/corpuses").raise_for_status().json()
+        assert corpuses == {source_storage: [corpus_name]}
+
+        corpuses = (
+            client.get(f"/corpuses?source_storage={source_storage}")
```
pmeier, line 97: Let the library handle the URL building.

```suggestion
            client.get("/corpuses", params={"source_storage": source_storage})
```

Same for the occurrence below.

pmeier (approved):
Thanks Blake. FYI: I pushed a small cleanup commit.

PR #496, opened 2024-08-21
gracefully stop API when started by ragna ui

pmeier:
By default Python does not handle the SIGTERM signal. Thus, currently, when we send a SIGTERM to the process that was started by `ragna ui --start-api`, the `finally` branch that is supposed to stop the API is never run:

https://github.com/Quansight/ragna/blob/2066dcdeb949060873a8552449b848b245e0e985/ragna/deploy/_cli/core.py#L170-L173

Meaning, the Ragna API stays up:

```
❯ ragna ui --start-api &; sleep 10 && kill $! && sleep 10 && ps aux | grep ragna
[1] 22747
INFO:   RagnaDemoAuthentication: You can log in with any username and a matching password.
INFO:     Started server process [22773]
INFO:     Waiting for application startup.
INFO:     Application startup complete.
INFO:     Uvicorn running on http://127.0.0.1:31476 (Press CTRL+C to quit)
INFO:     127.0.0.1:54172 - "GET / HTTP/1.1" 200 OK
Launching server at http://127.0.0.1:31477
INFO:     127.0.0.1:58566 - "GET / HTTP/1.1" 200 OK
[1]  + 22747 terminated  ragna ui
philip     22773 22.8  0.5 1630504 358968 pts/0  SNl  11:01   0:04 /home/philip/.conda/envs/ragna-deploy-dev/bin/python3.9 -m ragna api --config ./ragna.toml --no-ignore-unavailable-components
```

This pronounced in our test suite where each UI test spawns the Ragna API, but it never gets cleaned up

```
❯ pytest tests/deploy/ui --quiet && ps aux | grep ragna
..                                                                                                 [100%]
2 passed in 11.99s
philip     25447 35.7  0.2 1139644 195568 pts/0  Sl   11:11   0:03 /home/philip/.conda/envs/ragna-deploy-dev/bin/python -m ragna api --config /tmp/pytest-of-philip/pytest-26/test_health_chromium_0/ragna.toml --ignore-unavailable-components
philip     25521 68.2  0.3 1290196 199060 pts/0  Sl   11:11   0:03 /home/philip/.conda/envs/ragna-deploy-dev/bin/python -m ragna api --config /tmp/pytest-of-philip/pytest-26/test_start_chat_chromium_0/ragna.toml --ignore-unavailable-components
```

This PR turns the SIGTERM signal in a regular system exit for `ragna ui` and thus enables graceful shutdown.

```
[...]

PR #497, opened 2024-08-21
handle zombie processes in docs

pmeier:
Follow-up to #496. If the documentation build errors and we never get to calling `RestApi.stop()`, we have a zombie RestAPI process. However, here it is even worse than for the tests in #496: since we are not randomizing the port number, the next time we build the docs, the old REST API is still available and will be picked up instead of starting a fresh one invalidating most of the assumptions that we make, e.g. no chats available after startup.

### DIFF
--- a/docs/tutorials/gallery_custom_components.py
+++ b/docs/tutorials/gallery_custom_components.py
@@ -322,7 +322,7 @@ config = Config(
 rest_api = ragna_docs.RestApi()
 
 client, document = rest_api.start(config, authenticate=True, upload_document=True)
-raise Exception
+
 # %%
 # To pass custom parameters, define them in the `params` mapping when creating a new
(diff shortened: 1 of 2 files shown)


nenb (approved):
Nitpick: Does #496 apply here as well? Can something like a SIGTERM signal be sent here as well, and if so is it worth also turning it into a regular system exit?

pmeier:
> Does #496 apply here as well? Can something like a SIGTERM signal be sent here as well, and if so is it worth also turning it into a regular system exit?

Not really. I'm specifically targeting errors that happen during build. To get a SIGTERM signal one would need to manually kill a doc build that is for some reason running in the background. I guess this use case is rare at best. Normally one jsut CTRL+C in the terminal, which is handled by the current solution.

PR #498, opened 2024-08-21
improve corpus UX and fix docs

pmeier:
- Fix the doc builds so CI is green again. This does _not_ add proper docs just stubs so we can see if other PRs to the `corpus-dev` branch break the documentation. This supersedes #494.
- Remove the `corpus_name=None`  sentinel and use `corpus_name="default"` instead. See https://github.com/Quansight/ragna/issues/493#issuecomment-2298153629
- Add the default value for `corpus_name="default"` when creating a chat to avoid forcing users to pass something. The default handling is thus pushed to the author of the source storage.
- Add the default value for `input=None` when creating a chat. This way users with a corpus of documents can just ask away. I've added proper error handling in case we are unable to retrieve any sources.
- Add support for passing a single document or path as input when creating a chat

### DIFF
--- a/ragna/source_storages/_demo.py
+++ b/ragna/source_storages/_demo.py
@@ -3,6 +3,8 @@ import textwrap
 import uuid
 from typing import Any, Callable, Optional, cast
 
+from fastapi import status
+
 from ragna.core import (
     Document,
     MetadataFilter,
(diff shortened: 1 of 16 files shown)


ragna/source_storages/_demo.py
pmeier, line 1: Changes in this file are mostly to properly support the corpus feature. Previously the `corpus_name` was ignored and everything was stored in the global corpus instead.

tests/source_storages/test_source_storages.py
pmeier, line 1: Just minor cleanup in this file. Instead of hardcoding the corpus name in multiple places, I added a `corpus_name` variable and used that in the appropriate places.

pmeier:
This was "reviewed" already in #493. Thus, I'm going to self-merge when CI is green.

PR #499, opened 2024-08-23
UI extension for new Corpus feature

nenb:
_(Separate branch from #484 to ease initial review.)_


ragna/deploy/_ui/api_wrapper.py
```diff
@@ -56,6 +56,40 @@ async def auth(self, username, password):
     def update_auth_header(self):
         self.client.headers["Authorization"] = f"Bearer {self.auth_token}"
 
+    async def get_corpus_names(self):
```
nenb, line 59: Once #495 is merged.

ragna/deploy/_ui/api_wrapper.py
```diff
@@ -56,6 +56,40 @@ async def auth(self, username, password):
...
+        # TODO: replace with actual endpoint
+        # return (await self.client.get("/corpuses")).raise_for_status().json()
+        return {
+            "Chroma": ["corpus1", "corpus2"],
+            "LanceDB": ["corpus1"],
+        }
+
+    async def get_corpus_metadata(self, corpus_name=None):
+        # TODO: replace with actual endpoint, and add corpus_name as query parameter
```
nenb, line 68: Once #495 is merged.

ragna/deploy/_ui/components/file_uploader.py
```diff
@@ -5,13 +5,17 @@
...
+from .. import styles as ui
+
 
 class FileUploader(ReactiveHTML, Widget):  # type: ignore[misc]
     allowed_documents = param.List(default=[])
     allowed_documents_str = param.String(default="")
 
     file_list = param.List(default=[])
 
+    height_upload_container = param.String(default=ui.FILE_CONTAINER_HEIGHT)
```
nenb, line 17: The [modal resizing functionality](https://github.com/Quansight/ragna/blob/ef6c6a5267315caac7b4646d0bca1938b7f9439a/ragna/deploy/_ui/js.py#L12-L15) has been broken on `main` for a while (try expand the 'Advanced Configuration' and see what happens).

I spent some time trying to get it to work, but in the end the only solution I could get to work was by introducing another `param` object for the height of the upload container. This parameter is watched, and when the Advanced Configuration card is expanded, this parameter is updated to a smaller value. See also `modal_configuration.py`.

ragna/deploy/_ui/components/file_uploader.py
```diff
@@ -101,10 +105,6 @@ def perform_upload(self, event=None, after_upload_callback=None):
                 .fileUploadDropArea.draggedOver {
                     border-width: 3px;
                 }
-
```
nenb, line 104: Removed to reduce the number of resize operations on the height of the upload container (see previous comment on how I have introduced a new parameter for this). This has negligible UX impact IMO. The basic functionality was to make the upload container smaller when a 'pill' (icon for a newly upload file) was present below. I have now introduced some more spacing and a scroll (see below).

ragna/deploy/_ui/components/file_uploader.py
```diff
@@ -115,15 +115,20 @@ def perform_upload(self, event=None, after_upload_callback=None):
                 .fileListContainer {
                     display:flex;
                     flex-direction: row;
-                    height: 44px;
-                    overflow:scroll;
-                    padding-top:14px;
+                    flex-wrap: wrap;
```
nenb, line 118: Introduces wrapping so that if the pills are wider than the modal, they 'wrap' around to the next line, and scroll along the Y-axis. I have removed the scroll on the X-axis as I found having too scrolls to be hard to follow. I have also added more spacing to compensate for the fact that the file upload container no longer shrinks when a pill is added (see previous comment).

ragna/deploy/_ui/components/file_uploader.py
```diff
@@ -115,15 +115,20 @@ def perform_upload(self, event=None, after_upload_callback=None):
...
+                    margin-bottom: 10px;
+                    padding-top:5px;
                     padding-left:6px;
+                    padding-bottom:5px;
                 }
                 
                 .chat_document_pill {
                     background-color: rgb(241,241,241);
-                    margin-top:0px; 
+                    margin-top:5px;
```
nenb, line 131: To make things a bit more readable when many pills, introduce a small amount of white space between them.

ragna/deploy/_ui/css/main.css
```diff
@@ -28,7 +28,7 @@ div#content {
 /* Fix the size of the modal */
 #pn-Modal {
   --dialog-width: 800px !important;
-  --dialog-height: 500px !important;
+  --dialog-height: 700px !important;
```
nenb, line 31: Lots more fields now, need to make the default size bigger.

ragna/deploy/_ui/css/modal_configuration/intslider.css
```diff
@@ -4,7 +4,7 @@
 }
 
 :host(.modal_configuration_int_slider) {
-  margin-top: 20px;
+  margin-top: 0px;
```
nenb, line 7: As there are many more fields, reducing white space to try and fit everything into one modal.

ragna/deploy/_ui/js.py
```diff
@@ -1,20 +1,10 @@
...
-# Reset modal size
-def reset_modal_size(width, height):
+def set_modal_size(width, height):
     return f"""
     document.getElementById("pn-Modal").style.setProperty("--dialog-height","{height}px", "important")
     document.getElementById("pn-Modal").style.setProperty("--dialog-width","{width}px", "important")
     """
 
 
-JS_TOGGLE_CARD = f"""
```
nenb, line 12: Removing because no longer works on `main` (see previous comments where a new `param` object has been introduced for the upload container height).

ragna/deploy/_ui/main_page.py
```diff
@@ -38,6 +42,9 @@ def __init__(self, api_wrapper, template):
 
     async def refresh_data(self):
         self.chats = await self.api_wrapper.get_chats()
+        self.components = await self.api_wrapper.get_components()
```
nenb, line 45: On UI startup, we retrieve all the information that is needed to create a chat from the UI.

Note that `refresh_data` is also called after a chat has been created, and so these fields should always be accurate.

ragna/deploy/_ui/main_page.py
```diff
@@ -57,10 +64,19 @@ def after_update_chats(self):
...
 
-    # Modal and callbacks
-    def open_modal(self):
+    async def open_modal(self, event):
+        if (
+            self.components is None
+            or self.corpus_metadata is None
+            or self.corpus_names is None
+        ):
+            await self.refresh_data()
```
nenb, line 73: On UI startup, `refresh_data` is called by `asyncio.ensure_future`. It's possible (but probably unlikely) that it will not have been called by the time a user clicks on the 'New Chat' button, and so we introduce a guard here.

We also need to convert this function into an async function as a result.

ragna/deploy/_ui/main_page.py
```diff
@@ -57,10 +64,19 @@ def after_update_chats(self):
...
+
         self.modal = ModalConfiguration(
             api_wrapper=self.api_wrapper,
+            components=self.components,
```
nenb, line 77: New fields required for building the modal.

ragna/deploy/_ui/main_page.py
```diff
@@ -26,7 +30,7 @@ def __init__(self, api_wrapper, template):
 
         self.left_sidebar = LeftSidebar(api_wrapper=self.api_wrapper)
         self.left_sidebar.on_click_chat = self.on_click_chat
-        self.left_sidebar.on_click_new_chat = lambda event: self.open_modal()
+        self.left_sidebar.on_click_new_chat = self.open_modal
```
nenb, line 33: `open_modal` is now an async function (see below for explanation).

ragna/deploy/_ui/left_sidebar.py
```diff
@@ -25,12 +25,12 @@ def __init__(self, api_wrapper, **params):
             on_error=lambda x: print(f"error sync on {x}"),
         )
 
-    def trigger_on_click_new_chat(self, event):
+    async def trigger_on_click_new_chat(self, event):
```
nenb, line 28: `on_click_new_chat` now needs to be an async function (see `main_page.py` and `self.open_modal` for details).

ragna/deploy/_ui/central_view.py
```diff
@@ -186,20 +187,49 @@ def on_click_chat_info_wrapper(self, event):
         if self.on_click_chat_info is None:
             return
 
-        pills = "".join(
-            [
-                f"""<div class='chat_document_pill'>{d['name']}</div>"""
-                for d in self.current_chat["metadata"]["documents"]
-            ]
-        )
+        # see _api/schemas.py for `input` type definitions
```
nenb, line 190: Display in right sidebar depending on what has been used as `input` for the chat.

ragna/deploy/_ui/modal_configuration.py
```diff
@@ -77,15 +81,24 @@ class ModalConfiguration(pn.viewable.Viewer):
...
         self.api_wrapper = api_wrapper
 
+        self.corpus_names = corpus_names
+        self.corpus_metadata = corpus_metadata
+
         upload_endpoints = self.api_wrapper.upload_endpoints()
 
         self.chat_name_input = pn.widgets.TextInput.from_param(
-            self.param.chat_name,
+            self.param.chat_name, name=""
```
nenb, line 101: Driveby to remove the fact that we are displaying the `chat_name` twice here.

ragna/deploy/_ui/modal_configuration.py
```diff
@@ -115,12 +128,34 @@ def __init__(self, api_wrapper, **params):
...
-        else:
+        self.corpus_or_upload_radiobutton = pn.widgets.RadioButtonGroup.from_param(
+            self.param.corpus_or_upload,
+            button_style="outline",
+            button_type="primary",
+        )
+
+        # self.metadata_filters_builder = MetadataFiltersBuilder()
+
+        self.create_config(components)
```
nenb, line 139: There was a race-condition in the old related to the fact that `model_section` (which contained the logic for `create_config`) was an async function. Depending on how `model_section` was scheduled, the config could actually end up being created after a bunch of functions which depended on it, resulting in a blank modal.

I've simplified things by extracting the logic into a single synchronous function. There's nothing expensive in this function, and so it should be fine to do this. It has the benefit of removing the race-condition and making the code easier to reason about.

ragna/deploy/_ui/modal_configuration.py
```diff
@@ -155,13 +190,8 @@ def change_upload_files_label(self, mode="normal"):
         else:
             self.upload_files_label.object = "<b>Upload files</b> (required)"
 
-    async def model_section(self):
```
nenb, line 158: See previous comment on why this has been split into two synchronous functions (`create_config` and `model_section`).

Note that the only reason it was async previously was to make a call to `get_components`, which has now been moved to `main_page.py`.

ragna/deploy/_ui/modal_configuration.py
```diff
@@ -198,8 +229,7 @@ async def model_section(self):
             ),
         )
 
-    @pn.depends("config", "config.assistant_name", "config.source_storage_name")
-    def advanced_config_ui(self):
+    def advanced_config_upload(self):
```
nenb, line 232: Split into two functions (one for the upload tab, and another for the corpus tab), as a bunch of the options are not relevant for the corpus tab.

ragna/deploy/_ui/modal_configuration.py
```diff
@@ -219,12 +249,6 @@ def advanced_config_ui(self):
         card = pn.Card(
             pn.Row(
                 pn.Column(
-                    pn.pane.HTML(
```
nenb, line 222: Unfortunately removed to make space so that everything could fit in one modal.

ragna/deploy/_ui/modal_configuration.py
```diff
@@ -245,12 +269,6 @@ def advanced_config_ui(self):
                     width_policy="max",
                 ),
                 pn.Column(
-                    pn.pane.HTML(
```
nenb, line 248: Unfortunately removed to make space so that everything could fit in one modal.

ragna/deploy/_ui/modal_configuration.py
```diff
@@ -272,7 +290,7 @@ def advanced_config_ui(self):
                         "border-left": "1px solid var(--neutral-stroke-divider-rest)"
                     },
                 ),
-                height=250,
+                width=800,
```
nenb, line 293: Height is dynamic (see earlier comments on the new `param` object introduced) so no longer needs to be fixed. Width is pinned to modal width just to be on the safe side.

ragna/deploy/_ui/modal_configuration.py
```diff
@@ -297,41 +315,122 @@ def toggle_card(event):
...
         toggle_button = pn.widgets.Button(
-            name="Advanced Configurations   ▶",
+            name="Advanced Configuration of Assistants and Source Storages   ▶",
             button_type="light",
             css_classes=["modal_configuration_toggle_button"],
         )
-
         toggle_button.on_click(toggle_card)
 
-        toggle_button.js_on_click(
```
nenb, line 307: Removed as this is broken on `main` and I couldn't get it to work (see previous comments).

ragna/deploy/_ui/modal_configuration.py
```diff
@@ -297,41 +315,122 @@ def toggle_card(event):
...
-            args={"card": card},
-            code=js.JS_TOGGLE_CARD,
+        return pn.Column(toggle_button, card)
+
+    def advanced_config_corpus(self):
```
nenb, line 326: New function for `corpus` tab (but very similar to the `upload` tab function, minus a few options).

ragna/deploy/_ui/modal_configuration.py
```diff
@@ -199,10 +228,7 @@ async def model_section(self):
         )
 
     @pn.depends("config", "config.assistant_name", "config.source_storage_name")
-    def advanced_config_ui(self):
-        if self.config is None:
-            return
-
+    def advanced_config(self, is_corpus=False):
```
nenb, line 231: We now need to support two different configs - one for corpus, and one for file upload, as they have different features.

ragna/deploy/_ui/modal_configuration.py
```diff
@@ -216,69 +242,82 @@ def advanced_config_ui(self):
             *(["disabled"] if disabled_assistant else []),
         ]
 
-        card = pn.Card(
```
nenb, line 219: Not relevant to corpus.

ragna/deploy/_ui/modal_configuration.py
```diff
@@ -216,69 +242,82 @@ def advanced_config_ui(self):
...
-                        self.config.param.chunk_overlap,
-                        name="Chunk Overlap",
-                        bar_color=ui.MAIN_COLOR,
-                        css_classes=source_storage_css_classes,
-                        width_policy="max",
-                        disabled=disabled_source_storage,
-                    ),
-                    margin=(0, 20, 0, 0),
-                    width_policy="max",
+        if is_corpus:
```
nenb, line 245: Unfortunately explanatory text removed to make space so that everything could fit in one modal.

ragna/deploy/_ui/modal_configuration.py
```diff
@@ -216,69 +242,82 @@ def advanced_config_ui(self):
...
+            card = pn.Card(
+                pn.widgets.IntSlider.from_param(
+                    self.config.param.max_new_tokens,
+                    bar_color=ui.MAIN_COLOR,
+                    css_classes=assistant_css_classes,
+                    disabled=disabled_assistant,
+                    margin=(0, 0, 0, 11),
                 ),
-                pn.Column(
```
nenb, line 247: Not relevant to corpus.

ragna/deploy/_ui/modal_configuration.py
```diff
@@ -216,69 +242,82 @@ def advanced_config_ui(self):
...
+                            bar_color=ui.MAIN_COLOR,
+                            css_classes=source_storage_css_classes,
+                            width_policy="max",
+                            disabled=disabled_source_storage,
+                        ),
+                        margin=(0, 20, 0, 0),
                         width_policy="max",
-                        disabled=disabled_source_storage,
                     ),
-                    pn.widgets.IntSlider.from_param(
```
nenb, line 261: Some weird `git` stuff going on here, don't think anything has actually changed, just indented.

ragna/deploy/_ui/modal_configuration.py
```diff
@@ -296,42 +335,71 @@ def toggle_card(event):
             else:
                 toggle_button.name = toggle_button.name.replace("▶", "▼")
 
-        toggle_button = pn.widgets.Button(
```
nenb, line 299: Moved earlier as name will be different between corpus and upload tabs.

ragna/deploy/_ui/modal_configuration.py
```diff
@@ -198,11 +227,7 @@ async def model_section(self):
             ),
         )
 
-    @pn.depends("config", "config.assistant_name", "config.source_storage_name")
```
nenb, line 201: Removed as these now exist on `corpus_or_upload_config`.

ragna/deploy/_ui/modal_configuration.py
```diff
@@ -296,42 +334,71 @@ def toggle_card(event):
...
+        "config.assistant_name",
+        "config.source_storage_name",
+    )
+    def corpus_or_upload_config(self):
+        if self.corpus_or_upload == USE_CORPUS_LABEL:
+            return self.advanced_config(is_corpus=True)
+        else:
+            return self.advanced_config(is_corpus=False)
+
+    @pn.depends("advanced_config_collapsed", watch=True)
```
nenb, line 353: Watches for when the advanced config is expanded, and then reduces the size of the file upload container (see comments in `file_uploader.py`)

ragna/deploy/_ui/modal_configuration.py
```diff
@@ -296,42 +334,71 @@ def toggle_card(event):
...
+    def shrink_upload_container_height(self):
+        if self.advanced_config_collapsed:
+            self.document_uploader.height_upload_container = "160px"
+        else:
+            self.document_uploader.height_upload_container = "90px"
+
+    @pn.depends(
```
nenb, line 360: Updates the metadata filter or upload row, depending on which tab has been selected.

Also required to depend on the `source_storage` as this will determine the list of `corpus_names` and `corpus_metadata` available.

ragna/deploy/_ui/modal_configuration.py
```diff
@@ -99,6 +112,12 @@ def __init__(self, api_wrapper, **params):
         )
         self.cancel_button.on_click(self.cancel_button_callback)
 
+        self.corpus_name_input = pn.widgets.TextInput(
```
nenb, line 115: New field required when uploading documents - also pass the name of the corpus to upload to.

ragna/deploy/_ui/modal_configuration.py
```diff
@@ -296,42 +340,69 @@ def toggle_card(event):
...
+                pn.pane.HTML("<b>Corpus Name</b>"),
+                self.corpus_name_input,
+                self.upload_files_label,
+                self.upload_row,
+            )
+
     def __panel__(self):
         return pn.Column(
             pn.pane.HTML(
-                f"""<h2>Start a new chat</h2>
```
nenb, line 317: Unfortauntely text removed to make space.

ragna/deploy/_ui/modal_configuration.py
```diff
@@ -296,42 +340,69 @@ def toggle_card(event):
...
-            self.upload_files_label,
-            self.upload_row,
+            self.corpus_or_upload_row,
             pn.Row(self.cancel_button, self.start_chat_button),
-            min_height=ui.CONFIG_MODAL_MIN_HEIGHT,
+            min_height=ui.CONFIG_MODAL_HEIGHT,
             min_width=ui.CONFIG_MODAL_WIDTH,
-            sizing_mode="stretch_both",
-            height_policy="max",
+            styles={"overflow-y": "hidden"},
```
nenb, line 407: No need to have multiple y-scrolls on the modal, so disable to be certain.

ragna/deploy/_ui/components/file_uploader.py
```diff
@@ -37,7 +41,7 @@ async def did_finish_upload(self):
         if self.after_upload_callback is not None:
             await self.after_upload_callback(json.loads(self.uploaded_documents_json))
 
-    def perform_upload(self, event=None, after_upload_callback=None):
+    def perform_upload(self, event=None, after_upload_callback=None, corpus_name=""):
```
nenb, line 44: We now also need to pass a corpus name when uploading documents.

ragna/deploy/_ui/components/file_uploader.py
```diff
@@ -56,7 +60,7 @@ def perform_upload(self, event=None, after_upload_callback=None):
         self.custom_js = (
             final_callback_js
             + random_id
-            + f"""upload( self.get_upload_files(),  '{self.token}', '{self.informations_endpoint}', final_callback) """
+            + f"""upload( self.get_upload_files(), '{corpus_name}',  '{self.token}', '{self.informations_endpoint}', final_callback) """
```
nenb, line 63: We now also need to pass a corpus name when uploading documents.

ragna/deploy/_ui/modal_configuration.py
```diff
@@ -115,12 +134,37 @@ def __init__(self, api_wrapper, **params):
...
+        self.create_config(components)
+
+    async def did_click_on_start_chat_button(self, event):
+        if self.corpus_or_upload == USE_UPLOAD_LABEL:
+            if not self.document_uploader.can_proceed_to_upload():
+                self.change_upload_files_label("missing_file")
+            else:
+                self.start_chat_button.disabled = True
+                self.document_uploader.perform_upload(
+                    event, self.did_finish_upload, self.corpus_name_input.value
```
nenb, line 153: We now also need to pass a corpus name when uploading documents.

ragna/deploy/_ui/modal_configuration.py
```diff
@@ -115,12 +134,37 @@ def __init__(self, api_wrapper, **params):
...
+                )
+
+        elif self.corpus_or_upload == USE_CORPUS_LABEL:
+            if (
```
nenb, line 157: Don't send to API if invalid (and CSS should update to indicate to user that something is wrong).

ragna/deploy/_ui/resources/upload.js
```diff
@@ -1,8 +1,8 @@
-function upload(files, token, informationEndpoint, final_callback) {
-  uploadBatches(files, token, informationEndpoint).then(final_callback);
+function upload(files, corpus_name, token, informationEndpoint, final_callback) {
```
nenb, line 1: We now also need to pass a corpus name when uploading documents.

ragna/deploy/_ui/resources/upload.js
```diff
@@ -1,8 +1,8 @@
...
+  uploadBatches(files, corpus_name, token, informationEndpoint).then(final_callback);
 }
 
-async function uploadBatches(files, token, informationEndpoint) {
+async function uploadBatches(files, corpus_name, token, informationEndpoint) {
```
nenb, line 5: We now also need to pass a corpus name when uploading documents.

ragna/deploy/_ui/resources/upload.js
```diff
@@ -11,7 +11,7 @@ async function uploadBatches(files, token, informationEndpoint) {
   while (queue.length) {
     const batch = queue.splice(0, batchSize);
     await Promise.all(
-      batch.map((file) => uploadFile(file, token, informationEndpoint)),
+      batch.map((file) => uploadFile(file, corpus_name, token, informationEndpoint)),
```
nenb, line 14: We now also need to pass a corpus name when uploading documents.

ragna/deploy/_ui/resources/upload.js
```diff
@@ -20,7 +20,9 @@ async function uploadBatches(files, token, informationEndpoint) {
   return uploaded;
 }
 
-async function uploadFile(file, token, informationEndpoint) {
+async function uploadFile(file, corpus_name, token, informationEndpoint) {
+  // TODO: Need to decide how corpus_name is passed when uploading files
```
nenb, line 24: We need to decide on how this is done before merging this PR.

ragna/deploy/_ui/styles.py
```diff
@@ -97,11 +101,12 @@ def css(selector: Union[str, Iterable[str]], declarations: dict[str, str]) -> st
 
 MAIN_COLOR = "#DF5538"  # "rgba(223, 85, 56, 1)"
 
-# set modal height
-CONFIG_MODAL_MIN_HEIGHT = 610
-CONFIG_MODAL_MAX_HEIGHT = 850
+CONFIG_MODAL_HEIGHT = 705
```
nenb, line 104: Modal is no longer variable and has a fixed size.

ragna/deploy/_ui/css/modal_configuration/column.css
```diff
@@ -0,0 +1,3 @@
+:host(.metadata-filter-row-collection) {
+  overflow-y: scroll;
```
nenb, line 2: Allows scroll when user wants to have more than 5 rows of metadata filters.

ragna/deploy/_ui/css/modal_configuration/select.css
```diff
@@ -0,0 +1,16 @@
...
+}
+
+:host(.metadata-filter-operator) {
+  max-width: 12%;
+  min-width: 12%;
+  height: 30px;
+}
+
+:host(.metadata-filter-row-error) select.bk-input {
+  border: 2px red solid !important;
```
nenb, line 15: CSS that alerts a user when there is an issue in a row of metadata filters.

ragna/deploy/_ui/components/metadata_filters_builder.py
```diff
@@ -0,0 +1,249 @@
...
+class FilterRow(pn.viewable.Viewer):
+    key = param.Selector(objects=[""], default="")
+    operator = param.Selector(objects=[""] + ALLOWED_OPERATORS, default="")
+    value = param.Selector(objects=[""], default="")
+
+    def __init__(self, on_delete_callback, valid_key_value_pairs=None, **params):
+        super().__init__(**params)
+
+        key_select_disabled = True
+        if valid_key_value_pairs:
```
nenb, line 23: User is not allowed to type values, rather they must select from all available options. (Possibly a better UX.)

ragna/deploy/_ui/components/metadata_filters_builder.py
```diff
@@ -0,0 +1,248 @@
+import panel as pn
+import param
+
+from ragna.core._metadata_filter import MetadataFilter, MetadataOperator
+
+EXCLUDED_OPERATORS = ["RAW", "AND", "OR"]
```
nenb, line 6: Not supported in UI for now (`AND` is supported in the sense that it is used to combine all filters that user has provided).

ragna/deploy/_ui/components/metadata_filters_builder.py
```diff
@@ -0,0 +1,248 @@
...
+        self.delete_button = pn.widgets.ButtonIcon(
+            icon="trash",
+            width=25,
+            height=25,
+            css_classes=["metadata-filter-row", "metadata-filter-delete"],
+        )
+        self.delete_button.on_click(on_delete_callback)
+
+    @staticmethod
+    def set_css(widget, valid=True):
```
nenb, line 58: If error, alert the user via CSS (red outline around bad boxes).

ragna/deploy/_ui/components/metadata_filters_builder.py
```diff
@@ -0,0 +1,248 @@
...
+        if valid:
+            widget.css_classes = [
+                c for c in widget.css_classes if c != "metadata-filter-row-error"
+            ]
+        else:
+            widget.css_classes = widget.css_classes + ["metadata-filter-row-error"]
+
+    @param.depends("key", watch=True)
+    def key_did_change(self):
```
nenb, line 67: `Value` and `Operator` are only enabled, once a key has been selected. (So that we are able to provide a list of valid values to the user.)

ragna/deploy/_ui/components/metadata_filters_builder.py
```diff
@@ -0,0 +1,248 @@
...
+            self.key_select,
+            self.operator_select,
+            self.value_select,
+            self.delete_button,
+        )
+
+
+class MetadataFiltersBuilder(pn.viewable.Viewer):
+    filter_rows = param.List([])
+    corpus_names = param.List([])
```
nenb, line 117: Required parameters as every corpus name will have a potentially different set of key-values and so we need to refresh the values.

ragna/deploy/_ui/components/metadata_filters_builder.py
```diff
@@ -0,0 +1,248 @@
...
+        self.filter_rows = [
+            FilterRow(
+                key="",
+                operator="",
+                value="",
+                on_delete_callback=self.delete_filter_row,
+            )
+        ]
+
+        self.corpus_names_select.param.watch(self.activate_filter_rows, "value")
```
nenb, line 146: Whenever a user selects a corpus, then we activate the keys available for that corpus.

ragna/deploy/_ui/components/metadata_filters_builder.py
```diff
@@ -0,0 +1,248 @@
...
+
+    def activate_filter_rows(self, event):
+        if event.new != "":
```
nenb, line 149: In case user does not actually select a corpus, introduce this guard.

ragna/deploy/_ui/components/metadata_filters_builder.py
```diff
@@ -0,0 +1,248 @@
...
+                    ],
+                    key="",
+                    operator="",
+                    value="",
+                    on_delete_callback=self.delete_filter_row,
+                )
+            ]
+
+    def did_click_add_filter_row_button(self, event):
+        if not self.filter_rows[-1].is_empty():
```
nenb, line 163: Only add a new row if there is at least one non-empty element in the previous row.

ragna/deploy/_ui/components/metadata_filters_builder.py
```diff
@@ -0,0 +1,248 @@
...
+
+    def __panel__(self):
+        return pn.Column(
+            pn.pane.HTML("<b>Available Corpuses</b>"),
+            self.corpus_names_select,
+            pn.pane.HTML("<b>Metadata Filters</b>"),
+            pn.Column(
+                self.render_filters_rows,
+                pn.Row(self.add_filter_row_button),
+                max_height=160,
```
nenb, line 244: Restrict the height to 4 rows of metadata filters, and then scroll will be introduced along y-axis (see CSS).

ragna/deploy/_ui/api_wrapper.py
pmeier: This should be #500, right?

ragna/deploy/_ui/api_wrapper.py
```diff
@@ -81,13 +87,16 @@ def upload_endpoints(self):
         }
 
     async def start_and_prepare(
-        self, name, documents, source_storage, assistant, params
+        self, name, documents, corpus_name, source_storage, assistant, params
     ):
+        if corpus_name is None:
```
nenb, line 92: Not sure what name should be passed here...

ragna/deploy/_ui/components/metadata_filters_builder.py
```diff
@@ -0,0 +1,275 @@
...
+from ragna.core._metadata_filter import MetadataFilter, MetadataOperator
+
+EXCLUDED_OPERATORS = ["RAW", "AND", "OR"]
+ALLOWED_OPERATORS = [
+    op
+    for op in list(MetadataOperator.__members__.keys())
+    if op not in EXCLUDED_OPERATORS
+]
+
+DUNDER_METHODS = {f"__{op}__".lower(): op for op in ALLOWED_OPERATORS}
```
nenb, line 13: Can remove this validation if not helpful. It uses the type information passed by the source storage to infer a list of valid operators.

It's not super useful though as string will still allow 'GT' etc., which I don't think is helpful (e.g. `path > /my_path` is valid Python code, but I don't think it's what users will want).

See below (`compute_valid_operator_options` etc) for more details.

ragna/deploy/_ui/components/metadata_filters_builder.py
```diff
@@ -0,0 +1,275 @@
...
+            filter_row.construct_metadata_filter()
+            for filter_row in self.filter_rows
+            if filter_row.construct_metadata_filter() is not None
+        ]
+
+        if not metadata_filters:
+            return None
+        return MetadataFilter.and_(metadata_filters).to_primitive()
+
+    def validate(self):
```
nenb, line 249: Called in `modal_configuration.py` ...

ragna/deploy/_ui/components/metadata_filters_builder.py
```diff
@@ -0,0 +1,275 @@
...
+        self.filter_rows = new_filter_rows
+
+    @pn.depends("filter_rows")
+    def render_filters_rows(self):
+        return pn.Column(
+            *self.filter_rows,
+            css_classes=["metadata-filter-row-collection"],
+        )
+
+    def construct_metadata_filters(self):
```
nenb, line 238: Called in `modal_configuration.py`...

ragna/deploy/_ui/modal_configuration.py
```diff
@@ -115,21 +137,50 @@ def __init__(self, api_wrapper, **params):
...
             self.start_chat_button.disabled = True
-            self.document_uploader.perform_upload(event, self.did_finish_upload)
 
-    async def did_finish_upload(self, uploaded_documents):
+            await self.did_finish_upload(
+                self.metadata_filter_rows.construct_metadata_filters(),
+                corpus_name=self.metadata_filter_rows.corpus_names_select.value,
+            )
+
+    async def did_finish_upload(self, uploaded_documents, corpus_name=None):
```
nenb, line 171: We now need to pass `corpus_name` as it is attached to the chat.

ragna/deploy/_ui/modal_configuration.py
```diff
@@ -146,22 +197,20 @@ async def did_finish_upload(self, uploaded_documents):
             self.start_chat_button.disabled = False
 
     def change_upload_files_label(self, mode="normal"):
+        self.error = False
```
nenb, line 200: Required to introduce `error` so that we can watch it, and update the CSS whenever a user performs an invalid operation.

This was previously implicitly watched by other components, but now that we have added many more components to the modal, we need to introduce an explicit component to watch it (otherwise there are edge cases that arise where the CSS is not updated when a user performs an error).

ragna/deploy/_ui/modal_configuration.py
pmeier: Could we maybe move the definitions of these cards in two dedicated methods so we don't have an if else block that spans 70 LoC? Something like

```
card = self._make_corpus_card() if is_corpus else self._make_documents_card()
```

pmeier (commented):
Few comments, but mostly LGTM. Thanks Nick!

ragna/deploy/_ui/api_wrapper.py
nenb: When a user uploads files, they have the option to pass a Corpus Name. If they do not pass a Corpus Name, then we hit this.

ragna/deploy/_ui/components/metadata_filters_builder.py
```diff
@@ -0,0 +1,275 @@
...
+        return operators
+
+    def is_empty(self):
+        return self.key == "" and self.operator == "" and self.value == ""
+
+    def validate(self):
+        if self.key == "":
+            self.set_css(self.key_select, valid=False)
+        else:
+            self.set_css(self.key_select)
```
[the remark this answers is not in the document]
nenb: I have actually removed this completely now, as it was only handling the case where certain dropdowns were empty. This is no longer possible we no longer use `""` as default values.

ragna/deploy/_ui/components/metadata_filters_builder.py
```diff
@@ -0,0 +1,275 @@
...
+        else:
+            self.set_css(self.value_select)
+
+        # if any attribute is "" then return False
+        return not (self.key == "" or self.operator == "" or self.value == "")
+
+    def construct_metadata_filter(self):
+        # probably not required, but just in case
+        if self.key == "" or self.operator == "" or self.value == "":
+            return None
```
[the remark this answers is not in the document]
nenb: Removed, as this case is no longer possible (have removed the use of `""` as default values).

ragna/deploy/_ui/components/metadata_filters_builder.py
nenb: But does `<5` not achieve the same result?

ragna/deploy/_ui/api_wrapper.py
pmeier: Shouldn't we put the default corpus name as a default value in that text field instead of quietly resolving it in the background? Meaning, instead of allowing users to not fill the text field, just have `"default"` there as value. Functionally this should be the same, but the user knows explicitly what corpus the documents were uploaded to.

ragna/deploy/_ui/components/metadata_filters_builder.py
```diff
@@ -0,0 +1,220 @@
...
+        )
+
+        self.value_select = pn.widgets.Select.from_param(
+            self.param.value,
+            name="",
+            css_classes=["metadata-filter-value"],
+            disabled=True,
+        )
+
+        self.multi_value_select = pn.widgets.MultiChoice.from_param(
```
pmeier, line 54: 1. I like the idea and the widget in general, but right now I'm only able to see a single row 

   image

   Plus, it also leads to another vertical scroll bar so we now have two scroll bars on a modal, which is a big no-no according to Smera.
2. Not sure where this is handled, but the value is extracted wrong from the widget. For example if I select the following filter in the modal

   image

   I get the following output in the chat history

   image

   There are two issues here:

   1. It did pick a different document as I did select. In my test I couldn't make out a pattern other than it only ever picked a single one.
   2. The value is still a string rather than a list of strings. This leads to errors in the backed, e.g. Chroma, as reported in a previous round of review.

pmeier (approved):
Thanks Nick for all the hard work here! I found a few bugs, but I don't want to keep this PR open any longer. Let me send a small cleanup commit to fix any nits I might have and fix CI. I'll open issues for all of my comments so we can fix them in follow-up PRs.

PR #500, opened 2024-08-26
add list_metadata functionality to Python API

pmeier:
Addresses the backend part of #493.

### DIFF
--- a/ragna/source_storages/_chroma.py
+++ b/ragna/source_storages/_chroma.py
@@ -1,11 +1,18 @@
 from __future__ import annotations
 
 import uuid
+from collections import defaultdict
 from typing import TYPE_CHECKING, Any, Optional, cast
 
 import ragna
-from ragna.core import Document, MetadataFilter, MetadataOperator, Source
+from ragna.core import (
+    Document,
+    MetadataFilter,
+    MetadataOperator,
+    Source,
+)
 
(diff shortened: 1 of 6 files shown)


ragna/source_storages/_chroma.py
```diff
@@ -36,14 +43,60 @@ def __init__(self) -> None:
...
-        )
-
     def list_corpuses(self) -> list[str]:
         return [collection.name for collection in self._client.list_collections()]
 
+    def _get_collection(
+        self, corpus_name: str, *, create: bool = False
+    ) -> chromadb.Collection:
+        if create:
+            return self._client.get_or_create_collection(
```
nenb, line 53:
```suggestion
            return self._client.create_collection(
```

ragna/source_storages/_chroma.py
```diff
@@ -36,14 +43,60 @@ def __init__(self) -> None:
...
+                corpus_name, embedding_function=self._embedding_function
+            )
+
+        collections = list(self._client.list_collections())
```
nenb, line 57: Nitpick: Why is it important to distinguish between `no_corpuses_available` and `non_existing_corpus`?

If we didn't need to distinguish between them, then perhaps we could just use `self._client.get_collection`, which throws a `ValueError` if the collection (corpus) does not exist.

nenb (approved):

PR #502, opened 2024-08-29
drop Python 3.9 and add support for 3.12

pmeier:
`panel==1.5.*`, i.e. the next minor version, drops support for Python 3.9. As we'll need to upgrade at some point, we also do the switch now. In addition, this also adds support for Python 3.12 so we still have three compatible versions.


pyproject.toml
```diff
@@ -19,13 +19,12 @@ classifiers = [
...
     "Programming Language :: Python :: 3.11",
 ]
-requires-python = ">=3.9"
+requires-python = ">=3.10"
 dependencies = [
     "aiofiles",
     "emoji",
     "fastapi",
     "httpx",
-    "importlib_metadata>=4.6; python_version<'3.10'",
```
pmeier, line 28: No longer needed, since we at least have 3.10 and thus don't need a backport of `importlib.metadata`.

ragna/_compat.py
pmeier, line 1: This whole file is no longer needed, since all compatibility functions we had in there are natively supported by Python 3.10 and upwards.

ragna/deploy/_api/orm.py
```diff
@@ -30,6 +31,34 @@ def process_result_value(
         return json.loads(value)
 
 
+class UtcDateTime(types.TypeDecorator):
+    """UTC timezone aware datetime type.
+
+    This is needed because sqlalchemy.types.DateTime(timezone=True) does not
+    consistently store the timezone.
+    """
```
pmeier, line 39: Per class docstring.

ragna/deploy/_api/schemas.py
```diff
@@ -75,7 +85,7 @@ class Message(BaseModel):
     content: str
     role: ragna.core.MessageRole
     sources: list[Source] = Field(default_factory=list)
-    timestamp: datetime.datetime = Field(default_factory=datetime.datetime.utcnow)
```
pmeier, line 78: `datetime.datetime.utcnow` is deprecated starting with Python 3.12 and thus I refactored this whole thing to be timezone aware.

blakerosenthal (approved):
Awesome, works great for me on 3.12

PR #503, opened 2024-08-30
use bokeh_fastapi through panel

pmeier:
holoviz/panel#7205 adds support for `bokeh_fastapi` inside `panel`. This PR is here to test if everything works. cc @philippjfr


.github/actions/setup-env/action.yml
```diff
@@ -62,7 +62,7 @@ runs:
       run: |
         pip install \
           git+https://github.com/bokeh/bokeh-fastapi.git@main \
-          git+https://github.com/holoviz/panel@7377c9e99bef0d32cbc65e94e908e365211f4421
+          git+https://github.com/holoviz/panel@fastapi_serve
```
pmeier, line 65: This implicitly updates to `panel==1.5.*`. Thus, we'll have to check styling as well.

ragna/core/_components.py
pmeier, line 1: Changes in this file are drivebys of changes missed in d869bd0687a54a33d63c45689eebc0b75491a809.

ragna/deploy/_core.py
```diff
@@ -79,8 +82,14 @@ def server_available() -> bool:
...
-        panel_app = make_ui_app(engine)
-        panel_app.serve_with_fastapi(app, endpoint="/ui")
+        ui_app = make_ui_app(engine)
+        panel.io.fastapi.add_applications({"/ui": ui_app.index_page}, server=app)
+        for dir in ["css", "imgs"]:
+            app.mount(
+                f"/{dir}",
+                StaticFiles(directory=str(Path(__file__).parent / "_ui" / dir)),
+                name=dir,
+            )
```
pmeier, line 92: This should probably go into `BokehFastapi` under a `static_dirs` keyword similar to what `pn.serve` does.

pmeier:
Re https://github.com/Quansight/ragna/pull/503#discussion_r1738385891. The only difference I found is that the chat interface now seems to fill the whole vertical space of its container whereas before it only occupied the middle area.

### `panel==1.5.*` / PR

image

### `panel==1.4.4` / `main`

image

pmeier:
I just realized that the document pill in the top-right corner is also missing for `panel==1.5.*`.

pmeier:
Document upload is also broken:

image

pmeier:
@pierrotsmnrd We can use the [`FileDropper`](https://holoviz-dev.github.io/panel/reference/widgets/FileDropper.html) widget going forward. But if the old one can easily be fixed, we can also do this in a follow-up PR.

kcpevey:
For reference, the current `document_upload` widget is a FileInput. It used to be a custom widget, as is shown here - https://github.com/Quansight/ragna/blob/main/ragna/deploy/_ui/modal_configuration.py#L90-L94

pmeier:
@kcpevey re https://github.com/Quansight/ragna/pull/503?new_mergebox=true#issuecomment-2501915952

I broke that check in #441. And at that time I was well aware of it :melting_face: 

> Note that the change to the documents currently leaves the UI defunct.

Anyway, I pushed the necessary changes in 2d597943d46fd657da0e9c671e6894974473e22d to get the document pills back.

pmeier:
@kcpevey re https://github.com/Quansight/ragna/pull/503#issuecomment-2501933845

We had a custom widget because we previously used a complex system to upload documents. In the first step we registered the document with the API and got back a location where to upload the document to. This was usually our server, but could also be an S3 bucket or the like.

While very flexible, that also brought a lot of complex handling. We removed that in favor of always uploading to local in #441. I've hot-patched the UI code in #443 to make it work again. The code is messy right now, but that is my problem not yours.

> Should I just be adding in the new FileDropper widget from Panel

I've asked myself the same in https://github.com/Quansight/ragna/pull/503#issuecomment-2348748874. I guess this is a good move going forward, but I don't want to keep this PR up forever.

> and styling it to look similar to the custom one?

My vote is out for trying to get the `FileInput` to look like what we had before. Hopefully that is no too complicated. If it turns out to not be simple or a lot of work that we might throw away when going for the `FileDropper`, we can reconsider.

kcpevey:
Filedropper UI changes have been added, the widget now looks like this:
image

PR #509, opened 2024-12-11
deploy refactor

pmeier:


pmeier:
As of ac13a3cb316fd75090167c528562a1be6e25970d this PR is mostly feature complete. All individual features have been reviewed in the respective PRs. Thus, the only thing left here is cleanup

- [ ] fix documentation
- [ ] checkout testing segfault on 3.11 and 3.12
- [ ] checkout CI failure on Windows
- [ ] re-enable and fix UI tests
- [ ] fix styling on file input widget

Not everything from the list above has to be implemented as part of this PR and can be done in follow-ups.

pmeier:
[Build on RTD](https://readthedocs.org/projects/ragna/builds/26568496/) fails with

```
WARNING: Built wheel for Ragna is invalid: Wheel has unexpected file name: expected '0.3.0.dev58+g0d1e09b.d20241212103454', got '0.3.0.dev58+g0d1e09b.d20241212103509'
```

If I see this right it complains about dirty timestamp mismatch. I'm guessing this is fallout from pypa/setuptools-scm#962.

pmeier:
I think the Windows CI failure happens because the 

https://github.com/Quansight/ragna/blob/a6050caeb51ffefcedbc98f21d34480a6790230d/.github/actions/setup-env/action.yml#L53

call seems to not work properly and even corrupt the env somehow. We only get the following output from the command

```
Microsoft Windows [Version 10.0.20348.2849]
(c) Microsoft Corporation. All rights reserved.
D:\a\ragna\ragna>
```

and after that although the environment should be activated, the system Python is used. And that finally leads to the failures.

PR #510, opened 2024-12-11
Fix #504 : honor the indentation in the MetadataFilter formatting 

### LINKED ISSUE #504: Fix MetadataFilter formatting in chat info side bar
_Reported in https://github.com/Quansight/ragna/pull/499#discussion_r1745589754._

---

The snippet below does not honor the indentation

https://github.com/Quansight/ragna/blob/9221bb8d67b5002abd42a6fc985a017b7aa0072f/ragna/deploy/_ui/central_view.py#L208-L210

image

```python
form ragna.core import MetadataFilter

print(MetadataFilter.and_([MetadataFilter.in_("document_name", ["foo.md"])]))
```

```
AND(
  IN('document_name', ['foo.md']),
)
```

pierrotsmnrd:
Fix #504 : honor the indentation in the MetadataFilter formatting

### DIFF
--- a/ragna/deploy/_ui/central_view.py
+++ b/ragna/deploy/_ui/central_view.py
@@ -205,10 +205,16 @@ class CentralView(pn.viewable.Viewer):
             # `MetadataFilter`s provided as dict
             title = "Metadata Filters"
 
-            metadata_filters_readable = str(
-                MetadataFilter.from_primitive(self.current_chat["metadata"]["input"])
-            ).replace("\n", "<br>").replace(" ", "&nbsp;")
-            
+            metadata_filters_readable = (
+                str(
+                    MetadataFilter.from_primitive(
+                        self.current_chat["metadata"]["input"]
+                    )
+                )
+                .replace("\n", "<br>")
+                .replace(" ", "&nbsp;")
+            )
+
             details = f"<div class='details details_block' style='display:block;'><pre>{metadata_filters_readable}</pre></div><br />\n\n"
             grid_height = 1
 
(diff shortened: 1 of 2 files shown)


ragna/deploy/_ui/central_view.py
```diff
@@ -205,11 +205,17 @@ def on_click_chat_info_wrapper(self, event):
...
-            metadata_filters_readable = str(
-                MetadataFilter.from_primitive(self.current_chat["metadata"]["input"])
-            ).replace("\n", "<br>")
+            metadata_filters_readable = (
+                str(
+                    MetadataFilter.from_primitive(
+                        self.current_chat["metadata"]["input"]
+                    )
+                )
+                .replace("\n", "<br>")
```
pmeier, line 214: Shouldn't we able to remove this given that `<pre>` should honor line breaks?

ragna/deploy/_ui/central_view.py
```diff
@@ -205,11 +205,17 @@ def on_click_chat_info_wrapper(self, event):
...
+                .replace(" ", "&nbsp;")
```
pmeier, line 215: Could you explain that?

pierrotsmnrd: ```.replace("\n", "<br>")```
We need to put explicit `<br>` to have line-breaks at the right places, as explained in the other comment.

```.replace(" ", "&nbsp;")```
We need to replace the whitespaces as explicit non-breakable spaces to show indentation, otherwise it's ignored.

Try removing these lines to see how it behaves.

pmeier: So do we even need the `<pre>` tag in that case?

pierrotsmnrd: Yes, to prevent from any forced line-break, and it also renders with a nice monospaced font.

Here are two screenshots comparing both results
- with `<pre>`
Capture d’écran 2024-12-11 à 12 10 20

- without `<pre>`
Capture d’écran 2024-12-11 à 12 11 10


If you want the non-monospaced font, so no `<pre>`, I can probably fix the result with CSS.

pmeier (approved):
Thanks Pierre!

PR #511, opened 2024-12-11
Fix #505 : Remove "empty filter" row

### LINKED ISSUE #505: Remove "Empty filter" type in favor of no filter row
_Reported in https://github.com/Quansight/ragna/pull/499#discussion_r1745599287_

---

Instead of having an "Empty Filter" type that indicates that we don't want to apply any filter, we should just have no filter row at all.

https://github.com/Quansight/ragna/blob/9221bb8d67b5002abd42a6fc985a017b7aa0072f/ragna/deploy/_ui/components/metadata_filters_builder.py#L95-L96

image

See https://github.com/Quansight/ragna/pull/499#discussion_r1745599287 for a non-exhaustive list of problems that we'll have to solve otherwise.

Ultimately, we need to adapt the code such that we don't need the `if filter_row.key != NO_FILTER_KEY` in

https://github.com/Quansight/ragna/blob/9221bb8d67b5002abd42a6fc985a017b7aa0072f/ragna/deploy/_ui/components/metadata_filters_builder.py#L198-L207

pierrotsmnrd:
Fix #505 : Remove "empty filter" row

I have kept the line `if filter_row.key != NO_FILTER_KEY` as it is necessary to create a new chat after adding a first filter row, which appears empty by default. Run it and tell me what needs to be improved.


pmeier:
This doesn't seem to fix it:

image

image

To reiterate: there shouldn't be any line in the "Metadata Filters" section by default and one should only appear if the user clicks the "+" button. Functionalitywise, in case there is no line it should just work as right now if it displays "Empty Filter".

pmeier (changes requested):
See above

pierrotsmnrd:
I have replaced the string "Empty Filter" by an empty string, so the fact you see it makes me think you might have used the wrong branch. Is that possible ?

pmeier:
> Is that possible ?

It seems it is. No idea how that happened given that I even pushed a merge commit 6ba57abc71cc4904cef7a0742a32dc181bb9ac6d :face_with_spiral_eyes:

pmeier (commented):
- For source storages that do not have any corpus available, can we disablethe "+" button or potentially even remove it or the "Metadata Filters" section completely?
  
  image

- When selecting a source storage that has multiple corpora, selecting a corpus from the dropdown menu creates and empty filter that does not go away automatically

  [Screencast from 2024-12-12 16-55-09.webm](https://github.com/user-attachments/assets/fab6cd7c-2ad1-4f47-a384-622af8195903)

  This seems just like an UI issue, because the chat can be started with the row present and is correctly using no metadata filter.

pmeier (approved):
Thanks Pierre!

PR #512, opened 2024-12-11
Fix #506 : Fix value selector widget for IN and NOT_IN operator

### LINKED ISSUE #506: Fix value selector widget for IN and NOT_IN operator
_Reported in https://github.com/Quansight/ragna/pull/499#discussion_r1745608702_

---

1. The vertical spacing should allow for at least three values to be visible at all times.
   
   image
2. The value extracted from the widget is wrong. For example if I select the following filter in the modal

   image

   I get the following output in the chat history

   image

   There are two issues here:

   1. It did pick a different document as I did select. In my test I couldn't make out a pattern other than it only ever picked a single one.
   2. The value is still a string rather than a list of strings. This leads to errors in the backend, e.g. Chroma.

pierrotsmnrd:
Fix #506 : Fix value selector widget for IN and NOT_IN operator

This PR only fixes the UI issue with the vertical spacing. I was not able to reproduce the second problem. Let me know if it still occurs.


pmeier (changes requested):
This is what I see on Firefox and Chromium respectively

image

image

which looks like it was before.

I also still see the second issue reported in #506. Can you share the process of what you did that wasn't able to reproduce it?

pierrotsmnrd:
I've pushed a fix. If you still see the issue, make sure you reload the page without cache.

pmeier (approved):
Thanks Pierre!

Issue #523, opened 2024-12-20
investigate chroma changed filtering behavior


pmeier:
In be943a75fd8027e77fb14d75ebe2a8a830d4cba7 as part of #461 I changed the behavior of LanceDB to behave more like Chroma when filtering results. See https://github.com/Quansight/ragna/pull/461#issuecomment-2275737866 for details. However, I think upstream Chroma might have aligned with LanceDB in the meantime. If it weren't for the upper pin

https://github.com/Quansight/ragna/blob/71acf83d4114514c56e298f7acc431551933e522/ragna/source_storages/_vector_database.py#L49

we see some test failures. I couldn't find a changelog or issue indicating the change in behavior.

We should align Chroma and LanceDB on the latest version again. That might mean that we need to revert the patch mentioned above and fix the tests.

smokestacklightnin:
As an experiment, I tried updating the chromadb version. Most, if not all, of the test failures come from the same cause. The test failure mentions 

```
NotImplementedError: In Chroma v0.6.0, list_collections only returns collection names. Use Client.get_collection(default) to access query. See https://docs.trychroma.com/deployment/migration for more information.
```

The [migration guide (v0.6.0)](https://docs.trychroma.com/production/administration/migration#migration-log) discusses how to correctly use `list_collections` and get collection names in the current version. I tried this on one of the test failures and it started passing. I assume this solution will work for all failures of this type.

All that said, we can and probably should do some cleanup on the filtering code

pmeier:
Indeed `chromadb>=0.6` brings a whole lot of new issues and we should follow their upgrade guide. However, these are _not_ the issues that are relevant here. If you `pip install --upgrade 'chromadb>=0.5.12,<0.6'` you get

```shell
$ pytest -k chroma
[...]
======================================== FAILURES =========================================
______________________________ test_smoke[Chroma-or-nested] _______________________________
tests/source_storages/test_source_storages.py:137: in test_smoke
    assert actual_idcs == expected_idcs
E   assert [0, 1, 2] == [0, 1]
E     
E     Left contains one more item: 2
E     Use -v to get more diff
__________________________________ test_smoke[Chroma-ne] __________________________________
tests/source_storages/test_source_storages.py:137: in test_smoke
    assert actual_idcs == expected_idcs
E   assert [2, 3, 4, 5, 6] == [2, 5, 6]
E     
E     At index 1 diff: 3 != 5
E     Left contains 2 more items, first extra item: 5
[...]

PR #524, opened 2024-12-24
feat: Qdrant RAG Storage

Anush008:
## Description

This PR adds support for Qdrant - https://qdrant.tech/ to be used as RAG storage.

### DIFF
--- a/ragna-docker.toml
+++ b/ragna-docker.toml
@@ -4,7 +4,7 @@ document = "ragna.core.LocalDocument"
 source_storages = [
     "ragna.source_storages.Chroma",
     "ragna.source_storages.LanceDB",
-    "ragna.source_storages.QdrantDB"
+    "ragna.source_storages.QdrantDB",
 ]
 assistants = [
     "ragna.assistants.ClaudeHaiku",
--- a/ragna-docker.toml
+++ b/ragna-docker.toml
@@ -3,7 +3,8 @@ authentication = "ragna.deploy.RagnaDemoAuthentication"
 document = "ragna.core.LocalDocument"
 source_storages = [
     "ragna.source_storages.Chroma",
-    "ragna.source_storages.LanceDB"
+    "ragna.source_storages.LanceDB",
+    "ragna.source_storages.QdrantDB"
 ]
 assistants = [
     "ragna.assistants.ClaudeHaiku",
--- a/ragna/source_storages/_qdrant.py
+++ b/ragna/source_storages/_qdrant.py
@@ -22,8 +22,15 @@ if TYPE_CHECKING:
     from qdrant_client import models
 
 
-class QdrantDB(VectorDatabaseSourceStorage):
+class Qdrant(VectorDatabaseSourceStorage):
     """[Qdrant vector database](https://qdrant.tech/)
+
+    Use the `QDRANT_URL` and `QDRANT_API_KEY` env variables to configure connection to a qdrant server.
+
+    Eg:
+    > export QDRANT_URL="https://xyz-example.eu-central.aws.cloud.qdrant.io:6333"
+
+    > export QDRANT_API_KEY="<your-api-key-here>"
 
     !!! info "Required packages"
 
--- a/ragna/source_storages/_qdrant.py
+++ b/ragna/source_storages/_qdrant.py
@@ -46,7 +46,7 @@ class QdrantDB(VectorDatabaseSourceStorage):
 
         url = os.getenv("QDRANT_URL")
         api_key = os.getenv("QDRANT_API_KEY")
-        path = path = ragna.local_root() / "qdrant"
+        path = ragna.local_root() / "qdrant"
 
         # Cannot pass both url and path
         self._client = (
--- a/ragna/source_storages/_qdrant.py
+++ b/ragna/source_storages/_qdrant.py
@@ -1,0 +1,244 @@
+from __future__ import annotations
+
+import uuid
+from collections import defaultdict
+from typing import TYPE_CHECKING, Any, Optional, cast
+
+import ragna
+from ragna.core import (
+    Document,
+    MetadataFilter,
+    MetadataOperator,
+    PackageRequirement,
+    Requirement,
+    Source,
+)
+
+from ._utils import raise_no_corpuses_available, raise_non_existing_corpus
+from ._vector_database import VectorDatabaseSourceStorage
+
+if TYPE_CHECKING:
+    from qdrant_client import models
+
+
+class QdrantDB(VectorDatabaseSourceStorage):
+    """[Qdrant vector database](https://qdrant.tech/)
+
+    !!! info "Required packages"
+
+        - `qdrant-client>=1.12.0`
+    """
+
+    DOC_CONTENT_KEY = "__document"
+
+    @classmethod
+    def requirements(cls) -> list[Requirement]:
+        return [
+            *super().requirements(),
+            PackageRequirement("qdrant-client>=1.12.1"),
+        ]
+
+    def __init__(self) -> None:
+        super().__init__()
+
+        from qdrant_client import QdrantClient
+
+        self._client = QdrantClient(path=ragna.local_root() / "qdrant")
+
+    def list_corpuses(self) -> list[str]:
+        return [c.name for c in self._client.get_collections().collections]
+
+    def _ensure_table(self, corpus_name: str, *, create: bool = False):
+        table_names = self.list_corpuses()
+        no_corpuses = not table_names
+        non_existing_corpus = corpus_name not in table_names
+
+        if non_existing_corpus and create:
+            from qdrant_client import models
+
+            self._client.create_collection(
+                collection_name=corpus_name,
+                vectors_config=models.VectorParams(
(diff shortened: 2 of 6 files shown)


ragna/source_storages/_qdrant.py
```diff
@@ -0,0 +1,244 @@
...
+            *super().requirements(),
+            PackageRequirement("qdrant-client>=1.12.1"),
+        ]
+
+    def __init__(self) -> None:
+        super().__init__()
+
+        from qdrant_client import QdrantClient
+
+        self._client = QdrantClient(path=ragna.local_root() / "qdrant")
```
Anush008, line 46: Will use ENVs to support Qdrant Docker and Cloud.

ragna/source_storages/_qdrant.py
```diff
@@ -0,0 +1,249 @@
...
+            return models.FieldCondition(key=key, range=models.Range(gte=value))
+        elif operator == MetadataOperator.IN:
+            return models.FieldCondition(key=key, match=models.MatchAny(any=value))
+        elif operator in {MetadataOperator.NE, MetadataOperator.NOT_IN}:
+            except_value = [value] if operator == MetadataOperator.NE else value
+            return models.FieldCondition(
+                key=key, match=models.MatchExcept(**{"except": except_value})
+            )
+        else:
+            raise ValueError(f"Unsupported operator: {operator}")
```
pmeier, line 172: This only applies to the `RAW` operator, correct? Looking at the documentation, there are plenty of other operators. Is it possible to take a JSON filter even when running in Python? If so this is probably very similar to what Chroma is doing:

https://github.com/Quansight/ragna/blob/71acf83d4114514c56e298f7acc431551933e522/ragna/source_storages/_chroma.py#L153-L154

ragna/source_storages/_qdrant.py
```diff
@@ -0,0 +1,249 @@
...
+                        },
+                    )
+                )
+
+        self._client.upsert(collection_name=corpus_name, points=points)
+
+    def _build_condition(self, operator, key, value):
+        from qdrant_client import models
+
+        if operator == MetadataOperator.EQ:
```
pmeier, line 154: Please add a link to the documentation as comment here
```suggestion
        # See https://qdrant.tech/documentation/concepts/filtering/#range
        if operator == MetadataOperator.EQ:
```

ragna/source_storages/_qdrant.py
```diff
@@ -0,0 +1,249 @@
...
+        else:
+            raise ValueError(f"Unsupported operator: {operator}")
+
+    def _translate_metadata_filter(
+        self, metadata_filter: MetadataFilter
+    ) -> models.Filter:
+        from qdrant_client import models
+
+        if metadata_filter.operator is MetadataOperator.RAW:
+            return cast(models.Filter, metadata_filter.value)
```
pmeier, line 180: Oh well. Not deleting my comment above. This begs the question what the error above is supposed to do. Just future proofing?

ragna/source_storages/_qdrant.py
```diff
@@ -0,0 +1,249 @@
...
+        from qdrant_client import QdrantClient
+
+        url = os.getenv("QDRANT_URL")
+        api_key = os.getenv("QDRANT_API_KEY")
+        path = path = ragna.local_root() / "qdrant"
+
+        # Cannot pass both url and path
+        self._client = (
+            QdrantClient(url=url, api_key=api_key) if url else QdrantClient(path=path)
+        )
```
pmeier, line 54: IIUC, you only need the API key if you are using a remote database, i.e. one that needs an URL to connect to it. Will the `QdrantClient` raise an informative error message when the API key is not set?

Regardless, I'd like to refactor this slightly to make it more obvious what is happening

```suggestion
        
        if (url := os.getenv("QDRANT_URL")) is not None:
            kwargs = dict(
                url=url,
                api_key = os.getenv("QDRANT_API_KEY"),
            )
        else:
            kwargs = dict(ragna.local_root() / "qdrant")

        self._client = QdrantClient(**kwargs)
```

pmeier (commented):
Hi @Anush008 :wave: and thanks a ton for the contribution. The patch looks really good already! I have a few small comments, but otherwise this is almost good to go!

tests/source_storages/test_source_storages.py
```diff
@@ -10,9 +10,9 @@
     PlainTextDocumentHandler,
     RagnaException,
 )
-from ragna.source_storages import Chroma, LanceDB, RagnaDemoSourceStorage
+from ragna.source_storages import Chroma, LanceDB, QdrantDB, RagnaDemoSourceStorage
 
-SOURCE_STORAGES = [Chroma, LanceDB, RagnaDemoSourceStorage]
+SOURCE_STORAGES = [Chroma, LanceDB, QdrantDB, RagnaDemoSourceStorage]
```
pmeier, line 15: The fact that we can plug this in here and all tests "just work" is a little awkward since we are looking into failures for more recent Chroma versions in #523. @Anush008 could you join the discussion there?

Anush008:
Hi @pmeier. Thanks for reviewing.

ragna/source_storages/_qdrant.py
Anush008: RAW operator is handled below.
This else clause was really "just to be sure".

ragna/source_storages/_qdrant.py
Anush008: Not only cloud, but a local server with Docker.
```
docker run -p 6333:6333 qdrant/qdrant
```
URL would be http://localhost:6333.

Note that, when you pass the local file `path` argument instead of the `url`, the client does a Python based imitation of the Qdrant server APIs. This should only be used for prototyping and CI.

For real use cases, Qdrant(Rust based) server should be used.

Anush008: The current snippet should be fine IMO.

ragna/source_storages/_qdrant.py
Anush008: Yes. https://github.com/Quansight/ragna/pull/524#discussion_r1901276262

ragna/source_storages/_qdrant.py
pmeier: In that case only minor cleanup :slightly_smiling_face: 

```suggestion
        url = os.getenv("QDRANT_URL")
        api_key = os.getenv("QDRANT_API_KEY")
        path = ragna.local_root() / "qdrant"

        # Cannot pass both url and path
        self._client = (
            QdrantClient(url=url, api_key=api_key) if url else QdrantClient(path=path)
        )
```

ragna-docker.toml
```diff
@@ -3,7 +3,8 @@ authentication = "ragna.deploy.RagnaDemoAuthentication"
 document = "ragna.core.LocalDocument"
 source_storages = [
     "ragna.source_storages.Chroma",
-    "ragna.source_storages.LanceDB"
+    "ragna.source_storages.LanceDB",
+    "ragna.source_storages.QdrantDB"
```
pmeier, line 7: Nit: add a trailing comma

```suggestion
    "ragna.source_storages.QdrantDB",
```

ragna/source_storages/_qdrant.py
```diff
@@ -0,0 +1,250 @@
...
+)
+
+from ._utils import raise_no_corpuses_available, raise_non_existing_corpus
+from ._vector_database import VectorDatabaseSourceStorage
+
+if TYPE_CHECKING:
+    from qdrant_client import models
+
+
+class QdrantDB(VectorDatabaseSourceStorage):
```
pmeier, line 25: Do you want the `DB` postfix? We only did that for `LanceDB`, because that is how they have it themselves. From the Python package and docker image, I guess we should drop the postfix?

```suggestion
class Qdrant(VectorDatabaseSourceStorage):
```

ragna/source_storages/_qdrant.py
Anush008: Great save! Thanks.

ragna/source_storages/_qdrant.py
```diff
@@ -0,0 +1,250 @@
...
+    """[Qdrant vector database](https://qdrant.tech/)
+
+    !!! info "Required packages"
```
pmeier, line 28: Could you add something here that indicates that the user might need to set the `QDRANT_URL` and `QDRANT_API_KEY` env vars?

Anush008: Resolved with https://github.com/Quansight/ragna/pull/524/commits/c9d19e7b967f2151c0b66d93682aef51b4161239.

pmeier:
Thanks Anush, LGTM! Any remaining nits that I might have, I'll fix myself. However, we need to wait until https://github.com/Quansight/ragna/pull/524#discussion_r1899369374 is resolved. @smokestacklightnin is actively working on that. I'll merge afterwards if CI stays green.

pmeier (approved):
@Anush008 Turns out Chroma is the outlier and Qdrant is actually aligned with the filtering behavior of LanceDB and what Chroma used to do. Thus, there is no need to do anything else. Let's merge!

Anush008:
🥳

PR #540, opened 2025-01-16
Fix non-editable install

### LINKED ISSUE #529: fix non-editable install
RTD CI fails with

```
Building wheel for Ragna (pyproject.toml): started
  Building wheel for Ragna (pyproject.toml): finished with status 'done'
  Created wheel for Ragna: filename=Ragna-0.4.0.dev4+g355b28e.d20241231122325-py3-none-any.whl size=116105 sha256=bee64fc90565bc6c4d1ab31889c24b7a0b0496ef7b4beccba94c660230490e4f
  Stored in directory: /tmp/pip-ephem-wheel-cache-7v_8zvwt/wheels/f7/74/2b/bbd14c519f9e6d6d034f2d7e7d206003123768212ba61b1a3b
  WARNING: Built wheel for Ragna is invalid: Wheel has unexpected file name: expected '0.4.0.dev4+g355b28e.d20241231122314', got '0.4.0.dev4+g355b28e.d20241231122325'
[...]
Failed to build Ragna
ERROR: ERROR: Failed to build installable wheels for some pyproject.toml based projects (Ragna)
```

I've debugged this a little and I can reproduce the issue:

1. Create a fresh development environment with `conda env create -f environment-dev.yaml`
2. Add some random change to any file (this leads to `setuptools_scm` recognizing the state as dirty and change the version accordingly)
3. Run `pip install .`

This is likely a bug in `setuptools_scm`. I've tried to compile a minimal reproducible example, but so far wasn't successful. First step is to come up with such an example so we can report it.

However, we should also look into workarounds for our RTD build in the meantime. I'd like us to somehow set the `SETUPTOOLS_SCM_PRETEND_VERSION_FOR_RAGNA` environment variable to avoid other shenanigans as well (see pypa/setuptools-scm#962). However, we can't set environment variables directly as each command in the hooks that RTD runs is run as an independent process. Maybe we are lucky and can set the environment variable in a configuration file implicitly loaded by the shell, i.e. `~/.bashrc`?

smokestacklightnin:
This fixes #529 by changing the naming scheme for wheels created for non-editable installation

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -65,7 +65,7 @@ all = [
 [tool.setuptools_scm]
 write_to = "ragna/_version.py"
 version_scheme = "release-branch-semver"
-local_scheme = "node-and-timestamp"
+local_scheme = "node-and-date"
 
 [project.scripts]
 ragna = "ragna.__main__:app"


pmeier (approved):
Thanks William. Note that the local scheme is not just applied for non-editable installs, but we only observed the bug in this use case.

Issue #542, opened 2025-01-24
Add support for Python 3.13


pmeier:
Assuming we are not hitting any compatibility issues, this should be as easy as add the relevant compatibility marker in `pyproject.toml` and adjust the CI matrix to include it.

smokestacklightnin:
After testing locally, Python 3.13 seems to be affected by #522

smokestacklightnin:
Completed with #587

PR #544, opened 2025-01-25
Add support for Python 3.13

smokestacklightnin:
Addresses #542 

The included changes, per the issue, are:

- Add a compatibility marker in `pyproject.toml`
- Include Python 3.13 in tests in CI


smokestacklightnin:
We seem to be getting a [segmentation fault](https://github.com/Quansight/ragna/actions/runs/12960472716/job/36154375708?pr=544#step:4) that is reminiscent of #522, but only with Python 3.13 and not Python 3.11 or 3.12.

PR #546, opened 2025-01-28
Protect against local DNS misconfiguration

smokestacklightnin:
The changes included in this PR make the tests more resilient against misconfigured local DNS by using `127.0.0.1` instead of `localhost`.


pmeier (approved):
Thanks!

Issue #549, opened 2025-02-02
[RFD] Change wording to prevent misunderstanding


smokestacklightnin:
### Feature description

The title for the following [heading in the FAQ](https://ragna.chat/en/latest/references/faq/#why-should-i-use-ragna-and-not-x)
https://github.com/Quansight/ragna/blob/2a3c57123988526e2309dc2ee6fa18bb1e2b87c8/docs/references/faq.md?plain=1#L3
could possibly be confused with a certain social network. 

I propose that we change the `X` to something like underscores `___` or an ellipsis `...`.

I'm making this issue to get suggestions or opinions for what to replace the `X` with if others have better ideas than underscores or ellipses.

pmeier:
Thanks for spotting this. TBH, I didn't think about this when I put in the header, but now that you brought this up, I fully agree that we should avoid any confusion. I don't think we need to discuss this in a larger group. IMO using ellipsis is fine. Feel free to send a PR.

PR #550, opened 2025-02-03
[FIX] Change `X` to ellipsis `...`

smokestacklightnin:
This mini PR addresses #549 by changing the heading containing an `X` to an ellipsis `...`.


pmeier (approved):
Thanks!

PR #558, opened 2025-03-11
Allow `Assistant`s to specify their logos

smokestacklightnin:
Adds an `avatar` property to the `Assistant` and integrates its use into the UI.

### DIFF
--- a/ragna/core/_components.py
+++ b/ragna/core/_components.py
@@ -302,3 +302,8 @@ class Assistant(Component, abc.ABC):
             Answer.
         """
         ...
+
+    @property
+    def avatar(self) -> str:
+        """Return a default avatar for an assistant in a chat."""
+        return "imgs/ragna_logo.svg"


smokestacklightnin:
@pmeier I [removed `_ragna_avatar_lookup`](https://github.com/Quansight/ragna/pull/558/commits/585b5005b63794126dc757c02bd5b0bc9945d7be) because the logic was simplified. If you'd like me to move the logic back to a separate function, I can.

ragna/deploy/_ui/central_view.py
```diff
@@ -57,13 +56,16 @@ def __init__(
         timestamp=None,
         show_timestamp=True,
         assistant_toolbar_visible=True,  # hide the toolbar during streaming
+        engine=None,
```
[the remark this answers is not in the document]
smokestacklightnin: Fixed

ragna/deploy/_ui/central_view.py
```diff
@@ -57,6 +57,8 @@ def __init__(
         timestamp=None,
         show_timestamp=True,
         assistant_toolbar_visible=True,  # hide the toolbar during streaming
+        avatar=None,
```
pmeier, line 60: Is there ever a case where we want to pass the `avatar` directly instead of going through the lookup? If not, let's remove it from the input parameters.

ragna/deploy/_ui/central_view.py
```diff
@@ -163,9 +143,43 @@ def _build_message(self, *args, **kwargs) -> Optional[RagnaChatMessage]:
...
-            message.object, role="user", user=cast(str, pn.state.user)
+            message.object,
+            role="user",
+            user=cast(str, pn.state.user),
+            avatar_lookup=AvatarLookup(),
+        )
+
+
+class AvatarLookup:
+    def __init__(self, engine=None):
```
pmeier, line 154: In what case do we not have the engine when instantiating?

smokestacklightnin: [Here](https://github.com/Quansight/ragna/blob/0a3be0dafc7e80d2b872a4c9cee1537e7777c385/ragna/deploy/_ui/central_view.py#L134) and [here](https://github.com/Quansight/ragna/blob/0a3be0dafc7e80d2b872a4c9cee1537e7777c385/ragna/deploy/_ui/central_view.py#L149)

ragna/deploy/_ui/central_view.py
smokestacklightnin: No. I included it for flexibility and generality.

smokestacklightnin: Removed.

ragna/deploy/_ui/central_view.py
pmeier: What about passing the `avatar_lookup` to the `ChatInterface` so we actually only have to create it once?

smokestacklightnin: > What about passing the `avatar_lookup` to the `ChatInterface` so we actually only have to create it once?

So putting it in an otherwise empty `__init__` method? I don't mind doing that

pmeier (approved):
LGTM, thanks William!

PR #559, opened 2025-03-17
Add lines to git-related files for Pixi

smokestacklightnin:
Add a few lines to `.gitignore` and `.gitattributes` for a better experience for developers of Ragna who are using [Pixi](https://pixi.sh/).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -168,3 +168,7 @@ cython_debug/
 
 # VSCode
 .vscode/
+
+# pixi environments
+.pixi
+*.egg-info
--- a/.gitignore
+++ b/.gitignore
@@ -171,4 +171,3 @@ cython_debug/
 
 # pixi environments
 .pixi
-*.egg-info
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -168,3 +168,7 @@ cython_debug/
 
 # VSCode
 .vscode/
+
+# pixi environments
+.pixi
+*.egg-info
```
pmeier, line 174: That is not `pixi` specific, but rather a Python feature. As such, we already are ignoring it:

https://github.com/Quansight/ragna/blob/00ffa2b9aa990649e099e0b8d3e8cf9d7c81bc4c/.gitignore#L28

pmeier (approved):
Thanks William. Can be merged when the comment is addressed.

.gitignore
smokestacklightnin: Good point. Pixi should probably check for the presence of that line in the file before blindly adding it.

Issue #555, opened 2025-03-06
A new chat based on a corpus is not selected automatically


pmeier:
Per title. Chats that are started based on a corpus are never manually prepared and thus do not get a welcome message

https://github.com/Quansight/ragna/blob/00ffa2b9aa990649e099e0b8d3e8cf9d7c81bc4c/ragna/core/_rag.py#L260-L272

IIRC this was never a conscious decision, but rather an artifact of how we implemented it. We should probably revisit that and discuss whether each chat should have a welcome message.

Without the message, these chats are empty and thus sorted at the back of the list in the UI

https://github.com/Quansight/ragna/blob/00ffa2b9aa990649e099e0b8d3e8cf9d7c81bc4c/ragna/deploy/_ui/left_sidebar.py#L65-L72

That in itself is a problem, but even worse, the new chat is not even selected by default. Meaning, users by default will the "old" chat they were on before, which leads to confusion. This needs to be fixed.

PR #560, opened 2025-03-20
Add `datetime_created` to `Chat`s and select new chat based on corpus automatically

smokestacklightnin:
Addresses #555

### DIFF
--- a/ragna/core/_rag.py
+++ b/ragna/core/_rag.py
@@ -222,6 +222,7 @@ class Chat:
         source_storage: Source storage to use.
         assistant: Assistant to use.
         corpus_name: Corpus of documents to use.
+        datetime_created: Override datetime of creation
         **params: Additional parameters passed to the source storage and assistant.
     """
 
@@ -240,6 +241,7 @@ class Chat:
         source_storage: SourceStorage,
         assistant: Assistant,
         corpus_name: str = "default",
+        datetime_created: datetime.datetime | None = None,
         **params: Any,
     ) -> None:
         self._rag = rag
@@ -257,8 +259,8 @@ class Chat:
 
         self._messages: list[Message] = []
 
-        self.datetime_created: datetime.datetime = datetime.datetime.now(
-            tz=datetime.timezone.utc
+        self.datetime_created: datetime.datetime = (
+            datetime_created or datetime.datetime.now(tz=datetime.timezone.utc)
         )
 
     async def prepare(self) -> Message:
--- a/ragna/core/_rag.py
+++ b/ragna/core/_rag.py
@@ -257,6 +257,10 @@ class Chat:
 
         self._messages: list[Message] = []
 
+        self.datetime_created: datetime.datetime = datetime.datetime.now(
+            tz=datetime.timezone.utc
+        )
+
     async def prepare(self) -> Message:
         """Prepare the chat.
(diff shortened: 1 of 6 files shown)


ragna/core/_rag.py
```diff
@@ -257,6 +259,10 @@ def __init__(
 
         self._messages: list[Message] = []
 
+        self.created_at: datetime.datetime = created_at or datetime.datetime.now(
+            tz=datetime.timezone.utc
+        )
```
pmeier, line 264: We should make `created_at` part of the `SpecialChatParams` above and make the default chat name dependent on it.

ragna/deploy/_engine.py
```diff
@@ -220,6 +221,7 @@ def create_chat(
             metadata_filter=metadata_filter,
             documents=documents,
             prepared=prepared,
+            created_at=datetime.now(tz=timezone.utc),
```
pmeier, line 224: Instead of explicitly passing this here, can this be the default on `schemas.Chat`?

pmeier (commented):
Looking pretty good. A few comments.

ragna/core/_rag.py
smokestacklightnin: Done

ragna/deploy/_engine.py
smokestacklightnin: Done

ragna/core/_rag.py
```diff
@@ -249,10 +255,7 @@ def __init__(
         self.assistant = assistant
         self.corpus_name = corpus_name
 
-        special_params = SpecialChatParams().model_dump()
-        special_params.update(params)
-        params = special_params
-        self.params = params
+        self.params = SpecialChatParams(**params).model_dump()
         self._unpacked_params = self._unpack_chat_params(params)
```
pmeier, line 259: This has to be 
```suggestion
        self._unpacked_params = self._unpack_chat_params(self.params)
```

ragna/core/_rag.py
```diff
@@ -222,6 +227,7 @@ class Chat:
         source_storage: Source storage to use.
         assistant: Assistant to use.
         corpus_name: Corpus of documents to use.
+        created_at: Override datetime of creation
```
pmeier, line 230: Remove for now as we haven't spelled out the other special params as well.

ragna/deploy/_engine.py
```diff
@@ -384,4 +385,5 @@ def chat(self, chat: core.Chat) -> schemas.Chat:
             params=params,
             messages=[self.message(message) for message in chat._messages],
             prepared=chat._prepared,
+            created_at=chat.params["created_at"],
```
pmeier, line 388: Move this up to the `id` and `name` field and also use `pop` to avoid leaving it in the params.

ragna/core/_rag.py
smokestacklightnin: Fixed

ragna/core/_rag.py
smokestacklightnin: Removed

ragna/deploy/_engine.py
smokestacklightnin: Done

PR #561, opened 2025-03-21
fix default factories in merge_models

pmeier:
This is needed for #560, where we add a `default_factory` that takes the formerly validated fields as input and thus cannot be invoked without them.

The only reason we have invoked the `default_factory` before this PR is to determine whether or not we have multiple default values. However, that reasoning is flawed in two ways:

1. An invocation of the `default_factory` only captures a snapshot. Meaning, even a simple `default_factory` that returns the current datetime, will produce two different results when called multiple times. Thus, although multiple models might use the same `default_factory`, it will not be propagated to the merged model.
2. Even if there is only a single `default_factory` for a given field, we are completely removing the dynamic element of the `default_factory` by turning it into a static default value.

This PR resolves both issues above by using the `default_factory` directly for determining whether there are multiple default values. In addition it also propagates the `default_factory` if there is only one such value.

### DIFF
--- a/ragna/core/_utils.py
+++ b/ragna/core/_utils.py
@@ -7,7 +7,7 @@ import importlib
 import importlib.metadata
 import os
 from collections import defaultdict
-from typing import Any, Callable, Collection, Optional, Type, Union, cast
+from typing import Any, Collection, Optional, Type, Union, cast
 
 import packaging.requirements
 import pydantic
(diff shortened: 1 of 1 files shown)


smokestacklightnin (approved):
This seems to fix the issue I was having in #560. Thanks @pmeier!

PR #562, opened 2025-04-06
Retrieve all points for metadata in Qdrant source storage

### LINKED ISSUE #557: Investigate limited selection on metadata filters
We have an internal report that although 200+ values were available for a specific metadata key in a given corpus, the UI only displayed 8. This needs to be investigated and, if it turns out to be true, fixed. Easiest solution is probably just a scroll bar, although the UX might not be optimal.

smokestacklightnin:
Fixes #557 by retrieving metadata for all points.


smokestacklightnin:
I marked this issue as a draft because it will either need to be modified because of, or superseded by #570

smokestacklightnin:
Superseded by #574

PR #563, opened 2025-04-07
fix deprecation warnings coming from pydantic 2.11

pmeier:

PR #564, opened 2025-04-07
update chroma to 1.*

pmeier:
Sister PR to #563. Both together should fix all CI issues seen for example in #562.


pyproject.toml
```diff
@@ -102,6 +102,8 @@ testpaths = [
 filterwarnings = [
     "error",
     "ignore::ResourceWarning",
+    # chromadb
+    "ignore:Python 3.14 will:DeprecationWarning",
```
pmeier, line 106: I guess we never even hit this issue in #563 as it was shadowed by other things. We need to re-add it.

smokestacklightnin (approved):
Checks are passing, and I approve.

Issue #567, opened 2025-04-08
Use async source storage clients


smokestacklightnin:
### Feature description

Even with native Ragna machinery like `as_awaitable`, that machinery wraps synchronous tools, for example source storage clients, we will not get the full benefit of async. Particularly, if we want to fetch data/metadata from a source storage in parallel, we necessarily must use an async client (assuming we use a client instead of http requests).

I have checked, and all three currently supported source storages, Chroma, Lance, and Qdrant have async clients, so it is just a matter of making the migration.

PR #573, opened 2025-04-09
Use `async` for Qdrant

smokestacklightnin:
This PR changes Ragna to use `async` for Qdrant source storage client.

### DIFF
--- a/ragna/source_storages/_demo.py
+++ b/ragna/source_storages/_demo.py
@@ -52,7 +52,7 @@ class RagnaDemoSourceStorage(SourceStorage):
 
         return corpus
 
-    def list_metadata(
+    async def list_metadata(
         self, corpus_name: Optional[str] = None
     ) -> dict[str, dict[str, tuple[str, list[Any]]]]:
         if corpus_name is None:
@@ -79,7 +79,7 @@ class RagnaDemoSourceStorage(SourceStorage):
 
         return metadata
 
-    def store(self, corpus_name: str, documents: list[Document]) -> None:
+    async def store(self, corpus_name: str, documents: list[Document]) -> None:
         corpus = self._get_corpus(corpus_name, create=True)
         # FIXME: handle updating metadata (either introducing new or filling with None)
         #  and add a type check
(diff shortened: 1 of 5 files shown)


ragna/source_storages/_demo.py
pmeier, line 1: No need for any changes here. The demo source storage can just stay synchronous

tests/source_storages/test_source_storages.py
```diff
@@ -107,7 +107,8 @@
 @pytest.mark.parametrize(
     "source_storage_cls", set(SOURCE_STORAGES) - {RagnaDemoSourceStorage}
 )
-def test_smoke(
+@pytest.mark.asyncio
+async def test_smoke(
```
pmeier, line 111: This is correct. We basically cheated a little previously since all of our builtin source storages were sync.

tests/source_storages/test_source_storages.py
```diff
@@ -133,11 +134,11 @@ def test_smoke(
     source_storage = source_storage_cls()
     corpus_name = "default"
 
-    source_storage.store(corpus_name, documents)
+    await source_storage.store(corpus_name, documents)
```
pmeier, line 137: As explained, we do not know at this point whether or not `source_storage.store` is sync or async as both are allowed. Fortunately, we already have the right tools:

```suggestion
    await as_awaitable(source_storage.store, corpus_name, documents)
```

The same also applies to all other occurrences below.

pmeier (changes requested):
It seems we have misunderstood each other. When I said "make the source storages async", I meant #567 and its respective sub issues. The `ragna.core.SourceStorage` base class as well as the handling of the source storages inside `ragna.core.Rag` should not change. And they don't have to either. Since we are using `as_awaitable` in whenever we are calling potentially user-defined methods, we give the user the flexibility to make them sync or async. Please have a look at [this tutorial](https://ragna.chat/en/stable/generated/tutorials/gallery_custom_components) and specifically at [this section](https://ragna.chat/en/stable/generated/tutorials/gallery_custom_components/#sync-or-async)

tests/source_storages/test_source_storages.py
smokestacklightnin: Done

pmeier (approved):
LGTM, thanks William!

PR #574, opened 2025-04-10
Parallelize fetching metadata for Qdrant

smokestacklightnin:
This PR parallelizes fetching metadata for Qdrant for performance purposes. It parallelizes both across corpuses, as well as across pages of metadata for a given corpus.

Tests are also modified so that test corpuses are large enough to necessarily require pagination.

#557 is addressed in the process of implementing the fetching of metadata for a given corpus.

### DIFF
--- a/ragna/source_storages/_qdrant.py
+++ b/ragna/source_storages/_qdrant.py
@@ -86,7 +86,7 @@ class Qdrant(VectorDatabaseSourceStorage):
 
     async def _fetch_metadata(self, *args: Any, **kwargs: Any):
         limit = kwargs.pop("limit", None)
-        _ = kwargs.pop("with_payload", None)
+        with_payload = kwargs.pop("with_payload", None)
 
         ids = []
         while True:
@@ -115,7 +115,7 @@ class Qdrant(VectorDatabaseSourceStorage):
                 *[
                     self._client.scroll(
                         *args,
-                        with_payload=True,
+                        with_payload=with_payload,
                         limit=limit,
                         offset=offset,
                         **kwargs,
--- a/ragna/source_storages/_qdrant.py
+++ b/ragna/source_storages/_qdrant.py
@@ -125,6 +125,26 @@ class Qdrant(VectorDatabaseSourceStorage):
         ):
             yield payload
 
+    async def _transform_metadata(self, corpus_name: str) -> dict[str, Any]:
+        corpus_metadata = defaultdict(set)
+        async for point in self._fetch_metadata(corpus_name=corpus_name):
+            for key, value in point.items():
+                if any(
+                    [
+                        (key.startswith("__") and key.endswith("__")),
+                        key == self.DOC_CONTENT_KEY,
+                        not value,
+                    ]
+                ):
+                    continue
+
+                corpus_metadata[key].add(value)
+
+        return {
+            key: ({type(value).__name__ for value in values}.pop(), sorted(values))
+            for key, values in corpus_metadata.items()
+        }
+
     async def list_metadata(
         self, corpus_name: Optional[str] = None
     ) -> dict[str, dict[str, tuple[str, list[Any]]]]:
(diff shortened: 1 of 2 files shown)


ragna/source_storages/_qdrant.py
```diff
@@ -82,6 +84,47 @@ async def _ensure_table(self, corpus_name: str, *, create: bool = False) -> None
         elif non_existing_corpus:
             raise_non_existing_corpus(self, corpus_name)
 
+    async def _fetch_metadata(self, *args: Any, **kwargs: Any):
```
pmeier, line 87: Why are we using `*args`, `**kwargs` here?

ragna/source_storages/_qdrant.py
```diff
@@ -82,6 +84,47 @@ async def _ensure_table(self, corpus_name: str, *, create: bool = False) -> None
...
+            for records in await asyncio.gather(
+                *[
+                    self._client.scroll(
+                        *args,
+                        with_payload=True,
+                        limit=limit,
+                        offset=offset,
+                        **kwargs,
+                    )
+                    for offset in ids[::limit]
```
pmeier, line 123: So the offset is not an integer in terms of pagination, but rather the actual ID of the record we want to start "scrolling" from?

smokestacklightnin: Correct

> If provided - skip points with ids less than given offset
([docs](https://python-client.qdrant.tech/qdrant_client.async_qdrant_client#qdrant_client.async_qdrant_client.AsyncQdrantClient.scroll))

ragna/source_storages/_qdrant.py
smokestacklightnin: They are passed to the client. Since this is basically a wrapper around `scroll` that fetches all the metadata, we want the same signature. Those args are how we let `scroll` know which corpus(es) we want and whether to include the payload.

ragna/source_storages/_qdrant.py
```diff
@@ -82,6 +84,47 @@ async def _ensure_table(self, corpus_name: str, *, create: bool = False) -> None
...
+                break
+
+        # This limit is purely heuristic.
+        # There is no way to know a priori the size of the metadata, so
+        # we just limit ourselves to twenty requests to the database.
+        # This can change in the future.
+        limit = limit if limit is not None else max(len(ids) // 20, 10)
+
+        return itertools.chain.from_iterable(
+            records[0]
```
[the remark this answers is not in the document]
smokestacklightnin: `records` is a list of length at most `limit` of pairs of `Record`s together with offsets that we discard. 

I will clean it up with 

```python
records for records, _ in ...
```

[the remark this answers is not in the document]
smokestacklightnin: Changed to `dict[str,Any]`

ragna/source_storages/_qdrant.py
smokestacklightnin: Changed to return the payload

ragna/source_storages/_qdrant.py
```diff
@@ -82,6 +84,67 @@ async def _ensure_table(self, corpus_name: str, *, create: bool = False) -> None
...
+
+        # This limit is purely heuristic.
+        # There is no way to know a priori the size of the metadata, so
+        # we just limit ourselves to twenty requests to the database.
+        # This can change in the future.
+        limit: int = limit if limit is not None else max(len(ids) // 20, 10)
+
+        for payload in itertools.chain.from_iterable(
+            (cast(dict[str, Any], record.payload) for record in records)
+            for records, _ in await asyncio.gather(
```
pmeier, line 114: Since we now return an iterator, we also don't have to `await` the completion of everything here. You can use [`asyncio.as_completed`](https://docs.python.org/3/library/asyncio-task.html#asyncio.as_completed) for that. This means, we instantly yield if one of the `client.scroll` results is available.

The overall loop could look like

```python
for result in asyncio.as_completed([client.scroll(...) for ...]):
    records, _ = await result
    for record in records:
        yield cast(dict[str, Any], record.payload)
```

ragna/source_storages/_qdrant.py
```diff
@@ -82,6 +84,67 @@ async def _ensure_table(self, corpus_name: str, *, create: bool = False) -> None
         elif non_existing_corpus:
             raise_non_existing_corpus(self, corpus_name)
 
+    async def _fetch_raw_metadata_entries(
+        self, *, corpus_name: str, limit: Optional[int] = None
```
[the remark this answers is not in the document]
smokestacklightnin: Done

ragna/source_storages/_qdrant.py
smokestacklightnin: Made the change

pmeier (approved):
LGTM when CI is green. Thanks William!

PR #580, opened 2025-04-11
Run `update-docker-requirements.yml` on pull request for feedback

smokestacklightnin:
Gives the developer feedback when making changes to dependencies so as not to break the generation of [`requirements-docker.lock`](https://github.com/Quansight/ragna/blob/main/requirements-docker.lock)


pmeier (approved):
LGTM if CI is green. Thanks William!

PR #584, opened 2025-05-20
Include chat histories

smokestacklightnin:
This PR sends the full chat histories for each prompt for all assistants.


ragna/assistants/_utils.py
```diff
@@ -0,0 +1,15 @@
...
+
+def unpack_prompts_and_sources(
+    messages: list[Message],
+) -> tuple[list[str], list[Source]]:
+    return [message.content for message in messages if message.role != "system"], list(
+        set(
+            itertools.chain.from_iterable(
+                message.sources for message in messages if message.role != "system"
+            )
+        )
```
pmeier, line 14: Hmm, not sure about that. Shouldn't we just send the sources for the current message? Otherwise, the assistant might use sources that are relevant to this chat, but not to the last question asked.

ragna/assistants/_openai.py
```diff
@@ -40,10 +41,13 @@ def _call_openai_api(
                     "role": "system",
                     "content": self._make_system_content(sources),
                 },
-                {
-                    "role": "user",
-                    "content": prompt,
-                },
+                *(
+                    {
+                        "role": "assistant" if idx % 2 else "user",
```
pmeier, line 46: I don't love this. This requires a specific structure on the `prompts` input (misnomer now since it also contains answers). Can't this method take `messages: list[Message]` and handle the extraction internally?

ragna/assistants/_google.py
```diff
@@ -14,29 +15,47 @@ class GoogleAssistant(HttpApiAssistant):
     def display_name(cls) -> str:
         return f"Google/{cls._MODEL}"
 
-    def _instructize_prompt(self, prompt: str, sources: list[Source]) -> str:
+    def _instructize_system_prompt(self, sources: list[Source]) -> str:
         # https://ai.google.dev/docs/prompt_best_practices#add-contextual-information
         return "\n".join(
             [
-                "Answer the prompt using only the pieces of context below.",
+                "Answer the user prompts using only the pieces of context below.",
```
pmeier, line 22: We don't want to prompt the assistant to answer all prompts. The system prompt should be something like "These are the previous user prompts and you answers. The latest user prompt is ... and here are the sources you must use to answer it."

pmeier (commented):
> Only the Ollama assistant is tested because I do not have API keys for the remaining assistants, but it is identical logic across them all.

This is not your fault or responsibility, but we shouldn't merge without testing them. Let me work on a solution for that.

ragna/assistants/_openai.py
smokestacklightnin: > Can't this method take messages: `list[Message]` and handle the extraction internally?

I don't mind doing this. I just didn't want to change the function signature. I'll look into making the change.

smokestacklightnin: Made the change.

ragna/assistants/_utils.py
smokestacklightnin: Fixed.

ragna/assistants/_google.py
smokestacklightnin: The system prompt is given in a field separate from the message contents. 

I have changed the first line of the system prompt to

> Answer the last user prompt using only the pieces of context below.

ragna/assistants/_ai21labs.py
```diff
@@ -28,7 +28,9 @@ async def answer(
         # See https://docs.ai21.com/reference/j2-chat-api#chat-api-parameters
         # See https://docs.ai21.com/reference/j2-complete-api-ref#api-parameters
         # See https://docs.ai21.com/reference/j2-chat-api#understanding-the-response
-        prompt, sources = (message := messages[-1]).content, message.sources
+        *_, current_message = (
+            message for message in messages if message.role != "system"
+        )
```
pmeier, line 33: Likely the same output in most situations, but why not search for the last user message?

```suggestion
        current_prompt = next(m for m in reversed(messages) if m.role == "user")
```

ragna/assistants/_google.py
```diff
@@ -14,29 +14,55 @@ class GoogleAssistant(HttpApiAssistant):
...
+                    {
+                        "parts": [
+                            {
+                                "text": self._instructize_system_prompt(
+                                    current_message.sources
+                                ),
+                            }
+                        ]
+                    }
+                ],
```
pmeier, line 50: Please unexplode this if possible. If not, please create a `text` variable above and use an unexploded formatting then.

ragna/assistants/_google.py
```diff
@@ -14,29 +14,55 @@ class GoogleAssistant(HttpApiAssistant):
...
+                    {
+                        "role": {
+                            "user": "user",
+                            "assistant": "model",
+                        }[message.role],
+                        "parts": [
+                            {
+                                "text": message.content,
+                            },
+                        ],
```
pmeier, line 62:
```suggestion
                        "parts": [{"text": message.content}],
```

ragna/assistants/_ai21labs.py
smokestacklightnin: Yes, duh. This is what I wanted. Should have been obvious to me :roll_eyes:

ragna/assistants/_google.py
smokestacklightnin: Done

ragna/assistants/_google.py
smokestacklightnin: Done

PR #581, opened 2025-04-17
Migrate to Pixi

smokestacklightnin:
This PR migrates Ragna to use Pixi instead of Conda/Mamba.

Changes include using Pixi configuration and all package dependencies in `pyproject.toml`. 

CI workflows are also migrated. 

Remaining tasks:

- [x] ~Update `update-docker-requirements.yml`~ Remove `update-docker-requirements.yml` and `docker-requirements.lock`
- [x] Update `Dockerfile`
- [x] Update the Read The Docs build
- [x] Document all changes
- [x] Move Pixi config to the top of `pyproject.toml`

I can also [automatically update lockfiles](https://pixi.sh/dev/integration/ci/updates_github_actions/) if @pmeier thinks it would be useful.

PR #585, opened 2025-05-21
Followup to Pixi migration

smokestacklightnin:
This PR contains two minor changes as a followup to #581:

1. Update contribution docs to include installation instructions for dev environment
2. Remove `pip` as an explicit direct dependency


.github/workflows/lint.yml
```diff
@@ -65,6 +65,7 @@ jobs:
         uses: ./.github/actions/setup-env
         with:
           environment: "dev-all-py310"
+          editable: "true"
```
pmeier, line 68: Have you tried running this with a non-editable install? Because it seems this is the only time we use an editable install. And if this doesn't need it, we can get rid of the `sed` and all the other problems this brings.

smokestacklightnin: The `sed` is required to make the installation non-editable. The default config in pyproject.toml is to be editable for development; the `sed` makes it non-editable.

The type checks seem to run with a non-editable installation.

pmeier (commented):
> 2. Install Ragna in non-editable mode in Docker image

> 5. Use non-editable installation of Ragna by default in the CI

There is prefix-dev/pixi#3762 that requests official support for this. The issue is already assigned. Since we are not fixing any issues here, I'd rather not roll a custom solution.

> 3. Use Python 3.10 in Docker image in order to avoid issues described in #522

The reason we are using a higher than minimal version in Docker is that they usually bring performance upgrades. And since we can control the Python version we should just take this free win.

Since the underlying issue will be fixed in #586 can we revert the downgrade here and include an upgrade to Python 3.13 in #587?

smokestacklightnin:
> > 2. Install Ragna in non-editable mode in Docker image
> 
> > 5. Use non-editable installation of Ragna by default in the CI
> 
> There is [prefix-dev/pixi#3762](https://github.com/prefix-dev/pixi/issues/3762) that requests official support for this. The issue is already assigned. Since we are not fixing any issues here, I'd rather not roll a custom solution.
> 
> > 3. Use Python 3.10 in Docker image in order to avoid issues described in [investigate pymupdf segfaults #522](https://github.com/Quansight/ragna/issues/522)
> 
> The reason we are using a higher than minimal version in Docker is that they usually bring performance upgrades. And since we can control the Python version we should just take this free win.
> 
> Since the underlying issue will be fixed in #586 can we revert the downgrade here and include an upgrade to Python 3.13 in #587?

I have removed all the commits dealing with the editable installation

pmeier (approved):

PR #586, opened 2025-05-22
Fix PyMuPDF segfault

### LINKED ISSUE #522: investigate pymupdf segfaults
The upper pin in

https://github.com/Quansight/ragna/blob/71acf83d4114514c56e298f7acc431551933e522/ragna/core/_document.py#L226

exists, because I saw segfaults when running `pytest` on Python 3.11 and 3.12. I couldn't find anything on their issue tracker regarding this. We should come up with a minimal example that reproduces the issue and report it so we are not stuck on an outdated version for eternity. If an issue for the segfault already exists, we should add a code comment linking it.

smokestacklightnin:
This provisionally fixes #522 by adding `"ignore::DeprecationWarning"` to the Pytest `filterwarnings` section.

The version of PyMuPDF is also upgraded to 1.25.5.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -63,7 +63,7 @@ all = [
     "ijson",
     "lancedb>=0.2",
     "pyarrow",
-    "pymupdf<=1.24.10,>=1.23.6",
+    "pymupdf<1.26.0,>=1.25.0",
     "python-docx",
     "python-pptx",
     "qdrant-client>=1.12.1",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -196,6 +196,8 @@ filterwarnings = [
     "ignore:Python 3.14 will:DeprecationWarning",
     # https://github.com/lancedb/lancedb/issues/2315
     "ignore:Support for class-based `config`:pydantic.warnings.PydanticDeprecatedSince20",
+    # https://github.com/pymupdf/PyMuPDF/issues/4392
+    "ignore::DeprecationWarning",
 ]
 xfail_strict = true
 
--- a/ragna/core/_document.py
+++ b/ragna/core/_document.py
@@ -231,7 +231,7 @@ class PdfDocumentHandler(DocumentHandler):
 
     @classmethod
     def requirements(cls) -> list[Requirement]:
-        return [PackageRequirement("pymupdf<=1.24.10,>=1.23.6")]
+        return [PackageRequirement("pymupdf<1.26.0,>=1.25.0")]
 
     @classmethod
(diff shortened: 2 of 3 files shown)


pyproject.toml
```diff
@@ -196,6 +196,8 @@ filterwarnings = [
     "ignore:Python 3.14 will:DeprecationWarning",
     # https://github.com/lancedb/lancedb/issues/2315
     "ignore:Support for class-based `config`:pydantic.warnings.PydanticDeprecatedSince20",
+    # https://github.com/pymupdf/PyMuPDF/issues/4392
+    "ignore::DeprecationWarning",
```
pmeier, line 200: Could you try 

```suggestion
    "ignore:builtin type \w+ has no __module__ attribute:DeprecationWarning",
```

following the warnings listed in swig/swig@49a7d2b0c6aed00bd07a9a28569020d560c2658c

ragna/core/_document.py
```diff
@@ -231,7 +231,7 @@ class PdfDocumentHandler(DocumentHandler):
 
     @classmethod
     def requirements(cls) -> list[Requirement]:
-        return [PackageRequirement("pymupdf<=1.24.10,>=1.23.6")]
+        return [PackageRequirement("pymupdf<1.26.0,>=1.25.0")]
```
pmeier, line 234: 1. Why the upper pin?
2. Nit: if we keep the upper pin, I prefer to have it at the end. I read them from left to right and seeing the lower pin first makes more sense to me.

pyproject.toml
smokestacklightnin: FYI @pmeier we need to escape the slash:

```
"ignore:builtin type \\w+ has no __module__ attribute:DeprecationWarning"
```

Made the change.

ragna/core/_document.py
smokestacklightnin: 1. So we don't experience unexpected errors if something in PyMuPDF changes. This is not specific to PyMuPDF.
2. I agree that it is more intuitive to have the lower endpoint listed before the upper endpoint. I put it that way in to match the output of the update optional dependencies script in pyproject.toml.

@pmeier If you're comfortable removing the pins, then so am I.

pmeier: Yes, let's remove. We don't need to restrict users just because we turn warnings into errors in our test suite.

pmeier (approved):
LGTM after https://github.com/Quansight/ragna/pull/586#discussion_r2101869588 is resolved. Thanks William.

PR #587, opened 2025-05-22
Add support for Python 3.13

smokestacklightnin:
Adds support for Python 3.13 after fixing PyMuPDF segfaults in #586.

Also adds `"ignore::SyntaxWarning"` to Pytest filterwarnings. See https://github.com/qdrant/qdrant-client/issues/983

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -43,6 +43,8 @@ jobs:
             environment: "dev-all-py311"
           - os: ubuntu-latest
             environment: "dev-all-py312"
+          - os: ubuntu-latest
+            environment: "dev-all-py313"
 
       fail-fast: false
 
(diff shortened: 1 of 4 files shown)


pmeier:
> Also adds "ignore::SyntaxWarning" to Pytest filterwarnings. See https://github.com/qdrant/qdrant-client/issues/983

Can we please split this into a separate PR as this has nothing to do with the 3.13 upgrade? According to the issue this happens for >= 3.12. When this is in a separate PR, we can just revert it when the fix is released by Qdrant.

smokestacklightnin:
> > Also adds "ignore::SyntaxWarning" to Pytest filterwarnings. See [qdrant/qdrant-client#983](https://github.com/qdrant/qdrant-client/issues/983)
> 
> Can we please split this into a separate PR as this has nothing to do with the 3.13 upgrade? According to the issue this happens for >= 3.12. When this is in a separate PR, we can just revert it when the fix is released by Qdrant.

my guess is that the qdrant version got updated in the proces of adding python 3.13, so it really did show up in the 3.13 work

pmeier (approved):

PR #588, opened 2025-05-27
Release workflow

### LINKED ISSUE #367: improve release process
Currently, I'm manually creating and publishing releases. This has two downsides:

1. Releasing involves quite a few steps. Doing them manually opens up the window for mistakes.
2. All the knowledge on how to release currently is in my head and not documented anywhere else. If I'm unable to do a release for some reason, others will have a hard time replicating my process.

To overcome this, I propose two things:

1. Automate as much as possible through GitHub Actions. 
2. Write up the release process in the documentation so everyone with the right permissions can do it.

smokestacklightnin:
This resolves #367 by automating the release process.

A reusable composite action `build` is included that builds wheels.


.github/actions/build/action.yml
```diff
@@ -0,0 +1,28 @@
...
+
+  steps:
+    - name: Build wheel
+      shell: bash
+      run: SOURCE_DATE_EPOCH=$(git log -1 --pretty=%ct) pixi run build
+
+    - name: "Generate hashes"
+      id: hashes
+      shell: bash
+      run: cd dist && echo "hashes=$(sha256sum * | base64 -w0)" >> $GITHUB_OUTPUT
```
pmeier, line 20: What are these hashes for?

.github/actions/build/action.yml
```diff
@@ -0,0 +1,28 @@
...
+
+    - name: "Upload dists"
+      uses: "actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02"
+      with:
+        name: "dist"
+        path: "dist/"
+        if-no-files-found: error
+        retention-days: 5
```
pmeier, line 28: Why are we upload the package data to GH if it gets uploaded to PyPI anyway? What's the benefit of having this here over downloading from PyPI if required?

.github/workflows/release.yml
```diff
@@ -0,0 +1,70 @@
...
+    name: Provenance
+    needs: [build]
+    permissions:
+      actions: read
+      contents: write
+      id-token: write # Needed to access the workflow's OIDC identity.
+    uses: "slsa-framework/slsa-github-generator/.github/workflows/generator_generic_slsa3.yml@v2.1.0"
+    with:
+      base64-subjects: "${{ needs.build.outputs.hashes }}"
+      upload-assets: true
```
pmeier, line 41: What is this for? Is this recommended practice by the PyPA?

.github/workflows/release.yml
```diff
@@ -0,0 +1,70 @@
...
+    name: "Publish to PyPI"
+    if: github.event_name == 'push' && startsWith(github.ref, 'refs/tags/')
+    needs: ["build", "provenance"]
+    permissions:
+      contents: write
+      id-token: write
+    runs-on: "ubuntu-latest"
+    environment:
+      name: pypi
+      url: https://pypi.org/p/ragna
```
pmeier, line 53: These seem to be unused

pyproject.toml
```diff
@@ -111,6 +111,7 @@ dev-all-py310 = { features = ["all", "dev", "py310"], solve-group = "default" }
 dev-all-py311 = { features = ["all", "dev", "py311"], solve-group = "default-py311" }
 dev-all-py312 = { features = ["all", "dev", "py312"], solve-group = "default-py312" }
 dev-all-py313 = { features = ["all", "dev", "py313"], solve-group = "default-py313" }
+build = { features = ["build", "py310"], solve-group = "default" }
```
pmeier, line 114: Why not put `build` into the dev requirements instead of creating another env?

smokestacklightnin: The `dev` feature/environment includes all sorts of requirements not needed for the `build` feature/environment. The _only_ requirements for the `build` feature/environment are the version of Python and the `build` package. We save time on CI and during local environment creation by keeping the Pixi environment simple, all at the expense of a very minimal amount of added complexity. 

Also, developers aren't usually building the Ragna wheel -- that's the job of the CI, so there's no reason for the CI build workflow to include dev environment requirements or for developers to include build requirements when they aren't the ones building.

.github/workflows/release.yml
smokestacklightnin: It is _not_ unused. It is not the same as environment variables. 

It is _strongly_ recommended [in the docs](https://docs.pypi.org/trusted-publishers/adding-a-publisher/) to include it:

> Configuring an environment is optional, but *strongly* recommended: with a GitHub environment, you can apply additional restrictions to your trusted workflow, such as requiring manual approval on each run by a trusted subset of repository maintainers.)

It can be (and should be) used in the trusted publisher config.

.github/actions/build/action.yml
smokestacklightnin: They are used by the `base64-subjects` parameter of the provenance job

.github/workflows/release.yml
smokestacklightnin: It is, but it is also [included automatically with trusted publishing using the pypi publishing workflow](https://blog.trailofbits.com/2024/11/14/attestations-a-new-generation-of-signatures-on-pypi/#:~:text=In%20other%20words%2C%20so%20long,get%20build%20provenance%20by%20default!)

Will remove this manual provenance calculation.

.github/workflows/release.yml
```diff
@@ -0,0 +1,35 @@
...
+        with:
+          environment: "dev-py310"
+
+      - name: Build dists
+        run: pixi run build
+
+      - name: "Publish dists to PyPI"
+        uses: "pypa/gh-action-pypi-publish@76f52bc884231f62b9a034ebfe128415bbaabdfc"
+        with:
+          print-hash: true
```
pmeier, line 35: We don't need this

```suggestion
```

.github/workflows/release.yml
```diff
@@ -0,0 +1,35 @@
...
+  push:
+    branches:
+      - main
+    tags:
+      - "*"
+
+jobs:
+  build-and-publish:
+    name: "Build Package and Publish to PyPI"
+    if: github.event_name == 'push' && startsWith(github.ref, 'refs/tags/')
```
pmeier, line 13: The standard way is 

```suggestion
    if: ${{ github.ref_type == 'tag' }}
```

pyproject.toml
```diff
@@ -141,6 +141,7 @@ test = { cmd="pytest --ignore tests/deploy/ui" }
 # When it is supported, we should remove the argument definitions, as they will have become unnecessary.
 pre-commit = { cmd="pre-commit run {{ which_files }} {{ show_diff }} {{ colorize_output }}", args=[ { "arg" = "which_files", "default" = "--files $(git diff --name-only HEAD | tr '\n' ' ')" }, { "arg" = "show_diff", "default" = "" }, { "arg" = "colorize_output", "default" = "" } ] }
 ci = { depends-on = [ { "task" = "pre-commit", "args" = ["--all-files"] }, "types", "optional-dependencies", "test"] }
+build = { cmd="SOURCE_DATE_EPOCH=$(git log -1 --pretty=%ct) pixi run build" }
```
pmeier: 1. What is the `SOURCE_DATE_EPOCH=$(git log -1 --pretty=%ct)` part about?
2. Why do we need `pixi run` here, when we invoke the other commands directly?

pyproject.toml
```diff
@@ -166,7 +167,8 @@ dev = [
     "mkdocstrings[python]>=0.29.1,<0.30",
     "mkdocs-gen-files>=0.5.0,<0.6",
     "material-plausible-plugin>=0.3.0,<0.4",
-    "mdx-truly-sane-lists>=1.3,<2"
+    "mdx-truly-sane-lists>=1.3,<2",
+    "build"
```
pmeier, line 172: Nit. Just so we truly only ever touch one line when adding something. Had to look twice what about the dependency above changed.

```suggestion
    "build",
```

.github/workflows/release.yml
```diff
@@ -0,0 +1,35 @@
+name: release
+
+on:
+  push:
+    branches:
+      - main
+    tags:
+      - "*"
```
pmeier, line 8: We also want to build on PRs, just not publish

```suggestion
on:
  pull_request:
  push:
    tags: ['v[0-9]+\.[0-9]+\.[0-9]+']
```

.github/workflows/release.yml
pmeier: Plus, this condition needs to be moved into the publish step

pyproject.toml
smokestacklightnin: > 1. What is the `SOURCE_DATE_EPOCH=$(git log -1 --pretty=%ct)` part about?

This sets the datetime of in the metadata of the build to the datetime of the most recent commit. Builds from the same commit should have the same datetime



> Why do we need `pixi run` here, when we invoke the other commands directly?

This is a typo and is already fixed.

pmeier: Doesn't that conflict with `setuptools-scm` that will take the current datetime to add include in the local version? Also what kind of build metadata are we talking about here?

smokestacklightnin: I believe [`setuptools-scm` this](https://setuptools-scm.readthedocs.io/en/latest/config/#environment-variables)

smokestacklightnin: I have removed `SOURCE_DATE_EPOCH` from the build task.

CC: @pmeier

pmeier (approved):

PR #592, opened 2025-06-19
cleanup pixi config

pmeier:
Follow-up to #581.


pyproject.toml
```diff
@@ -92,28 +129,15 @@ platforms = ["linux-64", "win-64", "osx-arm64"]
...
-[tool.pixi.tasks]
+base-py310 = { features = ["dev", "py310"], solve-group = "py310" }
+base-py311 = { features = ["dev", "py311"], solve-group = "py311" }
+base-py312 = { features = ["dev", "py312"], solve-group = "py312" }
+base-py313 = { features = ["dev", "py313"], solve-group = "py313" }
+all-py310 = { features = ["dev", "py310", "all"], solve-group = "py310" }
+all-py311 = { features = ["dev", "py311", "all"], solve-group = "py311" }
+all-py312 = { features = ["dev", "py312", "all"], solve-group = "py312" }
+all-py313 = { features = ["dev", "py313", "all"], solve-group = "py313" }
+all-docker = {features = ["py313", "all"], solve-group="py313"}
```
pmeier, line 140: Reduce cuts down the number of envs from 20 to 9:

- Except for the env that we use as part of our Dockerfile (now named `all-docker`), we don't need the `default*` envs as we need dev dependencies for everything else.
- The `default`, `default-all`, `dev`, and `dev-all` envs are not required as they are just "aliases", while taking up real space.

PR #596, opened 2025-06-20
fix mkdocstrings config

pmeier:


mkdocs.yml
```diff
@@ -49,11 +49,12 @@ plugins:
       handlers:
         python:
           options:
+            docstring_style: google
             show_source: false
             show_root_heading: true
             show_bases: false
             unwrap_annotated: true
-          import:
+          inventories:
```
pmeier, line 57: This was emitting a deprecation warning

mkdocs.yml
```diff
@@ -49,11 +49,12 @@ plugins:
       handlers:
         python:
           options:
+            docstring_style: google
```
pmeier, line 52: This just makes it explicit. mkdocstrings figured it out implicitly before.

smokestacklightnin (approved):

